PR #141, opened 2018-03-23
pull groups out of deployment

### LINKED ISSUE #138: orders not closed on deployment close

### LINKED ISSUE #139: duplicated state due to nested fields in Deployment
the deployment also saves a snapshot of its groups on creation. These will get out of sync with the separate deploymentGroup objects.

aastein:
fixes #139 
fixes #138


app/deployment/app.go
```diff
@@ -401,16 +424,30 @@ func (a *app) doDeliverCloseTx(ctx apptypes.Context, tx *types.TxCloseDeployment
...
+
 	deployment.State = types.Deployment_CLOSING
 
-	for i, group := range deployment.Groups {
+	for i, group := range groups {
 		// begin for each group
 		if group.State != types.DeploymentGroup_CLOSING {
 			group.State = types.DeploymentGroup_CLOSING
-			deployment.Groups[i] = group
+			groups[i] = group
```
boz: we can use pointers (remove `gogoproto.nullable=false`) if it continues to be such a PITA/confusing.

types/types.proto
```diff
@@ -121,7 +121,7 @@ message DeploymentGroup {
 }
 
 message DeploymentGroups {
-  repeated DeploymentGroup items = 1;
+  repeated DeploymentGroup items = 1 [(gogoproto.nullable) = false];
```
boz: Kinda torn about this.

PR #154, opened 2018-03-27
Billing

### LINKED ISSUE #142: Lease Billing
1) Start on lease=active
2) Stop on lease=closing

### LINKED ISSUE #151: flapping test: cmd/akash/connection_test.go
These are failing and/or locking up randomly.

The lockup might be from the node created in `cmd/common/monitor_test.go` not being completely shut down - the client gets a connection just before the node shuts down and it waits forever.

somewhat related to #124

aastein:
fixes #142
fixes #151 

On lease create, check if tenant has balance >= lease price
process leases in market engine
for each active lease, send a billTenantTx
Lease app accepts and checks BillTenantTx
Lease app subtracts lease.Price from the tenant and sends lease.Price to the provider.


app/app.go
```diff
@@ -235,6 +235,8 @@ func (app *app) Commit() tmtypes.ResponseCommit {
 		app.mfacilitator.OnCommit(app.state)
 	}
 
+	lease.ProcessLeases(app.state)
```
boz: `log.Error()` or similar.

app/market/engine_test.go
```diff
@@ -4,25 +4,35 @@ import (
...
 	require.NoError(t, state_.Deployment().Save(deployment))
 
+	state_, tx := testOrder(t, state_, tenant, deployment, groups)
+	state_ = testLease(t, state_, provider, deployment, groups, tx)
+}
+
+func testOrder(t *testing.T, state state.State, tenant *types.Account, deployment *types.Deployment, groups *types.DeploymentGroups) (state.State, *types.TxCreateOrder) {
 	for idx := range groups.GetItems() {
-		require.NoError(t, state_.DeploymentGroup().Save(groups.GetItems()[idx]))
+		require.NoError(t, state.DeploymentGroup().Save(groups.GetItems()[idx]))
```
boz: should be able to do the

```go
for _, item := range groups.GetItems() {
  require.NoError(t,state.Save(item))
}
```

thing here.

cmd/akashd/init.go
```diff
@@ -5,10 +5,11 @@ import (
 	"github.com/ovrclk/akash/types/base"
 	"github.com/ovrclk/akash/util/initgen"
 	"github.com/spf13/cobra"
+	"math"
 )
 
 const (
-	maxTokens uint64 = 1000000000
+	maxTokens uint64 = math.MaxUint64
```
boz: I think this will screw up the integration tests [here](https://github.com/ovrclk/akash/blob/master/_integration/cmp/account.go#L28).

* Simple matter of changing the value in the test or ideally just referencing a constant.
* The json lib we're using there maxes out at `int64` which won't be able to hold `math.MaxUint64`.
* Also, I think we're supposed to have a billion tokens total.

testutil/state.go
```diff
@@ -20,6 +21,7 @@ func CreateAccount(t *testing.T, state state.State) (*types.Account, crypto.Priv
 	key := PrivateKey(t)
 	account := &types.Account{
 		Address: key.PubKey().Address(),
+		Balance: math.MaxUint64 / 2,
```
boz: yeah let's do a constant somewhere for initial token count.

app/market/engine.go
```diff
@@ -87,7 +87,7 @@ func (e engine) processDeployment(state state.State, w txBuffer, deployment type
 			if !activeFound && order.State == types.Order_OPEN || order.State == types.Order_MATCHED {
 				activeFound = true
 			}
-			if order.EndAt <= height {
+			if order.State == types.Order_OPEN && order.EndAt <= height {
```
boz: :1st_place_medal:

cmd/akashd/init.go
boz: Yeah it read it as `-1`:

```
Evaluation of main failed:
/main/account-send/account-balance/akash: ../akash -d ./data/client query account 22CDFF3E908986DE2F4B1D279667BA02B84CD7D6: balance: received -1 expected 1000000000
-===[BEGIN STDOUT]===-
{
  "address": "22CDFF3E908986DE2F4B1D279667BA02B84CD7D6",
  "balance": 18446744073709551615
}
```

Issue #186, opened 2018-04-23
Send deployment image to provider


aastein:
- [x] add public network address to provider yaml, so the value is on the chain
- [x] send yaml file with docker image location to the network address provided
  - [x] add yaml file path to deployment command
  - [x] todo: #187 encrypt with providers pubkey and sign the file with the tenant's pub key
- [x] in Akash, add a new command to start a web server to receive the yaml file with the docker image
  - [x]  todo: #187 once the web server receives the file, verify sig and decrypt to get raw yaml file
  - [x] write the file to a location, and notify whatever Adam builds that an image file has been received
- [x] add a send manifest command for tenant

PR #188, opened 2018-04-24
Tenant sends manifest to provider over http

aastein:
fixes: #186


app/provider/app.go
```diff
@@ -171,6 +171,14 @@ func (a *app) doCheckTx(ctx apptypes.Context, tx *types.TxCreateProvider) tmtype
 		return tmtypes.ResponseCheckTx{Code: code.INVALID_TRANSACTION, Log: "invalid nonce"}
 	}
 
+	// XXX: more address validation
```
boz: you can use [this](https://golang.org/pkg/net/url/#Parse) and make sure that there's a scheme & host.

types/types.proto
```diff
@@ -61,7 +61,8 @@ message Provider {
   option (gogoproto.compare) = true;
   bytes  address = 1 [(gogoproto.customtype)="github.com/ovrclk/akash/types/base.Bytes",(gogoproto.nullable) = false];
   bytes  owner = 2 [(gogoproto.customtype)="github.com/ovrclk/akash/types/base.Bytes",(gogoproto.nullable) = false];
-  repeated ProviderAttribute attributes = 3 [(gogoproto.nullable) = false];
+  string netaddr = 3;
```
boz: let's call this `host_uri` (`HostURI`), `akash_uri`, or similar.  `netaddr` kinda implies `ip:port` pair.

PR #209, opened 2018-05-31
provider: initial kube workload execution

boz:
* provider: run workloads in kubernetes
 * sdl: add `accept`, list of hosts to accept requests for.
 * sdl: populate service args, env.
 * types/provider: fix yaml parsing.
 * _run/kube: single-node workload demo.


aastein:
we should have the demo spin up akash and run the demo

PR #210, opened 2018-06-01
cleanup: move id methods to types package

boz:


provider/manifest/handler.go
```diff
@@ -218,12 +226,15 @@ func (h *handler) checkManifestState(ctx context.Context, mstate *manifestState,
...
 
 		// publish complete manifest
-		h.bus.Publish(event.ManifestReceived{
-			LeaseID:    *mstate.leaseID,
-			Manifest:   mstate.request.Manifest,
-			Deployment: mstate.deployment,
-		})
-		h.session.Log().Debug("manifest complete", "deployment", did.EncodeString())
+		for _, lev := range mstate.leases {
+			h.bus.Publish(event.ManifestReceived{
```
golangci: Error return value of `h.bus.Publish` is not checked

aastein: this is cool

PR #219, opened 2018-06-06
cluster deployments: teardown on cancel

boz:
* provider/cluster: teardown on deployment/fulfillment close.
* provider/manifest: allow multiple leases per manifest.
* provider: handle multiple manifest groups.
* provider: match Manifest and Deployment groups via Name field.
* types: add `GroupSpec.Name`, `DeploymentGroup.Name` fields.
* sdl: populate `GroupSpec.Name` field.


provider/cluster/mocks/client.go
```diff
@@ -0,0 +1,38 @@
+// Code generated by mockery v1.0.0
```
golangci: package comment should be of the form "Package mocks ..."

provider/cluster/monitor.go
```diff
@@ -0,0 +1,193 @@
...
+	"fmt"
+
+	lifecycle "github.com/boz/go-lifecycle"
+	"github.com/ovrclk/akash/types"
+	"github.com/tendermint/tmlibs/log"
+)
+
+type deploymentState string
+
+const (
```
golangci: only the first constant has an explicit type

provider/cluster/service.go
```diff
@@ -109,22 +123,101 @@ loop:
...
+				if err := state.monitor.update(mgroup); err != nil {
+					s.log.Error("updating deployment", "err", err)
+				}
+
+			case *event.TxCloseDeployment:
+
+				// teardown/undeploy managed deployments
+
+				for _, dm := range s.deployments {
+					if bytes.Compare(ev.Deployment, dm.reservation.OrderID().DeploymentID()) == 0 {
```
golangci: should use bytes.Equal(ev.Deployment, dm.reservation.OrderID().DeploymentID()) instead

boz:
adding more to this.

PR #236, opened 2018-06-12
Leases: transfer tokens

### LINKED ISSUE #235: leases: tokens not transferring
sad-trombone.wav

boz:
fixes #235

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -29,6 +29,8 @@ import (
 type Application interface {
 	tmtypes.Application
 	ActivateMarket(market.Actor, *tmtmtypes.EventBus) error
+
+	App(name string) apptypes.Application
 }
 
 type app struct {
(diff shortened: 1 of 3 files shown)


app/app_test.go
```diff
@@ -89,3 +95,80 @@ func TestApp(t *testing.T) {
...
+
+	_, keyfrom := testutil.PrivateKeySigner(t)
+	addrfrom := keyfrom.PubKey().Address().Bytes()
+
+	keyto := testutil.PrivateKey(t)
+	addrto := keyto.PubKey().Address().Bytes()
+
+	commitState, cacheState := testutil.NewState(t, &types.Genesis{
+		Accounts: []types.Account{
+			types.Account{Address: addrfrom, Balance: balance, Nonce: nonce},
```
golangci: File is not gofmt-ed with -s

app/app.go
```diff
@@ -241,12 +252,15 @@ func (app *app) EndBlock(req tmtypes.RequestEndBlock) tmtypes.ResponseEndBlock {
 func (app *app) Commit() tmtypes.ResponseCommit {
 	app.trace("Commit")
 
-	err := app.cacheState.Write()
-	if err != nil {
-		app.log.Error("error when writing to cache")
+	if err := lease.ProcessLeases(app.cacheState); err != nil {
```
aastein: why use cacheState here?

boz: thought was to keep all modifications in the ephemeral db until commit.  should we be using `commitState`?

aastein: I thought the process lease routine sends transactions, but it modifies state directly. Using cache state is the correct thing to do 👍

PR #258, opened 2018-06-27
run/multi: add make install and update README

gosuri:


_run/multi/env.sh
```diff
@@ -4,10 +4,5 @@ nodeuri(){
...
 
-akash(){
-  _akash -n "$(nodeuri)" "$@"
-}
-
-akashd(){
-  _akashd "$@"
-}
+export AKASH_NODE=$(nodeuri)
+export AKASH_DATA=$GOPATH/src/github.com/ovrclk/akash/_run/multi/data/client
```
boz: doesn't this break `run.sh`?

gosuri:
reverting, not a priority

PR #272, opened 2018-07-06
cmd/akash: add key show

### LINKED ISSUE #270: cli: add script optimized mechanism to display keys
Currently, the keys are retrieved using `akash key list` which is good for display but not optimized for scripting; for example to query the master account: 

```
$ akash query account $(akash key list  | grep master | awk '{print $2}')
```

Ideally, a simpler way to query for individual keys makes scripting experience better. I propose the adding `akash key show [key]`. The above command could be rewritten as follows:

```
$ akash query account $(akash key show master)
```

gosuri:
Fixes  #270 

Currently, the keys are retrieved using `akash key list` which is good for display but not optimized for scripting; for example to query the master account: 

```
$ akash query account $(akash key list  | grep master | awk '{print $2}')
```

This change add a simpler way to query for individual keys makes scripting experience better. The above command could be rewritten as follows:

```
$ akash query account $(akash key show master)
```

### DIFF
--- a/cmd/akash/key.go
+++ b/cmd/akash/key.go
@@ -20,9 +20,9 @@ func keyCommand() *cobra.Command {
 	}
 	cmd.AddCommand(keyCreateCommand())
 	cmd.AddCommand(keyListCommand())
+	cmd.AddCommand(keyShowCommand())
 	return cmd
 }
-
 func keyCreateCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "create [name]",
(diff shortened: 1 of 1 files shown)


cmd/akash/key.go
```diff
@@ -81,3 +81,37 @@ func doKeyListCommand(session session.Session, cmd *cobra.Command, args []string
...
+		return err
+	}
+
+	info, err := kmgr.Get(args[0])
+	if err != nil {
+		return err
+	}
+
+	if len(info.Address()) == 0 {
+		fmt.Errorf("key not found %s", args[0])
```
golangci: result of fmt.Errorf call not used

cmd/akash/key.go
```diff
@@ -81,3 +81,36 @@ func doKeyListCommand(session session.Session, cmd *cobra.Command, args []string
...
+		Use:   "show [name]",
+		Short: "display a key",
+		RunE:  session.WithSession(session.RequireRootDir(doKeyShowCommand)),
+	}
+	session.AddFlagKeyType(cmd, cmd.Flags())
+	return cmd
+}
+
+func doKeyShowCommand(session session.Session, cmd *cobra.Command, args []string) error {
+	if len(args) != 1 {
```
boz: Declare this in the command like [here](https://github.com/ovrclk/akash/blob/41e00ba0cf7ef5c811917ef9a6018686aaa5e280/cmd/akash/provider.go#L50)

PR #310, opened 2018-07-25
provider/cluster: inventory management

### LINKED ISSUE #215: Inventory mgt: Provider only bids on leases it can fulfill

boz:
fixes #215


types/base/bytes.go
```diff
@@ -49,6 +53,14 @@ func (this Bytes) Compare(that Bytes) int {
 	return bytes.Compare([]byte(this), []byte(that))
 }
 
+func (this Bytes) Equal(that Bytes) bool {
```
golangci: receiver name should be a reflection of its identity; don't use generic names such as "this" or "self"

PR #321, opened 2018-07-27
Add community issue template

nalesandro:


nalesandro:
Added items per your feedback thx

PR #337, opened 2018-07-30
unify validation

boz:


validation/config.go
```diff
@@ -21,3 +35,7 @@ func init() {
 		panic(err)
 	}
 }
+
+func Config() config {
```
golangci: exported func Config returns unexported type validation.config, which can be annoying to use

PR #342, opened 2018-07-31
pricing: fix panic when min=max

boz:


provider/bidengine/pricing.go
```diff
@@ -0,0 +1,48 @@
...
+
+	for _, group := range resources.GetResources() {
+		rmax += group.Price * uint64(group.Count)
+		mem += group.Unit.Memory * uint64(group.Count)
+	}
+
+	cmin := uint64(float64(mem) * float64(cfg.MinGroupMemPrice) / float64(unit.Gi))
+	cmax := uint64(float64(mem) * float64(cfg.MaxGroupMemPrice) / float64(unit.Gi))
+
+	if cmax > uint64(rmax) {
```
golangci: unnecessary conversion

provider/bidengine/pricing.go
```diff
@@ -0,0 +1,48 @@
...
+		cmax = uint64(rmax)
```
golangci: unnecessary conversion

PR #344, opened 2018-08-01
akash provider: multipl fixes

boz:
* fix required flags
 * show error if api host is not expected provider
 * actually show all providers


provider/grpc/server.go
```diff
@@ -165,28 +143,54 @@ func (s server) ServiceLogs(req *types.LogRequest, server types.Cluster_ServiceL
...
 		log:     log,
-		Client:  client,
 	}
-	types.RegisterClusterServer(s.Server, s)
-	return s
 }
 
-func (s *server) listenAndServe() error {
-	l, err := net.Listen(s.network, s.port)
+func run(server *server, ctx context.Context, address string) error {
```
golangci: context.Context should be the first parameter of a function

Issue #448, opened 2020-03-03
use 'akt' as denomination instead of default 'stake'


gosuri:
The genesis file,  `staking.params.bond_denom` should be 'akt'. 

See https://docs.akash.network/v/testnet2/testnet-2/testnet2#attach-genesis-account

```json
    "staking": {
      "params": {
        "unbonding_time": "1814400000000000",
        "max_validators": 100,
        "max_entries": 7,
        "historical_entries": 0,
        "bond_denom": "stake"
      },
      "last_total_power": "0",
      "last_validator_powers": null,
      "validators": null,
      "delegations": null,
      "unbonding_delegations": null,
      "redelegations": null,
      "exported": false
    },
```

anilCSE:
I think this should be handled via changes in genesis file directly. It doesn't look like a straight forward change in `app.go`, for sdk `params`.

boz:
Done by @anilCSE via editing the genesis file.

PR #451, opened 2020-03-04
Update bond denom as akt

anilCSE:
Fixes: #448
---
name: Pull Request
about: Submit a pull request
title: ''
labels: ''
assignees: '@boz @gosuri'

---

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/cosmos/cosmos-sdk/x/params"
 	"github.com/cosmos/cosmos-sdk/x/staking"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 
 	"github.com/ovrclk/akash/x/deployment"
 	"github.com/ovrclk/akash/x/market"
@@ -36,6 +37,7 @@ import (
 
 const (
 	appName = "akash"
+	denom   = "akt"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


anilCSE:
Looks like it is easy to modify in genesis file directly than something from code. There's no direct way to update `mint` token denom from code afaik. This should be raised as an issue agaist  `cosmos-sdk`. 

/cc @jackzampolin @adamlondon

boz:
This looks like it will work, but it is ultimately not the best route to take, I think.  In general, try not to use `init()` for overwriting variables in other packages - it leads to conflicts and difficult-to-track bugs (what if two packages do this?  what if this package isn't included? etc...).

There should be another way to configure this - either when bootstrapping the chain or via the cosmos-sdk parameters mechanism.

anilCSE:
> This looks like it will work, but it is ultimately not the best route to take, I think. In general, try not to use `init()` for overwriting variables in other packages - it leads to conflicts and difficult-to-track bugs (what if two packages do this? what if this package isn't included? etc...).
> 
> There should be another way to configure this - either when bootstrapping the chain or via the cosmos-sdk parameters mechanism.

I also feel that, this should go here. I didn't find a better way to handle this easily than editing these params in genesis directly.

anilCSE:
So, let's move this to ice-box for now. We'll revisit this when we find a better solution. For now, lets change the params in `genesis.json` directly

PR #464, opened 2020-03-19
Codebase migration refactor

### LINKED ISSUE #457: DCS-5: Cosmos SDK Codebase Migration
Rewrite Akash codebase to Cosmos SDK to improve scalability, efficiency, and security.

Proposal: [DCS-5](https://dcs.akash.network/spec/dcs-5)

Related Issues: #442, #454, #455

Scheduled Completion: March 2020.

Docs: [Common Commands](https://docs.akash.network/v/testnet2/usage/cli/general-commands)
Source: [ovrclk/akash](https://github.com/ovrclk/akash)
Questions: [Developer Chat](https://akash.network/chat)

fedekunze:
closes #457 

## Motivation and Context

## Types of changes

## Checklist:


app/app.go
```diff
@@ -9,276 +9,255 @@ import (
...
+	SlashingKeeper   slashing.Keeper
+	MintKeeper       mint.Keeper
+	DistrKeeper      distr.Keeper
+	GovKeeper        gov.Keeper
+	CrisisKeeper     crisis.Keeper
+	ParamsKeeper     params.Keeper
+	EvidenceKeeper   evidence.Keeper
+	DeploymentKeeper deployment.Keeper
+	MarketKeeper     market.Keeper
+	ProviderKeeper   provider.Keeper
```
boz: Why are these public?

fedekunze: making the keepers public leverages the `AkashApp` as the source for integration tests (as it should be). See https://github.com/cosmos/cosmos-sdk/issues/4875 for reference

app/app.go
```diff
@@ -348,7 +319,65 @@ func (app *AkashApp) ExportAppStateAndValidators(
...
+
+// GetSubspace returns a param subspace for a given module name.
+//
+// NOTE: This is solely to be used for testing purposes.
+func (app *AkashApp) GetSubspace(moduleName string) params.Subspace {
+	return app.subspaces[moduleName]
+}
+
+// GetMaccPerms returns a copy of the module account permissions
+func GetMaccPerms() map[string][]string {
```
boz: Why not use the previous definition?

app/app.go
boz: fwiw I think that the reality is that they don't need to be fields in the struct at all

app/app.go
```diff
@@ -9,276 +9,255 @@ import (
...
 	return cdc.Seal()
 }
 
 // https://github.com/cosmos/sdk-tutorials/blob/c6754a1e313eb1ed973c5c91dcc606f2fd288811/app.go#L73
 
-// NewApp creates and returns a new Akash App.
-func NewApp(
-	logger log.Logger, db dbm.DB, tio io.Writer, options ...func(*bam.BaseApp),
+// NewAkashApp creates and returns a new Akash App.
+func NewAkashApp(
```
boz: Why are you changing the name?

app/app.go
```diff
@@ -9,276 +9,255 @@ import (
...
-		genutil.ModuleName,
-
-		// akash
-		deployment.ModuleName,
-		provider.ModuleName,
-		market.ModuleName,
+		auth.ModuleName, distr.ModuleName, staking.ModuleName, bank.ModuleName,
+		slashing.ModuleName, gov.ModuleName, mint.ModuleName, supply.ModuleName,
+		crisis.ModuleName, genutil.ModuleName, evidence.ModuleName,
+		deployment.ModuleName, provider.ModuleName, market.ModuleName, // akash
```
boz: why are you changing the format here?

app/app.go
```diff
@@ -309,26 +280,26 @@ func NewApp(
...
-	return app.mm.InitGenesis(ctx, genesisState)
-}
+// Name returns the name of the App
+func (app *AkashApp) Name() string { return app.BaseApp.Name() }
 
-// application updates every begin block
-func (app *AkashApp) beginBlocker(
-	ctx sdk.Context, req abci.RequestBeginBlock) abci.ResponseBeginBlock {
+// BeginBlocker application updates every begin block
+func (app *AkashApp) BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock) abci.ResponseBeginBlock {
```
boz: why are you making these public?

app/app.go
boz: You need to access all of the keepers?

boz: anyways, please keep them in a separate struct for organization/isolation and only export the keepers that you need to.

gosuri:
great! @fedekunze please also to make sure to document your findings and reasons, especially for major interface changes (names, etc) since we have other members in the team that need to rebase and we want to make sure practices are documented well to reduce knowledge asymmetry.

add a section "refactor considerations" for [DCS-5].(https://github.com/ovrclk/dcs/blob/master/spec/dcs-5/README.md)

app/app.go
fedekunze: because the constructor name should match the returned value (i.e `AkashApp`)

app/app.go
fedekunze: > fwiw I think that the reality is that they don't need to be fields in the struct at all?

As you prefer, I'd still need to make the `keeper` internal struct public tho.

> You need to access all of the keepers?

I need access to all the keepers used by the Akash modules

anilCSE:
> great! @fedekunze please also to make sure to document your findings and reasons, especially for major interface changes (names, etc) since we have other members in the team that need to rebase and we want to make sure practices are documented well to reduce knowledge asymmetry.
> 
> add a section "refactor considerations" for [DCS-5].(https://github.com/ovrclk/dcs/blob/master/spec/dcs-5/README.md)

I checked all these updates and are in sync with the standards sdk/gaia follows. There are a plenty of discussions before making all these changes in gaia and simapp. 

I also feel that, keeping akash code close to gaia will help in reducing redundant work for upgrades (sdk changes). Otherwise, it will be an overhead to upgrade.

app/app.go
fedekunze: uses too many lines. I usually break the line at ~80 chars

app/app.go
anilCSE: One reason is, it helps in simapp tests

fedekunze: that's the standard of the SDK framework. It doesn´t make much difference since the function calls a private function. Otherwise, you can call the module manager, `mm`, directly so there's no need to have a private function call another private function imho.

app/app.go
boz: Please keep the previous format.

app/app.go
boz: Exporting symbols comes with many costs.  Please don't export any symbols unless absolutely necessary.

app/app.go
boz: ok please keep them in an isolated struct and export them on an as-needed basis.

boz:
The Gia and cosmos-sdk "standards" are constantly-moving targets.  They also tend to diverge from idiomatic go patterns.

Refactoring for the sake of refactoring can be necessary at times, but it is not here.

app/app.go
```diff
@@ -9,276 +9,255 @@ import (
...
+		evidence.AppModuleBasic{},
 
 		// akash
 		deployment.AppModuleBasic{},
 		market.AppModuleBasic{},
 		provider.AppModuleBasic{},
 	)
+
+	// module account permissions
+	maccPerms = map[string][]string{
```
boz: the reason for having these be functions was to prevent mutation - the functions return a new copy every time.  please revert to the previous pattern.

app/app.go
boz: okay.  would have preferred to change the struct name to `App`.

app/app.go
fedekunze: that's what `GetMaccPerms` does too 🙂

boz: why did you change it then?

x/deployment/handler/handler.go
```diff
@@ -12,19 +12,19 @@ import (
 func NewHandler(keeper keeper.Keeper, mkeeper MarketKeeper) sdk.Handler {
 	return func(ctx sdk.Context, msg sdk.Msg) (*sdk.Result, error) {
 		switch msg := msg.(type) {
-		case types.MsgCreate:
+		case types.MsgCreateDeployment:
```
anilCSE: +1

x/deployment/types/msgs.go
```diff
@@ -2,107 +2,103 @@ package types
...
-type MsgCreate struct {
-	Owner sdk.AccAddress `json:"owner"`
-	// Sequence uint64         `json:"sequence"`
-	// Version []byte      `json:"version"`
-	Groups []GroupSpec `json:"groups"`
+// msg types
+const (
+	TypeMsgCreateDeployment = "create_deployment"
+	TypeMsgUpdateDeployment = "update_deployment"
+	TypeMsgCloseDeployment  = "close_deployment"
```
boz: Can you make these values be the same as those in the [codec registration](https://github.com/ovrclk/akash/pull/464/files/fc49dbd9eca7cb378cf096094bc213d5e28d777a..1f149e55e87b155d142e93543cffa76859043f0b#diff-e5c9ba0cb12c3a49b6d1e97c98885903R15-R17) and use these constants there?

If they don't need to be exported, please make them private.

x/market/types/id.go
```diff
@@ -39,6 +41,12 @@ func (id OrderID) Equals(other OrderID) bool {
 
 // Validate method for OrderID and returns nil
 func (id OrderID) Validate() error {
+	if id.Owner.Empty() {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, "owner cannot be empty")
+	}
+	if id.DSeq == 0 && id.GSeq == 0 && id.OSeq == 0 {
```
boz: I believe you can use `return id.GroupID().Validate()`.  Same goes for the rest of them, I'll illustrate below.

x/market/types/id.go
```diff
@@ -95,6 +103,12 @@ func (id BidID) DeploymentID() dtypes.DeploymentID {
 
 // Validate validates bid instance and returns nil
 func (id BidID) Validate() error {
+	if id.Owner.Empty() {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, "owner cannot be empty")
+	}
+	if id.DSeq == 0 && id.GSeq == 0 && id.OSeq == 0 {
+		return sdkerrors.Wrap(ErrInvalidOrder, "at least one id must be > 0")
+	}
```
boz: `return id.OrderID().Validate()`

x/market/types/msgs.go
```diff
@@ -2,20 +2,28 @@ package types
...
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+// msg types
+const (
+	TypeMsgCreateBid  = "create_bid"
+	TypeMsgCloseBid   = "close_bid"
+	TypeMsgCloseOrder = "close_order"
```
boz: Same comment from `x/deployment/types` above.

x/provider/types/msgs.go
```diff
@@ -4,92 +4,109 @@ import (
...
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
-// MsgCreate defines an SDK message for creating a provider
-type MsgCreate Provider
+// msg types
+const (
+	TypeMsgCreateProvider = "create_provider"
+	TypeMsgUpdateProvider = "update_provider"
+	TypeMsgDeleteProvider = "delete_provider"
```
boz: same comment as `x/deployment/types` above.

app/app.go
```diff
@@ -68,24 +68,28 @@ var (
...
-	// module account permissions
-	maccPerms = map[string][]string{
+	// module accounts that are allowed to receive tokens
+	allowedReceivingModAcc = map[string]bool{
+		distr.ModuleName: true,
+	}
+)
+
+// MaccPerms returns the module account permissions
+func MaccPerms() map[string][]string {
```
boz: Please just revert this to the original.  Does it need to be exported?

app/app.go
```diff
@@ -9,274 +9,267 @@ import (
...
 
 var (
-	mbasics = module.NewBasicManager(
-		genutil.AppModuleBasic{},
 
-		// accounts, fees.
+	// ModuleBasics defines the module BasicManager is in charge of setting up basic,
+	// non-dependant module elements, such as codec registration
+	// and genesis verification.
+	ModuleBasics = module.NewBasicManager(
```
boz: Please revert this to the original.  Exporting it as a function prevents other modules from re-assigning it.

app/app.go
```diff
@@ -9,274 +9,267 @@ import (
...
-		supply.StoreKey,
-		staking.StoreKey,
-		mint.StoreKey,
-		deployment.StoreKey,
-		market.StoreKey,
-		provider.StoreKey,
+		bam.MainStoreKey, auth.StoreKey, staking.StoreKey,
+		supply.StoreKey, mint.StoreKey, distr.StoreKey, slashing.StoreKey,
+		gov.StoreKey, params.StoreKey, evidence.StoreKey,
+		deployment.StoreKey, market.StoreKey, provider.StoreKey,
```
boz: revert to original format for all of these, please.

app/app.go
```diff
@@ -348,7 +333,62 @@ func (app *AkashApp) ExportAppStateAndValidators(
...
 
-	validators = staking.WriteValidators(ctx, app.keeper.staking)
+	validators = staking.WriteValidators(ctx, app.Keepers.Staking)
 
 	return appState, validators, nil
 }
+
+// ModuleAccountAddrs returns all the app's module account addresses.
+func (app *App) ModuleAccountAddrs() map[string]bool {
+	modAccAddrs := make(map[string]bool)
```
boz: general template like was done [originally](https://github.com/ovrclk/akash/pull/464/files/fc49dbd9eca7cb378cf096094bc213d5e28d777a..1f149e55e87b155d142e93543cffa76859043f0b#diff-e5c9ba0cb12c3a49b6d1e97c98885903R15-R17)
```go
coll := generator()
vals := make(map[string]bool,len(coll))
for k, v := range coll {
  // ...
}
return vals
```

x/deployment/types/id.go
```diff
@@ -50,5 +59,11 @@ func (id GroupID) Equals(other GroupID) bool {
 
 // Validate method for GroupID and returns nil
 func (id GroupID) Validate() error {
+	if id.Owner.Empty() {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, "owner cannot be empty")
+	}
+	if id.DSeq == 0 && id.GSeq == 0 {
```
boz: `return id.DeploymentID().Validate()`

x/market/types/id.go
fedekunze: I can't bc `OSeq` is not defined for `GroupID`

x/deployment/types/id.go
fedekunze: ditto, I can't because `GSeq` is not defined for `DeploymentID`

x/market/types/id.go
boz: * `BidID = OrderID + Provider`; `Validate(BidID) = Validate(OrderID) + Validate(Provider)`
* `OrderID = GroupID + Oseq`; `Validate(OrderID) = Validate(GroupID) + Validate(Oseq)` 
* `GroupID = DeploymentID + Gseq`; `Validate(GroupID) = Validate(DeploymentID) + Validate(GSeq)`

x/deployment/types/id.go
boz: `GroupID = DeploymentID + GSeq`; `Validate(GroupID) = Validate(DeploymentID) + Validate(GSeq)`

x/market/types/id.go
boz: In other words, re-use the validations from the contained structures.  Unrolling at every level will lead to bugs, missed validations, and make refactoring more difficult.  There are [utilities for converting between the various IDs](https://github.com/ovrclk/akash/blob/9161a1abe6bf21223ced0b81a2f5167a02d012e2/x/market/types/id.go).

See the [`Equals()`](https://github.com/ovrclk/akash/blob/9161a1abe6bf21223ced0b81a2f5167a02d012e2/x/market/types/id.go#L66-L69) methods for an example.

x/deployment/types/msgs.go
```diff
@@ -7,9 +7,9 @@ import (
 
 // msg types
 const (
-	TypeMsgCreateDeployment = "create_deployment"
-	TypeMsgUpdateDeployment = "update_deployment"
-	TypeMsgCloseDeployment  = "close_deployment"
+	typeMsgCreateDeployment = "create_deployment"
+	typeMsgUpdateDeployment = "update_deployment"
+	typeMsgCloseDeployment  = "close_deployment"
```
boz: Either

```go
// msgs.go
typeMsgCreateDeployment = ModuleName + "/msg-create-deployment"

// codec.go
cdc.RegisterConcrete(MsgCreateDeployment{}, typeMsgCreateDeployment, nil)
```

or
```go
// msgs.go
typeMsgCreateDeployment = msg-create-deployment"

// codec.go
cdc.RegisterConcrete(MsgCreateDeployment{}, ModuleName + "/" + typeMsgCreateDeployment, nil)
```
Same goes for the rest of these.

x/market/types/codec.go
```diff
@@ -12,9 +12,9 @@ func init() {
 
 // RegisterCodec register concrete types on codec
 func RegisterCodec(cdc *codec.Codec) {
-	cdc.RegisterConcrete(MsgCloseOrder{}, ModuleName+"/msg-close-order", nil)
-	cdc.RegisterConcrete(MsgCreateBid{}, ModuleName+"/msg-create-bid", nil)
-	cdc.RegisterConcrete(MsgCloseBid{}, ModuleName+"/msg-close-bid", nil)
+	cdc.RegisterConcrete(MsgCloseOrder{}, ModuleName+"/MsgCloseOrder", nil)
+	cdc.RegisterConcrete(MsgCreateBid{}, ModuleName+"/MsgCreateBid", nil)
+	cdc.RegisterConcrete(MsgCloseBid{}, ModuleName+"/MsgCloseBid", nil)
```
boz: * use the constants defined in `msgs.go`.
* use the original names.

x/market/types/types.go
```diff
@@ -12,12 +12,12 @@ import (
...
+	// OrderClosed is used when state of order is closed
+	OrderClosed OrderState = iota // default (0)
 	// OrderOpen is used when state of order is open
-	OrderOpen OrderState = iota
+	OrderOpen
 	// OrderMatched is used when state of order is matched
-	OrderMatched OrderState = iota
-	// OrderClosed is used when state of order is closed
-	OrderClosed OrderState = iota
+	OrderMatched
```
boz: * `Closed` is not the initial state.
* Some linters will complain about this formatting.  Why are you changing it to begin with?

x/market/types/types.go
```diff
@@ -85,14 +85,14 @@ func (o Order) validateMatchableState() error {
...
-	BidOpen BidState = iota
+	BidOpen
 	// BidMatched is used when state of bid is matched
-	BidMatched BidState = iota
+	BidMatched
 	// BidLost is used when state of bid is lost
-	BidLost BidState = iota
-	// BidClosed is used when state of bid is closed
-	BidClosed BidState = iota
+	BidLost
```
boz: Same as above.

x/market/types/types.go
```diff
@@ -111,12 +111,12 @@ func (obj Bid) ID() BidID {
...
-	// LeaseActive is used when state of lease is active
-	LeaseActive LeaseState = iota
-	// LeaseInsufficientFunds is used when lease has insufficient funds
-	LeaseInsufficientFunds LeaseState = iota
 	// LeaseClosed is used when state of lease is closed
 	LeaseClosed LeaseState = iota
+	// LeaseActive is used when state of lease is active
+	LeaseActive
+	// LeaseInsufficientFunds is used when lease has insufficient funds
+	LeaseInsufficientFunds
```
boz: Again.

app/test_helpers.go
```diff
@@ -0,0 +1,34 @@
...
+	abci "github.com/tendermint/tendermint/abci/types"
+	"github.com/tendermint/tendermint/libs/log"
+	dbm "github.com/tendermint/tm-db"
+
+	"github.com/cosmos/cosmos-sdk/codec"
+	"github.com/cosmos/cosmos-sdk/simapp"
+)
+
+// Setup initializes a new SimApp. A Nop logger is set in SimApp.
+func Setup(isCheckTx bool) *App {
```
boz: Is this used for anything other than tests?  Does it need access to private fields or functions?

Not sure it should be here.  If nothing else, call it `SetupTest()` and put it in a file that ends in `_test.go` so that it's only available during testing.

x/deployment/types/msgs.go
fedekunze: if you check the godoc for the `RegisterConcrete` the name must match the type name

x/market/types/codec.go
fedekunze: see above

x/market/types/types.go
fedekunze: the default value for `uint8` is 0, that means that all default states would have been open

fedekunze: check https://github.com/uber-go/guide/blob/master/style.md#start-enums-at-one for reference

fedekunze: > Some linters will complain about this formatting.

which linter complain about this format?

x/market/types/types.go
fedekunze: see reply

x/market/types/types.go
boz: Yes of course I know the initial value for `iota`.

When orders are created they are `open`, hence that being the default state.  `closed` is an end state.

I think it would be reasonable to not use `0` as a value at all to make sure things are always explicit.  To do this, replace `iota` with `iota + 1`

x/deployment/types/msgs.go
boz: great then use

```go
// msgs.go
typeMsgCreateDeployment = `MsgCreateDeployment`

// codec.go
cdc.RegisterConcrete(MsgCreateDeployment{}, ModuleName + "/" + typeMsgCreateDeployment, nil)
```

I feel like the request was pretty obvious, I don't know why we're still talking about it.

boz:
I think it'd be best to re-start this PR - create a fresh branch and include only the changes necessary for one isolated change - "support using `app.App` in simulation tests", for instance.

x/deployment/types/msgs.go
fedekunze: Events have different formats (they use snake case) than amino registration, hence the difference. 

> I feel like the request was pretty obvious, I don't know why we're still talking about it.

I totally agree, idk why are you still nitpicking on these small things...

x/market/types/id.go
fedekunze: I respectfully disagree, not having any validation or the lack of testing leads to bugs. The same applies for private functions and vars, they don't provide any extra security guarantee if you define the public functions within the scope of the module.

x/deployment/types/msgs.go
boz: > I totally agree, idk why are you still nitpicking on these small things...

because you resolved the conversation without saying why you didn't address it.

app/app.go
fedekunze: > Exporting it as a function prevents other modules from re-assigning it.

modules don't have access to other modules, nor the module manager or other application-level fields unless explicitly given as arguments to the keepers.

x/market/types/id.go
boz: > not having any validation or the lack of testing leads to bugs

That's why we have a task to do those things.

Those things are, of course, completely separate from the topic of the benefits of abstraction, unless you count the benefits of re-using already-tested code (when `OrderID.Validate()` is tested, `BidID.Validate()` would use that tested code).

> The same applies for private functions and vars, they don't provide any extra security guarantee if you define the public functions within the scope of the module.

I don't really know what you mean by that last part, but keeping the surface area of a package to a minimum interface is for abstraction (internal implementation can change without affecting users of the package), isolation (prevents and/or minimizes paths to state mutation), and yes, security (vars can be re-assigned).  This is software engineering 101 stuff.

app/app.go
boz: sorry, I meant packages as in golang packages.

The fact of the matter is that there was absolutely no reason to change this.  The same is true for most of this PR.  I won't approve any of these (not necessarily) stylistic changes.

x/deployment/types/msgs.go
boz: Can you explain why the snake case is necessary?  The [documentation](https://github.com/cosmos/cosmos-sdk/blob/6a9c8bcb781c10ebe27beea75dfa49e19c94add9/types/tx_msg.go#L17-L18) for this method says:

> Returns a human-readable string for the message, intended for utilization within tags

It doesn't say anything about snake case.

PR #493, opened 2020-04-23
Add swagger docs for rest cli

akhilkumarpilli:
Fixes ovrclk/docs#55


akhilkumarpilli:
Swagger docs can be accesed by starting lcd server. So, I did it in akash instead of docs repo.

anilCSE:
> Swagger docs can be accesed by starting lcd server. So, I did it in akash instead of docs repo as I need to edit lcd routes.

Makes sense. So we can host swagger by running the lcd directly.

PR #496, opened 2020-04-24
Update Makefile dependencies

Ropes:
Unknown history of these test requirements, however currently tests are
failing due to missing the _build/akash[d] binaries.

This change adds the Make `image-bins` step to all of the targets which
run tests. Locally this has corrected the reproduced test failures we're
seeing on BuildKite.


Ropes:
@boz when you have a chance, this get's us back to green with Buildkite.

PR #522, opened 2020-05-05
feat: ci - reduce image-bin invocations

boz:
Improve CI build times

* remove unnecessary image-bin builds.
* remove unused targets

### DIFF
--- a/.buildkite/script/test.sh
+++ b/.buildkite/script/test.sh
@@ -12,6 +12,11 @@ do_bins(){
   make bins
 }
 
+do_image_bins(){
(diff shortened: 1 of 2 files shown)


.buildkite/script/test.sh
```diff
@@ -12,6 +12,11 @@ do_bins(){
   make bins
 }
 
+do_image_bins(){
+  echo "--- :hammer_and_pick: building image binaries"
```
Ropes, line 16: :hammer_and_pick: nice!

PR #529, opened 2020-05-06
Lite Akash node and CLI example

Ropes:
This change rebuilds the original `demo` project for an example which is more lightweight and accessible for learning the Akash network tools and structure.

### Features
* Does *not* require Docker, k8s, or helm
* Only requires building the `akashd` & `akashctl` binaries
* README documents the functionality
* Makefile defines the normal actions
  * Initializes and cleans up the environment
  * Smooths out repetitive command parameters
* Small `akashctl` documentation fixes


gosuri:
👍

Ropes:
Hmm, I'm not sure what to do about the `@codecov
codecov/patch — 0.00% of diff hit (target 20.79%)` build failure...

_run/lite/Makefile
```diff
@@ -0,0 +1,100 @@
+include ../common.mk
+
+KEY_OPTS       = --keyring-backend=test
+CHAIN_NAME     = local
+CHAIN_OPTS     = --chain-id $(CHAIN_NAME)
+akashctl       = ../../akashctl --home $(CLIENT_HOME) 
+akashd         = ../../akashd --home $(NODE_HOME)
```
boz, line 7: can use `$(AKASHCTL)`, `$(AKASHD)`

boz: fwiw it _is_ nice to have these in `env.sh` so you can source it from the shell for manual interactions.

boz:
@Ropes don't worry about the codecov failure - it doesn't block the build right now.

_run/lite/Makefile
Ropes: Completely agree, thanks for the reminder! I forgot we had those defined in the common.mk.

Issue #481, opened 2020-04-07
provider cluster: kubernetes CRDs


boz:
* Update the CRD [types](https://github.com/ovrclk/akash/blob/0bfe385b1f20dcebf489f4759942d36611c78a9a/pkg/apis/akash.network/v1/types.go) so that they're compatible with our new type design.
* [re-generate](https://github.com/ovrclk/akash/blob/0bfe385b1f20dcebf489f4759942d36611c78a9a/Makefile#L102-L105) the [CRD clients](https://github.com/ovrclk/akash/tree/0bfe385b1f20dcebf489f4759942d36611c78a9a/pkg/client).

Ropes:
Using a checkout of k8s.io/code-generator:v0.18.2, I've been able to generate all the `akash` clientset and integration packages. `akash` and `akashd` build fine, so now we primarily need to sort out our k8s dependency woes. There are some vscode flagged error messages which I'm a little confused by, but they might be vscode being confused.

I still need to go back and update the CRD types, but now I have a path see results of a build.

hydrogen18:
I think this can be considered done?

PR #553, opened 2020-05-07
CRD Types Updated, use `v1` APIs with context passed

Ropes:
pkg/apis/akash.network/v1

This will make future upgrades cleaner and consolidate APIs
used. Regenerated the deepcopy code with `make kubetypes`. There were
few changes, so at first it was unclear if the kube types generation was
actually functioning.

- Correcting LeaseID type passage through CRD
- LeaseID has to have some type cohersion to be passed through CRD Akash
- manifests.
- CRD Types use `v1` APIs
- This will hopefully make future upgrades cleaner and consolidate APIs
- used.
- deepgen tags updated and CRDs regenerated
- This change did pick up LeaseID changes.
- Additional testing, documentation of Manifest type
- Correcting Makefile and reference to use akashctl
- Minikube ip address configured via variable
- Context parameterized, stored on the Client struct

Fixes TODOs to accept context parameters needed for the new k8s API
calls. For some cases storing the context on the Client struct for now
until a proper shutdown hook is decided.

Signed-off-by: Joshua Roppo <josh@akash.network>

Implements: https://github.com/ovrclk/akash/issues/481

Full implementation of the k8s provider still needed, plus testing framework.


provider/cluster/kube/client.go
```diff
@@ -114,9 +118,7 @@ func (c *client) shouldExpose(expose *manifest.ServiceExpose) bool {
 }
 
 func (c *client) Deployments() ([]cluster.Deployment, error) {
-	// TODO: accept context as parameter
-	ctx := context.Background()
-	manifests, err := c.mc.AkashV1().Manifests(c.ns).List(ctx, metav1.ListOptions{})
+	manifests, err := c.mc.AkashV1().Manifests(c.ns).List(c.ctx, metav1.ListOptions{})
```
boz, line 121: Let's pass `context.Context` as an argument to these methods instead of saving it on `client`.

Ropes: :+1:

Ropes: :+1:

Ropes: @boz, so updating all of the functions to accept `ctx` results in the [Client interface](https://github.com/ovrclk/akash/blob/65bac28086e0e1a196897e39c05d7df675696cc7/provider/cluster/client.go#L22) needing to be updated. Do we want to go that far? That's a larger API change, and I'm less sure how far reaching it will be.

boz: Yeah let's definitely do it, but as a follow-on PR/issue.

PR #574, opened 2020-05-12
refactor: Client interface accepts Context params

### LINKED ISSUE #555: refactor: Client interface functions accept Context parameters
Per the discussion on the [CRD PR](https://github.com/ovrclk/akash/pull/553#pullrequestreview-407902957) it would be ideal if the provider `Client` interface accepted `ctx context.Context` as their first parameter instead of embedding on the `kube.client` implementation. 

For consistency each of these functions should accept a `context.Context` as the first function parameter to signal safe shutdown behavior.
```go
type Client interface {
	Deploy(mtypes.LeaseID, *manifest.Group) error
	TeardownLease(mtypes.LeaseID) error
	Deployments() ([]Deployment, error)
	LeaseStatus(mtypes.LeaseID) (*LeaseStatus, error)
	ServiceStatus(mtypes.LeaseID, string) (*ServiceStatus, error)
	ServiceLogs(context.Context, mtypes.LeaseID, int64, bool) ([]*ServiceLog, error)
	Inventory() ([]Node, error)
}
```

And all downstream implementations will need updating.

Ropes:
Kube provider client now accepts contexts

Removed struct-stored context.Context from k8s client{}. Now all functions
accept the caller's Context to manage shutdown behavior.

fixes #555

Signed-off-by: Joshua Roppo <joshroppo@gmail.com>

### DIFF
--- a/provider/cluster/inventory.go
+++ b/provider/cluster/inventory.go
@@ -170,6 +170,7 @@ type inventoryResponse struct {
 func (is *inventoryService) run(reservations []*reservation) {
 	defer is.lc.ShutdownCompleted()
 	defer is.sub.Close()
+	ctx, cancel := context.WithCancel(context.Background())
 
 	t := time.NewTimer(time.Hour)
 	t.Stop()
@@ -177,7 +178,7 @@ func (is *inventoryService) run(reservations []*reservation) {
 
 	var inventory []Node
 	ready := false
-	runch := is.runCheck()
+	runch := is.runCheck(ctx)
 
 	var fetchCount uint
 
@@ -273,7 +274,7 @@ loop:
 			// run cluster inventory check
 
 			t.Stop()
-			runch = is.runCheck()
+			runch = is.runCheck(ctx)
 
 		case res := <-runch:
 			// inventory check returned
@@ -307,15 +308,16 @@ loop:
 			fetchCount++
 		}
 	}
+	cancel()
 
 	if runch != nil {
 		<-runch
 	}
 }
 
-func (is *inventoryService) runCheck() <-chan runner.Result {
+func (is *inventoryService) runCheck(ctx context.Context) <-chan runner.Result {
 	return runner.Do(func() runner.Result {
-		return runner.NewResult(is.client.Inventory())
+		return runner.NewResult(is.client.Inventory(ctx))
 	})
 }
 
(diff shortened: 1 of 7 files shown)


provider/cluster/inventory.go
```diff
@@ -315,7 +315,9 @@ loop:
 
 func (is *inventoryService) runCheck() <-chan runner.Result {
 	return runner.Do(func() runner.Result {
-		return runner.NewResult(is.client.Inventory())
+		// TODO: Handle shutdown via context.Context.
```
boz, line 318: can make one based off of `is.lc.ShuttingDown()` or just from a `cancel()` in the mainloop.

provider/cluster/manager.go
```diff
@@ -194,11 +195,13 @@ func (dm *deploymentManager) startTeardown() <-chan error {
 }
 
 func (dm *deploymentManager) doDeploy() error {
-	return dm.client.Deploy(dm.lease, dm.mgroup)
+	ctx := context.Background() // TODO: manage contexted within the  deploymentManager
```
boz, line 198: same as before - in `run()`:

```go
ctx, cancel := context.WithCancel(context.Background())
// ...
loop:
for {
// ...
}
cancel()
// ...

Comment (Ropes) [reply provider/cluster/manager.go:198]: I started going down that path, between the anonymous functions being initialized in goroutines which would need to take the context as a parameter, and then the builder functions which would also need to take context as a parameter, the state managing functions would need to take context as a parameter... all of this makes for tightly coupled and brittle code, which is not what this issue was about. I think we should punt on updating `deploymentManager` in this changeset.

Comment (boz) [reply provider/cluster/manager.go]: agreed - was a suggestion for next round.

File: provider/cluster/inventory.go [hunk 3]
```diff
@@ -170,14 +170,15 @@ type inventoryResponse struct {
 func (is *inventoryService) run(reservations []*reservation) {
 	defer is.lc.ShutdownCompleted()
 	defer is.sub.Close()
+	ctx, can := context.WithCancel(context.Background())
```
boz, line 173: `can` -> `cancel` pls.

provider/cluster/manager.go
boz: fwiw it'd look something like [this](https://github.com/ovrclk/akash/compare/cluster-manager-ctx-example#diff-eebca299d9c7076c0017fb68023f0ca4)

boz: also agree that there's probably better ways to structure that manager.

PR #576, opened 2020-05-13
WIP feat: rest api for provider daemon

troian:
implement status endpoint


.golangci.yaml
```diff
@@ -1,6 +1,12 @@
 linters:
   enable-all: true
 
+# all available settings of specific linters
+linters-settings:
+  govet:
+    # report about shadowed variables
+    check-shadowing: true
```
boz, line 8: eh, I'd like to see what it complains about.  I prefer shadowed variables to contrived names.

cmd/akashctl/provider.go
```diff
@@ -62,34 +86,117 @@ func providerCmd(cdc *codec.Codec) *cobra.Command {
 				return err
 			}
 
-			log := log.NewTMLogger(log.NewSyncWriter(os.Stdout))
+			lg := log.NewTMLogger(log.NewSyncWriter(os.Stdout))
```
boz, line 89: this is fine to shadow for instance.

cmd/akashctl/provider.go
```diff
@@ -62,34 +86,117 @@ func providerCmd(cdc *codec.Codec) *cobra.Command {
...
 
-			session := session.New(log, aclient, pinfo)
+			ses := session.New(lg, aclient, pinfo)
```
boz, line 91: as is this.

cmd/akashctl/provider.go
```diff
@@ -24,8 +30,26 @@ import (
 	pmodule "github.com/ovrclk/akash/x/provider"
 )
 
-var _ = providerCmd(codec.New()) // Disables deadcode lint check until provderCmd is ready for use.
+type serviceRestConfig struct {
```
boz, line 33: would want to put all of this in `provider/rest` or something.

cmd/akashctl/provider.go
```diff
@@ -62,34 +86,117 @@ func providerCmd(cdc *codec.Codec) *cobra.Command {
...
+				return err
+			}
 
-			service, err := provider.NewService(ctx, session, bus, cclient)
-			if err != nil {
+			var rest *serviceRest
+			if rest, err = newServiceRest(serviceRestConfig{
+				ctx:           ctx,
+				log:           lg,
+				serviceStatus: service,
```
boz, line 110: would want to pass `service` into this so that it can call, say, `service.Status()`.

cmd/akashctl/provider.go
```diff
@@ -62,34 +86,117 @@ func providerCmd(cdc *codec.Codec) *cobra.Command {
...
+			}); err != nil {
+				_ = service.Close()
 				return err
 			}
 
 			<-service.Done()
 
+			rest.shutdown()
```
boz, line 118: we'd want to shut everything down if the this exits early.

boz:
implemented in #617

PR #580, opened 2020-05-13
fix: updated code using unsafe package

### LINKED ISSUE #550: lint: fix gosec
* Fix `gosec` warnings from the [lint checker](https://github.com/ovrclk/akash/blob/master/.github/workflows/tests.yml#L39) step for it.
 * Remove `continue-on-error: true` from `lint gosec` step.

Ropes:
The usage of sha1 was removed and updated to sha256. The use of `sha*`
is minor compared to the changes which fix the functionality of creating
unique k8s object names based on the LeaseID.

Added test to assert that the name is properly passed into the k8s
type.

fixes #550

### DIFF
--- a/provider/cluster/kube/builder.go
+++ b/provider/cluster/kube/builder.go
@@ -1,7 +1,7 @@
 package kube
 
 import (
-	"crypto/sha1"
+	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
 	"strconv"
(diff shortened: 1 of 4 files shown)


provider/cluster/kube/builder.go
```diff
@@ -288,18 +288,18 @@ func exposeExternalPort(expose *manifest.ServiceExpose) int32 {
...
 }
 
+// lidNS generates a unique sha256 sum for identifying a provider's object name.
 func lidNS(lid mtypes.LeaseID) string {
-	// TODO
-	var path string
-	// path := lid.String()
-	sha := sha1.Sum([]byte(path))
+	path := mtypes.BidIDString(lid.BidID())
+	sha := sha256.Sum256([]byte(path))
```
boz, line 294: might run into length restrictions here (and in a lot of places), but this is necessary to expose - we need a compact representation for these things.

This doesn't need to be cryptographically secure - using a v5 uuid migtht be a good option

provider/cluster/kube/builder_test.go
```diff
@@ -0,0 +1,62 @@
...
+	testing.TB
+}
+
+func (tw testWriter) Write(p []byte) (int, error) {
+	tw.Helper()
+	tw.Logf("%s", p)
+	return len(p), nil
+}
+
+func TestLidNsSanity(t *testing.T) {
```
boz, line 31: :+1:

provider/cluster/kube/builder.go
Ropes: Yeah, length is something I wasn't sure about. I added a test to log the length of encoded bytes. hex256 pushes it up to 64.

```
=== RUN   TestShaLengths
    TestShaLengths: builder_test.go:88: sha1: 20 sha256: 32
    TestShaLengths: builder_test.go:91: hex1: 40 hex256: 64
```

PR #582, opened 2020-05-13
fixes: unused variable cleanup

### LINKED ISSUE #548: lint: fix staccheck
* Fix `staticcheck` warnings from the [lint checker](https://github.com/ovrclk/akash/blob/master/.github/workflows/tests.yml#L39) step for it.
 * Remove `continue-on-error: true` from `lint staticcheck` step.

Ropes:
ci-lint detected unused variables returned from function calls and type
comparison. All minor changes.

fixes #548


sdl/units.go
```diff
@@ -51,8 +52,8 @@ func (u *cpuQuantity) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		if err != nil {
 			return err
 		}
-		if val < 0 {
-			return errNegativeValue
+		if val == 0 {
```
boz, line 55: I think `0` is okay here.

x/deployment/query/path.go
```diff
@@ -90,6 +90,9 @@ func ParseGroupPath(parts []string) (types.GroupID, error) {
 	}
 
 	gseq, err := strconv.ParseUint(parts[2], 10, 32)
+	if err != nil {
```
boz, line 93: damn how did these get missed.  thanks!

sdl/units.go
Ropes: fixed

sdl/units.go
```diff
@@ -51,9 +51,6 @@ func (u *cpuQuantity) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		if err != nil {
 			return err
 		}
-		if val < 0 {
-			return errNegativeValue
```
boz, line 55: `errNegativeValue` can probably be removed.

Ropes: It's still used below when parsing a float, I think it's fine to leave for now.

PR #583, opened 2020-05-14
fix: preallocate slices

### LINKED ISSUE #552: lint: fix prealloc
* Fix `prealloc` warnings from the [lint checker](https://github.com/ovrclk/akash/blob/master/.github/workflows/tests.yml#L39) step for it.
 * Remove `continue-on-error: true` from `lint prealloc` step.

Ropes:
Small change to fix lint preallocation of types instead of depending on
`append(...)` to expand the slice. `make(<T>, 0, capacity)` defined in
slice allocation where there is an expected size to use.

fixes #552


provider/cluster/kube/client.go
```diff
@@ -122,8 +122,7 @@ func (c *client) Deployments(ctx context.Context) ([]cluster.Deployment, error)
 		return nil, err
 	}
 
-	var deployments []cluster.Deployment
-
+	deployments := make([]cluster.Deployment, len(manifests.Items))
```
boz, line 125: this should use `make([]cluster.Deployment, 0, len(manifests.Items))`

Ropes: good catch, thanks!

PR #592, opened 2020-05-15
lint: unparam cleanup of unused varaibles

### LINKED ISSUE #547: lint: fix unparam
* Fix `unparam` warnings from the [lint checker](https://github.com/ovrclk/akash/blob/master/.github/workflows/tests.yml#L39) step for it.
 * Remove `continue-on-error: true` from `lint unparam` step.

Ropes:
'_'ing function parameters which are not used, updating some CLI
commands to use their 'key' parameters to enhance Cobra descriptions, no
changes to interfaces or function calls.

Used `// nolint` to silence all lint violations which were caused by
functions returning values which were zero-value or errors that were
always nil. Fixing would require chaning methods which might need to be
changed back later.

fixes #547

### DIFF
--- a/provider/cluster/kube/builder.go
+++ b/provider/cluster/kube/builder.go
@@ -53,7 +53,7 @@ func (b *nsBuilder) name() string {
 	return b.ns()
 }
 
-func (b *nsBuilder) create() (*corev1.Namespace, error) {
+func (b *nsBuilder) create() (*corev1.Namespace, error) { // nolint:golint,unparam
 	return &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:   b.ns(),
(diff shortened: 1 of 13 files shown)


provider/cluster/kube/builder.go
```diff
@@ -62,7 +62,7 @@ func (b *nsBuilder) create() (*corev1.Namespace, error) {
 	}, nil
 }
 
-func (b *nsBuilder) update(obj *corev1.Namespace) (*corev1.Namespace, error) {
+func (b *nsBuilder) update(obj *corev1.Namespace) (*corev1.Namespace, error) { // nolint
```
troian, line 65: maybe suppress a particular linter instead of all?

Ropes: Thanks for the  idea, I didn't know that was a feature! Implemented.

Issue #455, opened 2020-03-10
_run{single, multi, kube}: migrate run scripts


gosuri:
migrate run scripts under _run to reflect latest changes

Ropes:
This issue has been primarily completed. *Multi* is till in need of being updated, however `lite` and `kube` are functional to the point of Provider  New issue for `multi`: https://github.com/ovrclk/akash/issues/627

Issue #571, opened 2020-05-12
x/: improve validation by parsing addresses


boz:
In many address validations, we only check for non-empty addresses instead of validating that they are well formed.

[example](https://github.com/ovrclk/akash/blob/master/x/provider/types/msgs.go#L24):

```go
	if msg.Owner.Empty() {
		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, "MsgCreate: Invalid Provider Address")
	}
```

This should be
```go
	if _, err := sdk.AccAddressFromBech32(msg.Owner); err != nil {
		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, "MsgCreate: Invalid Provider Address")
	}
```

RiccardoM:
@boz I would like to help you with this issue, however I've just checked into the current `master` code and it appears to me that the only places where the `.Empty()` form is used is when dealing with messages. 

As far as I know, an `sdk.AccAddress` object can represent either: 

1. an empty address; 
2. a valid address. 

If a user tries to send a `Msg` instance with an invalid address inside, this will not be properly parsed into an `sdk.AccAddress` object, thus failing before reaching the point in which it can be checked for emptiness. 

Also, the second code snippet should be: 

```golang
if _, err := sdk.AccAddressFromBech32(msg.Owner.String()); err != nil {
	return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, "MsgCreate: Invalid Provider Address")
}
```

Since the `AccAddressFromBech32` method accepts a `string` and not an `sdk.AccAddress` object. 

This would be pointless as the `String` method only works if the address was valid in the first place. 

In the end, I think you can close this issue safely and keep using the `Empty()` method if you are dealing with `sdk.AccAddress` objects. Otherwise, if you are dealing with addresses in the `string` format, you need to use the `AccAddressFromBech32` method.

boz:
I see, thanks @RiccardoM.  I didn't look into this enough - `sdk.VerifyAddressFormat()` is more appropriate.

It does indeed look like `sdk.VerifyAddressFormat()` is used in `AccAddress.UnmarshalJSON()`, but given that `cosmos-sdk` is moving towards `protobuf`, I think it'd be nice to use `sdk.VerifyAddressFormat(addr)` instead of `addr.Empty()`.

akhilkumarpilli:
[...]

PR #612, opened 2020-05-21
feature: Update 'kube' example for akash network

Ropes:
This change mostly recreates the `_run/kube` example environment. I'd say it's worth commiting at this point while continuing to debug the manifests being applied to cluster issues. It puts some guardrails around invalid YAML being passed to the manifest parser and resulting in a red herring error reported back from Tendermint.

See [README](https://github.com/ovrclk/akash/pull/612/files#diff-685af98dae699e18a60e10fee93c181b) for what's implemented with hopefully useful example documentation.

* Makefile to execute akash* built binaries with consistency.
`akashctl provider` command re-enabled.
* Debugging and testing code around read provider config file.
* Functional `...tx send...<from> <to> <value>` command added as
`send-to-main`. Additional `query-txn`
* README added. Functional querying of market information.
* Makefile to execute akash* built binaries with consistency.
`akashctl provider` command re-enabled.
* Removing more differences betweene lite and kube _run scripts
* Identified Provider creation missing validation
* Fixed: Empty kv pairs could be applied to Provider Create and Update Messages.
This results in a Tendermint error, which is a big red herring to the
actual problem. Discovered via YAML configuration being parsed as empty.
* Unit tests added to verify the `provider` module's Msg validation
code.
* `make run-provider` command added.
*Documenting running Provider and Bidding in README.md

references #455

TODO:
* [ ] Still debugging Provider<->Minikube integration.


_run/kube/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+__t1__: Create Deployment
+```sh
+$ make deploy
+```
+Then view the order created from the deployment via `make query-market` or in __t3__. There will be a new Order created.
+
+Bid on the new Order by referencing its sequence IDs:
+```sh
+../../akashctl --home cache/client query market order list
```
boz, line 86: this is why `env.sh` & co is so nice to have.

x/provider/types/msgs.go
```diff
@@ -30,6 +31,9 @@ func (msg MsgCreateProvider) ValidateBasic() error {
 	if msg.Owner.Empty() {
 		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, "MsgCreate: Invalid Provider Address")
 	}
+	if len(msg.Attributes) == 0 {
```
boz, line 34: It's okay for providers to not have any attributes.

x/provider/types/msgs_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/tendermint/tendermint/libs/kv"
+)
+
+func TestConfigPath(t *testing.T) {
+	tests := []struct {
+		path   string
+		expErr error
+	}{
+		{
+			path:   "/home/ropes/go/src/github.com/ovrclk/akash/_run/kube/provider.yml",
```
boz, line 19: this can just be `"/foo"`

Also try:
| value | error? |
|---|---|
|`localhost`|true|
|`localhost/foo`|true|
|`localhost:80`|true|
|`localhost:80/foo`|true|
|`127.0.0.1`|true|
|`127.0.0.1/foo`|true|
|`127.0.0.1:80`|true|
|`127.0.0.1:80/foo`|true|
|`http://localhost`|false|
|`http://localhost/foo`|false|
|`http://localhost:80`|false|
|`http://localhost:80/foo`|false|
|`https://localhost:80`|false|
|`https://localhost:80/foo`|false|

Probably need to audit how this is used/parsed and make sure the validation makes sense.  I think that allowing schemes other than `http(s)` probably doesn't make sense.  Can be done later.

x/provider/types/msgs_test.go
```diff
@@ -0,0 +1,168 @@
+package types
```
boz, line 1: **thank you**

x/provider/types/msgs_test.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+var msgCreateTests = []struct {
+	msg    Provider
+	expErr error
+	delErr error
+}{
+	{
+		msg: Provider{
+			Owner:   sdk.AccAddress("hihi"),
```
boz, line 45: this should be disallowed by #571.  can change it with the fix for it tho.

x/provider/types/msgs_test.go
```diff
@@ -0,0 +1,168 @@
...
+			HostURI: "http://localhost:3001/",
+		},
+		expErr: ErrAttributes,
+	},
+	{
+		msg: Provider{
+			Owner:   sdk.AccAddress("hihi"),
+			HostURI: "ht tp://foo.com",
```
boz, line 80: :+1:

x/provider/types/msgs_test.go
boz: might as well throw some ipv6 failing tests in there as well so we don't have to deal with it.

_run/kube/README.md
Ropes: That was just output from the `make` command, it handles the weirdness, but yeah, not super pretty.

x/provider/types/msgs.go
Ropes: Is that something we might want to consider change/enforce a spec down the road? If a Provider has no attributes, then it has nothing to match/filter on; total wildcard. Or am I missing some use case?

_run/kube/README.md
boz: It sucks for tinkering.

x/provider/types/msgs.go
boz: As a tenant, I can declare the attributes that I want providers that bid on my orders to have.

PR #620, opened 2020-05-28
refactor: DeploymentManager state machine

Ropes:
* Deployment state managed with function types via the Rob Pike method of state machines.
the deploymentManager is refactored into deployState.
* Unit Tests will eventually be able to check the state of the
machine after submitting messages.
* Context added after rebase.

### Testing

Usually this pattern helps creating deterministic tests, and evaluation of state achieved given envents XYZ, however the tight coupling of 'service's makes testing based on events very difficult to initialize.

* Prototyped deployment state testing.
* Currently the test setup panics due to complex initializers, so disabled.

### DIFF
--- a/provider/cluster/service.go
+++ b/provider/cluster/service.go
@@ -87,6 +87,8 @@ func NewService(ctx context.Context, session session.Session, bus pubsub.Bus, cl
 	return s, nil
 }
 
+var _ Service = (*service)(nil)
+
 type service struct {
 	session session.Session
 	client  Client
@@ -220,7 +222,7 @@ loop:
 					"lease", dm.lease, "group-name", dm.mgroup.Name)
 			}
 
-			// todo: unreserve resources
+			// TODO: unreserve resources
 
 			delete(s.managers, mquery.LeasePath(dm.lease))
 		}
@@ -232,6 +234,7 @@ loop:
 		if manager != nil {
 			manager := <-s.managerch
 			s.log.Debug("manager done", "lease", manager.lease)
+			// TODO: ??? shutdown each manager?
 		}
 	}
 
--- a/provider/cluster/state.go
+++ b/provider/cluster/state.go
@@ -1,0 +1,188 @@
+package cluster
+
+import (
+	"context"
+	"sync"
+
+	"github.com/ovrclk/akash/manifest"
+	"github.com/ovrclk/akash/provider/session"
+	"github.com/ovrclk/akash/pubsub"
+	mtypes "github.com/ovrclk/akash/x/market/types"
+	"github.com/tendermint/tendermint/libs/log"
+)
+
+type stateFunc func(*deployState) stateFunc
+
+// deployState manages the a Lease's ManifestGroup being deployed
+// to a provider's client.
+type deployState struct {
+	ctx   context.Context
+	state stateFunc
+	mux   sync.Mutex
+
+	// Unused fields
+	bus     pubsub.Bus
+	session session.Session
+
+	client  Client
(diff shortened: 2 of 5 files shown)


provider/cluster/service.go
```diff
@@ -232,6 +234,7 @@ loop:
 		if manager != nil {
 			manager := <-s.managerch
 			s.log.Debug("manager done", "lease", manager.lease)
+			// TODO: ??? shutdown each manager?
```
boz, line 237: this is "drain mode" - it waits for all of the resources it created to complete.

the managers are signaled to shutdown automatically by [this](https://github.com/ovrclk/akash/blob/e126a9d1838a8b369221188e97b5b895239d87fe/provider/cluster/manager.go#L64)

provider/cluster/state.go
```diff
@@ -0,0 +1,188 @@
...
+func stateDeploying(ds *deployState) stateFunc {
+	if ds.monitor != nil {
+		ds.monitor.shutdown() //
+		ds.monitor = nil
+	}
+	// TODO: Pass ds context to Deploy()
+	err := ds.client.Deploy(ds.ctx, ds.lease, ds.mgroup)
+	if err != nil {
+		ds.log.Error("deploying error", err)
+		if ds.errch != nil {
```
boz, line 142: race condition.  consider this series of events:

1. `newDeployState()` -> `stateDeploying`.
2. while `ds.client.Deploy()` is executing, someone calls [`update()`](https://github.com/ovrclk/akash/pull/620/files#diff-3508d8a34017e711c0d9bc5b341e2febR101) or [`teardown()`](https://github.com/ovrclk/akash/pull/620/files#diff-3508d8a34017e711c0d9bc5b341e2febR110).

provider/cluster/state.go
```diff
@@ -0,0 +1,188 @@
...
+
+// stateErrored engages when an uncrecoverable point has been reached.
+func stateErrored(ds *deployState) stateFunc {
+	ds.errch = nil // reset the error channel
+	// TODO: Determine if there are other steps to take in an errored state.
+	return stateNominal
+}
+
+// stateTearingdown handles safe shutdown.
+func stateTearingdown(ds *deployState) stateFunc {
```
boz, line 174: what happens when

1. `shutdown()` is called
2. `update()` is called

It looks like `update()` will hang forever.  there are more instances of this issue.

provider/cluster/state.go
```diff
@@ -0,0 +1,188 @@
...
+		ds.monitor = nil
+	}
+	// TODO: Pass ds context to Deploy()
+	err := ds.client.Deploy(ds.ctx, ds.lease, ds.mgroup)
+	if err != nil {
+		ds.log.Error("deploying error", err)
+		if ds.errch != nil {
+			ds.errch <- err
+		}
+		return stateErrored
```
boz, line 145: deadlock if someone calls `update()`, `teardown()`

provider/cluster/state.go
```diff
@@ -0,0 +1,188 @@
...
+	case <-ds.ctx.Done():
+		return stateTearingdown
+	case <-ds.teardownch:
+		return stateTearingdown
+	}
+}
+
+// stateErrored engages when an uncrecoverable point has been reached.
+func stateErrored(ds *deployState) stateFunc {
+	ds.errch = nil // reset the error channel
```
boz, line 168: race condition.  consider this series of events:

1. `update()` is called, the update fails and state gets set to `stateErrored`.
2. as `stateErrored` begins executing, someone calls `update()` or `teardown()`.

PR #621, opened 2020-05-28
refactor: replaces events type const

### LINKED ISSUE #607: events: replace event type
We are currently emitting evens with type of [`sdk.EventTypeMessage`](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/x/deployment/types/event.go#L26).

These appear to be use for actual transaction messages.  It looks like other projects use their own name for the type; we should too.  `akash` might be a reasonable name for our message types, not sure.

All of the generated events in `x/*/types/event.go` would need to be modified as well as the processing that's done in [`sdkutil`](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/sdkutil/event.go#L30)

Ropes:
* Previously used the sdk.EventTypeMessage == "message" string to
identify all of our messages. This provided little value.
* Updated all ephemeral messages to be prefixed with
"sdkutil.EventTypeMessage" == "akash.v1".
* Added tests to assert successful event parsing, though they just check 
there are no parsing errors right now.

fixes #607

### DIFF
--- a/sdkutil/event.go
+++ b/sdkutil/event.go
@@ -7,6 +7,13 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
+const (
+	akashEventMessageV1 = "akash.v1"
+
+	// EventTypeMessage defines the Akash message string
+	EventTypeMessage = akashEventMessageV1
+)
+
 var (
 	// ErrNotFound is the error with message "Not found"
 	ErrNotFound = errors.New("Not found")
--- a/x/deployment/types/events_test.go
+++ b/x/deployment/types/events_test.go
@@ -1,0 +1,187 @@
+package types
+
+import (
+	"fmt"
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/pkg/errors"
+
+	"github.com/ovrclk/akash/sdkutil"
+)
+
+var (
+	keyAcc, _ = sdk.AccAddressFromBech32("akash1qtqpdszzakz7ugkey7ka2cmss95z26ygar2mgr")
+	//keyParams = sdk.NewKVStoreKey(params.StoreKey)
+
+	errWildcard = errors.New("wildcard string error can't be matched")
+)
+
+type testEventParsing struct {
+	msg    sdkutil.Event
+	expErr error
+}
+
+func (tep testEventParsing) testMessageType() func(t *testing.T) {
+	_, err := ParseEvent(tep.msg)
+	return func(t *testing.T) {
+		t.Logf("ERR: %v", err)
+		// expected error doesn't match returned    || error returned but not expected
+		if (tep.expErr != nil && errors.Is(err, tep.expErr)) || (err != nil && tep.expErr == nil) {
+			// if the error expected is errWildcard to catch untyped errors, don't fail the test, the error was expected.
+			if errors.Is(tep.expErr, errWildcard) {
+				t.Errorf("unexpected error: %v exp: %v", err, tep.expErr)
+				t.Logf("%T %v", errors.Cause(err), err)
+				t.Logf("%+v", tep)
+			}
+		}
+	}
+}
+
+var TEPS = []testEventParsing{
+	{
+		msg: sdkutil.Event{
+			Type: "nil",
+		},
+		expErr: sdkutil.ErrUnknownType,
+	},
+	{
+		msg: sdkutil.Event{
+			Type: sdkutil.EventTypeMessage,
+		},
+		expErr: sdkutil.ErrUnknownModule,
+	},
+
+	{
+		msg: sdkutil.Event{
+			Type:   sdkutil.EventTypeMessage,
+			Module: ModuleName,
+		},
+		expErr: sdkutil.ErrUnknownAction,
+	},
+	{
+		msg: sdkutil.Event{
+			Type:   sdkutil.EventTypeMessage,
+			Module: "nil",
+		},
+		expErr: sdkutil.ErrUnknownModule,
+	},
+
+	{
+		msg: sdkutil.Event{
+			Type:   sdkutil.EventTypeMessage,
+			Module: ModuleName,
+			Action: "nil",
+		},
+		expErr: sdkutil.ErrUnknownAction,
+	},
+
+	{
+		msg: sdkutil.Event{
+			Type:   sdkutil.EventTypeMessage,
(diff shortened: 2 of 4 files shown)


x/deployment/types/events_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth"
+	"github.com/pkg/errors"
+
+	"github.com/ovrclk/akash/sdkutil"
+)
+
+var (
+	keyAcc = sdk.NewKVStoreKey(auth.StoreKey)
```
jackzampolin, line 15: Interested to know why you are using a `KVStore` key name here?

Ropes: copy-pasta from other test utils for an address string... no good reason.

Have a better suggestion?

sdkutil/event.go
```diff
@@ -16,6 +18,9 @@ var (
 	ErrUnknownModule = errors.New("Unknown module")
 	// ErrUnknownAction is the error with message "Unknown action"
 	ErrUnknownAction = errors.New("Unknown action")
+
+	// EventTypeMessage defines the Akash message string
+	EventTypeMessage = akashEventMessageV1
```
boz, line 23: any reason why this isn't a constant?

x/deployment/types/events_test.go
```diff
@@ -0,0 +1,174 @@
...
+		},
+		expErr: nil,
+	},
+
+	{
+		msg: sdkutil.Event{
+			Type:   sdkutil.EventTypeMessage,
+			Module: ModuleName,
+		},
+		expErr: nil,
```
boz, line 62: Looking at [this](https://github.com/ovrclk/akash/blob/e126a9d1838a8b369221188e97b5b895239d87fe/x/deployment/types/event.go#L116-L139), I believe these things should cause an error:

* `Action` is unpopulated
* `Action` is neither `evActionDeploymentCreate` nor `evActionDeploymentUpdate`
* `DeploymentID` attributes are not set.

x/deployment/types/events_test.go
```diff
@@ -0,0 +1,174 @@
...
+		},
+		expErr: errWildcard,
+	},
+}
+
+func TestEventParsing(t *testing.T) {
+	for i, test := range TEPS {
+		t.Run(fmt.Sprintf("%d", i), func(tep testEventParsing) func(t *testing.T) {
+			return func(t *testing.T) {
+				t.Run("parse-message-type", test.testMessageType())
```
boz, line 170: It's not clear to me why all these closures are needed?

```go
for i := range TEPS {
  t.Run("bla",func(t *testing.T){
      result, err := Parse(TEPS[i].msg)
      // ...
  })
}
```

you can also just add a message to the end of, say, `assert.Error(t,err,"blablabla")` - we use `github.com/stretchr/testify/assert`.

sdkutil/event.go
Ropes: It references the set constant. 

Reasoning: In the potential case we need to switch to `akash.v2` the `const` values don't need to change, we just change what `EventTypeMessage` references, and we can still reference current and prior const values within the package.

Not something I'm super concerned about, just trying to make potential future logic checks based on version possible. Since messages are ephemeral though... maybe not worth the effort?

x/deployment/types/events_test.go
Ropes: Yeah... I don't know how I missed that, and clearly the err assertion isn't working.

boz: I think **LAMBDA BOWL** is causing one of those annoying closure-in-a-loop bugs.

sdkutil/event.go
boz: I don't really understand.  Even if you want to keep the symbol name constant over upgrades, can't you do this?

```
const (
   foov1 = "foo.v1"
   FOO  = foov1
)
```

I'm always hesitant to use exported `var`s, and I don't understand what it buys us here.

Ropes: Fixed with const's

x/deployment/types/events_test.go
Ropes: Originally I was thinking if we wanted to have a step using the same test data, but different test label than `parse-message-type`, use the second layer of anonymous functions.
However this function being tested isn't worth additional separation. 

Removed.

boz: I believe this was the problem too @Ropes  - needed to use `TEPS[i]` instead of `test`

x/deployment/types/events_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/pkg/errors"
+
+	"github.com/ovrclk/akash/sdkutil"
+)
+
+var (
+	keyAcc, _ = sdk.AccAddressFromBech32("akash1qtqpdszzakz7ugkey7ka2cmss95z26ygar2mgr")
```
boz, line 14: good idea.  maybe can have a thing in testutil w/ like 100 hard-coded addresses.

x/deployment/types/events_test.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+type testEventParsing struct {
+	msg    sdkutil.Event
+	expErr error
+}
+
+func (tep testEventParsing) testMessageType() func(t *testing.T) {
+	_, err := ParseEvent(tep.msg)
+	return func(t *testing.T) {
```
boz, line 27: I'm guessing that you're returning a closure so that `t.Run()` executes in parallel... but `t.Parallel()` is not called and `ParseEvent()` wouldn't be run in it anyways.

I think just putting all of this in the `TestEventParsing` function instead of as a method would make this a lot easier to understand.

Or if you really want to use a method, put the closure in `TestEventParsing` instead of here.

x/deployment/types/events_test.go
Ropes: Yeah I just pulled that out of local setup.

x/deployment/types/events_test.go
Ropes: This pattern is purely due to the linter being tyranical about what state can be referenced in the main `t.Run(...)` function.  Nothing to do with t.Parallel().

Linter is harsh...

PR #622, opened 2020-05-28
perf: remove provider transaction event subscription

### LINKED ISSUE #606: events/publish: remove transaction subscription
[`events.Publish`](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/events/publish.go#L19) subscribes to the tendermint event bus and re-publishes (processed) events to the rest of the `provider/` modules.

It is currently listening for both [transaction events](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/events/publish.go#L29-L33) and [new block events](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/events/publish.go#L35-L39).

Event lifecycle/flow is currently:

1. keepers [create events](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/x/deployment/keeper/keeper.go#L110-L112)
1. handlers [return created events](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/x/provider/handler/handler.go#L37) which are added to the block.
3. `events.Publish` processes [events in a block](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/events/publish.go#L68-L69) and re-publishes them to a local event bus.

These events are present in tendermint's [new block events](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/events/query.go#L16-L19); the transaction events are unnecessarily and likely incredibly costly.

Ropes:
* Deactivated events.Publish subscribes to the tendermint event bus
and re-published events to the rest of the provider/
modules.
* Small docstring additions.

fixes #606

### DIFF
--- a/events/publish.go
+++ b/events/publish.go
@@ -15,22 +15,16 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-// Publish publishes events along tm buses to clients
+// Publish events using tm buses to clients. Waits on context
+// shutdown signals to exit.
 func Publish(ctx context.Context, tmbus tmclient.EventsClient, name string, bus pubsub.Bus) error {
 
 	const (
 		queuesz = 100
 	)
 	var (
-		txname  = name + "-tx"
 		blkname = name + "-blk"
 	)
-
-	txch, err := tmbus.Subscribe(ctx, txname, txQuery().String(), queuesz)
-	if err != nil {
-		return err
-	}
-	defer tmbus.UnsubscribeAll(ctx, txname)
 
 	blkch, err := tmbus.Subscribe(ctx, blkname, blkQuery().String(), queuesz)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


provider/cmd/run.go
```diff
@@ -47,6 +47,7 @@ func runCmd(cdc *codec.Codec) *cobra.Command {
 	return cmd
 }
 
+// doRunCmd initializes all of the Provider functionality, hangs, and awaits shutdown signals.
```
jackzampolin, line 50: comments ++

events/publish.go
```diff
@@ -15,23 +15,17 @@ import (
...
 
 	const (
 		queuesz = 100
 	)
 	var (
-		txname  = name + "-tx"
 		blkname = name + "-blk"
 	)
 
-	txch, err := tmbus.Subscribe(ctx, txname, txQuery().String(), queuesz)
```
boz, line 29: I think we can nix the `txQuery()` definition.

Ropes: `deadcode` linter caught it too!

PR #573, opened 2020-05-12
refactor: fixes gocritic complains

troian:
fixes #543

combine if statements executing same bodies

PR #625, opened 2020-05-29
feat: Makefile target matching GH linters

Ropes:
Provides the ability to run the same golangci-lint tests which are
triggered for every PR.

~~`make test-gh-lint`~~ `make test-sublinters` utilizes the same config file as GH tests. Two of
the linters are disabled same as they are for GH PRs with
`continue-on-error`.

When a linter check is broken, it fails the Make target and prints cause
of failure.


Makefile
```diff
@@ -79,10 +79,54 @@ test-coverage:
...
+	#golangci-lint run ./... --disable-all --enable gocritic
+	golangci-lint run ./... --disable-all --enable goconst
+	#golangci-lint run ./... --disable-all --enable gochecknoglobals
+	golangci-lint run ./... --disable-all --enable ineffassign --no-config
+	golangci-lint run ./... --disable-all --enable unparam
+	golangci-lint run ./... --disable-all --enable staticcheck
+	golangci-lint run ./... --disable-all --enable golint
+	golangci-lint run ./... --disable-all --enable gosec
+	golangci-lint run ./... --disable-all --enable scopelint
+	golangci-lint run ./... --disable-all --enable prealloc
```
boz, line 98: remove :point_up: and use make targets in the gh workflow?

Ropes:
Will clean up and rebase once https://github.com/ovrclk/akash/pull/573 is merged.

Ropes:
ping @boz, I think we're good to go!

boz:
@Ropes still waiting for it to be used in the GH workflow so that the two don't get out of sync.

Ropes:
@boz GH action now uses `make test-sublinters` target in tests.yml.

PR #636, opened 2020-06-03
feat: implement AnteHandler for Deployment

### LINKED ISSUE #618: feat: guard transactions with ante handlers
Right now, there is no state validation for messages/transactions.  Use cosmos-sdk ante-handlers to reject messages that would not be written to the chain otherwise.

For example:

* [x] [deployment exists](https://github.com/ovrclk/akash/blob/0782ac7f9b707e36912fe3d501569313e7bbe7ec/x/deployment/handler/handler.go#L39-L41)
* [x] [deployment not found](https://github.com/ovrclk/akash/blob/0782ac7f9b707e36912fe3d501569313e7bbe7ec/x/deployment/handler/handler.go#L39-L41)
* [x] [deployment already closed](https://github.com/ovrclk/akash/blob/0782ac7f9b707e36912fe3d501569313e7bbe7ec/x/deployment/handler/handler.go#L39-L41)
* [ ] [order not exist](https://github.com/ovrclk/akash/blob/0782ac7f9b707e36912fe3d501569313e7bbe7ec/x/market/handler/handler.go#L26-L29)
* [ ] [order invalid state](https://github.com/ovrclk/akash/blob/0782ac7f9b707e36912fe3d501569313e7bbe7ec/x/market/handler/handler.go#L31-L33) (be careful with this one - not deterministic/actual state in ante handler)
* [ ] [bid price higher than order price](https://github.com/ovrclk/akash/blob/0782ac7f9b707e36912fe3d501569313e7bbe7ec/x/market/handler/handler.go#L35-L37)
* [ ] [bid not exist](https://github.com/ovrclk/akash/blob/0782ac7f9b707e36912fe3d501569313e7bbe7ec/x/market/handler/handler.go#L58-L61)
* [ ] [lease not exist](https://github.com/ovrclk/akash/blob/0782ac7f9b707e36912fe3d501569313e7bbe7ec/x/market/handler/handler.go#L63-L66)
* [ ] [order not exist](https://github.com/ovrclk/akash/blob/0782ac7f9b707e36912fe3d501569313e7bbe7ec/x/market/handler/handler.go#L68-L71)
* [ ] [bad states](https://github.com/ovrclk/akash/blob/0782ac7f9b707e36912fe3d501569313e7bbe7ec/x/market/handler/handler.go#L73-L79) again, careful

troian:
ref #618

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -25,6 +25,8 @@ import (
 	stakingTypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 
 	"github.com/ovrclk/akash/x/deployment"
+	dante "github.com/ovrclk/akash/x/deployment/ante"
+
 	"github.com/ovrclk/akash/x/market"
 	"github.com/ovrclk/akash/x/provider"
 
@@ -136,7 +138,6 @@ func MakeCodec() *codec.Codec {
 }
 
 // https://github.com/cosmos/sdk-tutorials/blob/c6754a1e313eb1ed973c5c91dcc606f2fd288811/app.go#L73
-
 // NewApp creates and returns a new Akash App.
 func NewApp(
 	logger log.Logger, db dbm.DB, tio io.Writer, invCheckPeriod uint, skipUpgradeHeights map[int64]bool, options ...func(*bam.BaseApp),
@@ -410,10 +411,15 @@ func NewApp(
 	app.SetBeginBlocker(app.BeginBlocker)
 
 	app.SetAnteHandler(
-		auth.NewAnteHandler(
-			app.keeper.acct,
-			app.keeper.supply,
-			auth.DefaultSigVerificationGasConsumer,
+		chainAnteHandlers(
+			auth.NewAnteHandler(
+				app.keeper.acct,
+				app.keeper.supply,
+				auth.DefaultSigVerificationGasConsumer,
+			),
+			sdk.ChainAnteDecorators(
+				dante.NewDecorator(app.keeper.deployment),
+			),
 		),
 	)
 
--- a/x/deployment/ante/ante.go
+++ b/x/deployment/ante/ante.go
@@ -1,0 +1,72 @@
+package ante
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/ovrclk/akash/x/deployment/keeper"
+	"github.com/ovrclk/akash/x/deployment/types"
+)
+
+type deploymentDecorator struct {
+	keeper keeper.Keeper
+}
+
+// NewDecorator returns a decorator for "deployment" type messages
+func NewDecorator(keeper keeper.Keeper) sdk.AnteDecorator {
+	return deploymentDecorator{
+		keeper: keeper,
+	}
+}
+
+func (dd deploymentDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	for _, msg := range tx.GetMsgs() {
+		var err error
+		switch msg := msg.(type) {
+		case types.MsgCreateDeployment:
+			err = handleMsgCreate(ctx, dd.keeper, msg)
+		case types.MsgUpdateDeployment:
+			err = handleMsgUpdate(ctx, dd.keeper, msg)
+		case types.MsgCloseDeployment:
+			err = handleMsgClose(ctx, dd.keeper, msg)
+		default:
+			continue
+		}
+
+		if err != nil {
+			return ctx, err
+		}
+	}
+	return next(ctx, tx, simulate)
+}
+
+func handleMsgCreate(ctx sdk.Context, keeper keeper.Keeper, msg types.MsgCreateDeployment) error {
+	if _, found := keeper.GetDeployment(ctx, types.DeploymentID{
+		Owner: msg.Owner,
+		DSeq:  uint64(ctx.BlockHeight()),
+	}); found {
+		return types.ErrDeploymentExists
+	}
+
+	return nil
+}
+
+func handleMsgUpdate(ctx sdk.Context, keeper keeper.Keeper, msg types.MsgUpdateDeployment) error {
+	if _, found := keeper.GetDeployment(ctx, msg.ID); !found {
+		return types.ErrDeploymentNotFound
+	}
+
+	return nil
+}
+
+func handleMsgClose(ctx sdk.Context, keeper keeper.Keeper, msg types.MsgCloseDeployment) error {
+	deployment, found := keeper.GetDeployment(ctx, msg.ID)
+	if !found {
+		return types.ErrDeploymentNotFound
+	}
+
+	if deployment.State == types.DeploymentClosed {
+		return types.ErrDeploymentClosed
+	}
+
+	return nil
+}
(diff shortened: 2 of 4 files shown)


app/app.go
```diff
@@ -135,8 +137,20 @@ func MakeCodec() *codec.Codec {
 	return cdc.Seal()
 }
 
-// https://github.com/cosmos/sdk-tutorials/blob/c6754a1e313eb1ed973c5c91dcc606f2fd288811/app.go#L73
+func ChainAnteChains(chains ...sdk.AnteHandler) sdk.AnteHandler {
```
boz, line 140: this can be private.  It really joins antes, maybe:

```go
func mergeAnteHandlers(handlers ...sdk.AnteHandler) sdk.AnteHandler {
```

x/deployment/ante/ante.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	"github.com/ovrclk/akash/x/deployment/keeper"
+	"github.com/ovrclk/akash/x/deployment/types"
+)
+
+type DeploymentDecorator struct {
```
boz, line 11: this can be private.  also, it probably doesn't need to be specific to deployment-qualified:

```go
type anteDecorator struct {
  keeper keeper.Keeper
}
```

x/deployment/ante/ante.go
```diff
@@ -0,0 +1,47 @@
...
+	keeper keeper.Keeper
+}
+
+// NewHandler returns a handler for "deployment" type messages
+func NewDeploymentDecorator(keeper keeper.Keeper) DeploymentDecorator {
```
boz, line 16:
```go
func NewHandler(keeper keeper.Keeper) sdk.AnteHandler {
```

x/deployment/ante/ante.go
```diff
@@ -0,0 +1,47 @@
...
+	return DeploymentDecorator{
+		keeper: keeper,
+	}
+}
+
+func (dd DeploymentDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	for _, msg := range tx.GetMsgs() {
+		switch msg := msg.(type) {
+		case types.MsgCreateDeployment:
```
boz, line 25: break these cases out into smaller functions like is done in the handler package.

x/deployment/handler/handler.go
```diff
@@ -36,9 +36,9 @@ func handleMsgCreate(ctx sdk.Context, keeper keeper.Keeper, _ MarketKeeper, msg
 		// Version: sdk.Address.Bytes(),
 	}
 
-	if _, found := keeper.GetDeployment(ctx, deployment.ID()); found {
-		return nil, types.ErrDeploymentExists
-	}
+	// if _, found := keeper.GetDeployment(ctx, deployment.ID()); found {
+	// 	return nil, types.ErrDeploymentExists
+	// }
```
boz, line 41: I think we should keep this but maybe `panic()` here?

app/app.go
troian: Either option will go. Reason why I did it this way is compatibility with sdk API. If they allow at some point to append handlers to default chain we won't need to change fn signatures if go with 2nd option

```go
app.SetAnteHandler(
  mergeAnteHandlers(
    auth.NewAnteHandler(),
    dante.NewHandler(),
  ),
)
```

```go
app.SetAnteHandler(
  mergeAnteHandlers(
    auth.NewAnteHandler(),
    sdk.ChainAnteDecorators(
        dante.NewHandler(),
    ),
  ),
)
```

boz: right on, thanks.  your call.

app/app.go
```diff
@@ -135,8 +137,26 @@ func MakeCodec() *codec.Codec {
...
 }
 
-// https://github.com/cosmos/sdk-tutorials/blob/c6754a1e313eb1ed973c5c91dcc606f2fd288811/app.go#L73
+// chainAnteChains point of this function is to sequentially run multiple chains
+// like it would be one single chain. It becomes useful in case app implements own
+// AnteHandlers and app wants run them with all default handlers from the sdk
+// Purpose of the function came from an issue that sdk.AnteHandler does not have next parameter
+// and default chain can be accessed only from ante.NewAnteHandler which returns sdk.AnteHandler.
+// It eliminates possibility for apps to chain own Ante
+func chainAnteChains(chains ...sdk.AnteHandler) sdk.AnteHandler {
```
boz, line 146: Can you put this in a different file?

It consumes handlers and produces a handler, I still think the name is a bit off, but that's okay I guess.

app/app.go
```diff
@@ -410,10 +430,15 @@ func NewApp(
...
-			app.keeper.supply,
-			auth.DefaultSigVerificationGasConsumer,
+		chainAnteChains(
+			auth.NewAnteHandler(
+				app.keeper.acct,
+				app.keeper.supply,
+				auth.DefaultSigVerificationGasConsumer,
+			),
+			sdk.ChainAnteDecorators(
+				dante.NewDecorator(app.keeper.deployment),
```
boz, line 440: :metal: :fire: `dante.Inferno()` :fire: :metal:

x/deployment/ante/ante.go
```diff
@@ -0,0 +1,80 @@
...
+	return next(ctx, tx, simulate)
+}
+
+func handleMsgCreate(ctx sdk.Context, keeper keeper.Keeper, msg types.MsgCreateDeployment) error {
+	deployment := types.Deployment{
+		DeploymentID: types.DeploymentID{
+			Owner: msg.Owner,
+			DSeq:  uint64(ctx.BlockHeight()),
+		},
+		State: types.DeploymentActive,
```
boz, line 49: If you're only using the ID then you don't need the entire `types.Deployment`.

x/deployment/handler/handler.go
```diff
@@ -36,10 +36,6 @@ func handleMsgCreate(ctx sdk.Context, keeper keeper.Keeper, _ MarketKeeper, msg
 		// Version: sdk.Address.Bytes(),
 	}
 
-	if _, found := keeper.GetDeployment(ctx, deployment.ID()); found {
-		return nil, types.ErrDeploymentExists
-	}
```
boz, line 41: should we `panic()` here or in `keeper.Create()`?

app/app.go
troian: `tools.go` in app ?

app/app.go
troian: :boom: tbh I wanted to name it this way initially

x/deployment/handler/handler.go
troian: `keeper.Create()` looks as right place

app/app.go
boz: sure sounds good.  or `utils.go`.  thank you.

troian:
Seen them failing on other PRs too

boz:
@troian they work okay in my draft PR.  I think that since you guys are touching `x/*` it is breaking some assumptions of their code.  It makes sense especially with the added ante handler - transactions are being rejected now.

troian:
Oh, I see. Will take a look

Issue #637, opened 2020-06-03
perf: Bid querying uses prefix 


Ropes:
Improve the search performance of querying for Bids. 

Function to update: https://github.com/ovrclk/akash/blob/master/x/market/keeper/keeper.go#L343-L349

Potential implementation:
```
		store := ctx.KVStore(k.skey)
		iter := sdk.KVStorePrefixIterator(store, orderPrefix)
		for ; iter.Valid(); iter.Next() {
			var val types.Order
			k.cdc.MustUnmarshalBinaryBare(iter.Value(), &val)
			if stop := fn(val); stop {
				break
			}
		}
```

Needs additional filtering to match the Order being matched though.

akhilkumarpilli:
Sorry, I didn't get this, Currently, what we implemented is using prefix right? `k.WithBids` is a function which uses prefix search.

boz:
Indexes for bids are `(tenant-id,dseq,gseq,oseq,provider-id)` - `WithBidsForOrder()` can do a range on `(tenant-id,dseq,gseq,oseq)`.

PR #638, opened 2020-06-03
enhancement: equivalent bids deterministic winner

### LINKED ISSUE #619: order matching: handle same-price bids
orders and bids need to be matched deterministically.  currently, bids are ordered by [price](https://github.com/ovrclk/akash/blob/0782ac7f9b707e36912fe3d501569313e7bbe7ec/x/market/handler/endblock.go#L82-L85), but this is insufficient - if two bids have the same price, the same bid might not be chosen everywhere.

Add additional ordering, including:

* block with which the bid was created - earlier wins.
* when all else fails, find a deterministic way to select a winner that gives all tied bids an equal chance of winning.

Ropes:
Deterministically choose a winner from a set of bids for an order.

* Sort bids on order: collect lowest with matching bid values
* From lowest bids, aggregate all bids' strings and fnv hash
* `fnv value` %(remainder) len(matching bids) = winning bid.
* Unit tests to assert deterministic return values.

* TODO: https://github.com/ovrclk/akash/issues/637 to improve 
bid query performance.

fixes #619

### DIFF
--- a/testutil/ids.go
+++ b/testutil/ids.go
@@ -4,12 +4,12 @@ import (
 	"math/rand"
 	"testing"
 
-	ckeys "github.com/cosmos/cosmos-sdk/client/keys"
 	"github.com/cosmos/cosmos-sdk/crypto/keys"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/tendermint/tendermint/crypto/ed25519"
+
 	dtypes "github.com/ovrclk/akash/x/deployment/types"
 	mtypes "github.com/ovrclk/akash/x/market/types"
-	"github.com/stretchr/testify/require"
 )
 
 func Keyring(t testing.TB) keys.Keybase {
@@ -17,11 +17,12 @@ func Keyring(t testing.TB) keys.Keybase {
 	return obj
 }
 
+// AccAddress provides an Account's Address bytes from a ed25519 generated
+// private key.
 func AccAddress(t testing.TB) sdk.AccAddress {
 	t.Helper()
-	info, _, err := Keyring(t).CreateMnemonic("test", keys.English, ckeys.DefaultKeyPass, keys.Secp256k1)
-	require.NoError(t, err)
-	return info.GetAddress()
+	privKey := ed25519.GenPrivKey()
+	return sdk.AccAddress(privKey.PubKey().Address())
 }
 
 func DeploymentID(t testing.TB) dtypes.DeploymentID {
--- a/x/market/handler/endblock.go
+++ b/x/market/handler/endblock.go
@@ -1,10 +1,12 @@
 package handler
 
 import (
+	"hash/fnv"
 	"sort"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/ovrclk/akash/x/market/types"
+	"github.com/pkg/errors"
 )
 
 // OnEndBlock transfer funds for active leases and update order states
@@ -55,6 +57,52 @@ func transferFundsForActiveLeases(ctx sdk.Context, keepers Keepers) error {
 	return nil
 }
 
+var errNoBids error = errors.New("no bids to pick winner from")
+
+func pickBidWinner(bids []types.Bid) (winner *types.Bid, err error) {
+	// open bids; match by lowest price; sort bids by price
+	sort.Slice(bids, func(i, j int) bool {
+		// The BidID DSeq is pulled from the original OrderID.
+		// So it can't be used to determine who bid first like a timestamp.
+		return bids[i].Price.IsLT(bids[j].Price)
+	})
+	switch len(bids) {
+	case 0:
+		// This is a fatal case
+		return nil, errNoBids
+	case 1:
+		return &bids[0], nil
+	}
+
+	if !bids[0].Price.IsEqual(bids[1].Price) {
+		// Lowest bid(0) is unique, return the winner
+		return &bids[0], nil
+	}
+
+	// There are equivalent bid prices; select winner with deterministic
+	// random ordering based on given bids.
+	// FNV hash provider addresses all of the bids
+	h := fnv.New32a()
+	bidIndex := 0
+	_, err = h.Write(bids[bidIndex+1].Provider.Bytes())
+	if err != nil {
+		return nil, err
+	}
+	for ; bidIndex+1 < len(bids); bidIndex++ {
+		if !bids[bidIndex].Price.IsEqual(bids[bidIndex+1].Price) {
+			break
+		}
+		_, err := h.Write(bids[bidIndex+1].Provider.Bytes())
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Create a numeric hash from the Stringified Bid values
+	n := int(h.Sum32()) % (bidIndex + 1) // Calculate the remainder to select index of equal bids
+	return &bids[n], nil
+}
+
 func matchOrders(ctx sdk.Context, keepers Keepers) error {
 
 	// match unmatched orders.
@@ -78,22 +126,24 @@ func matchOrders(ctx sdk.Context, keepers Keepers) error {
 			return false
 		}
 
-		// open bids; match by lowest price
-		sort.Slice(bids, func(i, j int) bool {
-			// TODO handle same price
-			return bids[i].Price.IsLT(bids[j].Price)
-		})
-
-		winner := bids[0]
+		winner, err := pickBidWinner(bids)
+		if err != nil {
+			pErr := errors.Wrap(err, "picking bid winner returned unrecoverable error")
+			panic(pErr.Error())
+		}
 
 		// create lease
-		keepers.Market.CreateLease(ctx, winner)
+		keepers.Market.CreateLease(ctx, *winner)
 
 		// set winning bid state to matched
-		keepers.Market.OnBidMatched(ctx, winner)
+		keepers.Market.OnBidMatched(ctx, *winner)
 
 		// set losing bids to state lost
-		for _, bid := range bids[1:] {
+		// Set all but winning bid to State: Lost
+		for _, bid := range bids {
+			if winner.Equals(bid.BidID) {
+				continue // skip setting state to lost
+			}
 			keepers.Market.OnBidLost(ctx, bid)
 		}
 
--- a/x/market/handler/endblock_test.go
+++ b/x/market/handler/endblock_test.go
@@ -1,0 +1,190 @@
+package handler
+
+import (
+	"errors"
+	"strconv"
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/ovrclk/akash/testutil"
+	"github.com/ovrclk/akash/x/market/types"
+)
+
+type winnerTest struct {
+	desc      string
+	bids      []types.Bid
+	expWinner *types.Bid
+	expErr    error
+}
+
+func (w *winnerTest) testFunc(t *testing.T) {
+	winner, err := pickBidWinner(w.bids)
+	if !errors.Is(err, w.expErr) {
+		t.Errorf("returned err: %v does not match %v", err, w.expErr)
+	}
+	if w.expWinner != nil && !winner.Equals(w.expWinner.ID()) {
+		t.Errorf("unexpected winner: %#v\n%q : %v", winner, types.BidIDString(winner.BidID), winner.Price)
+		t.Logf("winner: %+v coin: %s", winner.ID(), winner.Price.String())
+	}
+}
+
+func TestBidWinner(t *testing.T) {
+	originOID := testutil.OrderID(t)
+
+	bid0 := createBid(t, originOID, 3)
+	bid1 := createBid(t, originOID, 5)
+	bid2 := createBid(t, originOID, 5)
+	bid3 := createBid(t, originOID, 4)
+	bid4 := createBid(t, originOID, 5)
+	bid5 := createBid(t, originOID, 5)
+	bid6 := createBid(t, originOID, 5)
+	bid7 := createBid(t, originOID, 8)
+	bid8 := createBid(t, originOID, 6)
+	bid9 := createBid(t, originOID, 5)
+
+	var winnerTests = []winnerTest{
+		{
+			desc:      "no bids",
+			bids:      []types.Bid{},
+			expWinner: nil,
+			expErr:    errNoBids,
+		},
+		{
+			desc:      "single bid",
+			bids:      []types.Bid{bid9},
+			expWinner: &bid9,
+		},
+		{
+			desc:      "two bids",
+			bids:      []types.Bid{bid7, bid4},
+			expWinner: &bid4,
+		},
+		{
+			desc:      "two matching bids",
+			bids:      []types.Bid{bid9, bid7, bid4},
+			expWinner: &bid4,
+		},
+		{
+			desc:      "all the same bid values",
+			bids:      []types.Bid{bid9, bid9, bid9, bid9, bid9},
+			expWinner: &bid9,
+		},
+		{
+			desc:      "two of the same",
+			bids:      []types.Bid{bid9, bid3},
+			expWinner: &bid3,
+		},
+		{
+			desc:      "multiple of the same, but one lowest value",
+			bids:      []types.Bid{bid1, bid0, bid9, bid2, bid3, bid4},
+			expWinner: &bid0,
+		},
+		{
+			desc:      "confirm last bid in list is picked",
+			bids:      []types.Bid{bid1, bid9, bid2, bid3, bid4, bid0},
+			expWinner: &bid0,
+		},
+		{
+			desc:      "two matching low bids",
+			bids:      []types.Bid{bid5, bid6, bid7, bid8},
+			expWinner: &bid6,
+		},
+	}
+
+	for _, test := range winnerTests {
+		t.Run(test.desc, test.testFunc)
+	}
+}
+
+type testDist struct {
+	desc      string
+	bidNum    int
+	rounds    int
+	expErr    error
+	expUneven bool
+}
+
+func (td *testDist) testFunc(t *testing.T) {
+	originOID := testutil.OrderID(t)
+
+	distributionSpread := make(map[int]int, td.bidNum)
+	for i := 0; i < td.rounds; i++ {
+		bIndex := make(map[string]int, td.bidNum)
+		bids := make([]types.Bid, 0, td.bidNum)
+		// generate N bids all with the same bidding amount
+		for j := 0; j < td.bidNum; j++ {
+			b := createBid(t, originOID, 5)
+			bIndex[b.Provider.String()] = j
+			bids = append(bids, b)
+		}
+
+		winner, err := pickBidWinner(bids)
+		if !errors.Is(err, td.expErr) {
+			t.Errorf("returned err: %v does not match %v", err, td.expErr)
+		}
+		// Check provider
+		slot := bIndex[winner.Provider.String()]
+		distributionSpread[slot]++
+	}
+
+	for i := 0; i < td.bidNum; i++ {
+		v := distributionSpread[i]
+		if v == 0 && !td.expUneven {
+			t.Errorf("index[%d] had no winners when even distribution was expected", i)
+			continue
+		}
+		// calculate a reasonable low expectation of wins given the number of test rounds and number of bidders
+		r := float64(td.rounds/td.bidNum) * 0.5
+		if float64(v) < r {
+			t.Errorf("less than %d winnings of [%d]", td.bidNum, i)
+		}
+		t.Logf("winner distribution: %d: %d", i, v)
+	}
+
+}
+
+func TestWinningDistribution(t *testing.T) {
+	tests := []testDist{
+		{
+			desc:   "one bidder",
+			bidNum: 1,
+			rounds: 50,
+			expErr: nil,
+		},
+		{
+			desc:   "five winners",
+			bidNum: 5,
+			rounds: 50,
+			expErr: nil,
+		},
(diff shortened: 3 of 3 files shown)


x/market/handler/endblock.go
```diff
@@ -55,6 +56,49 @@ func transferFundsForActiveLeases(ctx sdk.Context, keepers Keepers) error {
 	return nil
 }
 
+func pickBidWinner(bids []types.Bid) (winner types.Bid) {
+	// open bids; match by lowest price; sort bids by price
+	sort.Slice(bids, func(i, j int) bool {
+		// AFAICT: the BidID DSeq is pulled from the original OrderID.
+		// So it can't be used to determine who bid first.
```
boz, line 63: yeah, all of these bids should be for the same order (same order id -> same group id -> same deployment id).

x/market/handler/endblock.go
```diff
@@ -55,6 +56,49 @@ func transferFundsForActiveLeases(ctx sdk.Context, keepers Keepers) error {
...
+	if len(bids) == 0 {
+		// This is a fatal case
+		winner = types.Bid{}
+		return
+	} else if len(bids) == 1 {
+		winner = bids[0]
+		return
+	} // else > 2 bids
+
+	if bids[0].Price.IsEqual(bids[1].Price) {
```
boz, line 75: invert this - do a `!` here and return early.

x/market/handler/endblock.go
```diff
@@ -55,6 +56,49 @@ func transferFundsForActiveLeases(ctx sdk.Context, keepers Keepers) error {
...
+		// There are equivalent bid prices; select winner with deterministic
+		// random ordering based on given bids.
+		equalBids := make([]types.Bid, 0)
+		for i := 0; bids[0].Price.IsEqual(bids[i].Price); i++ {
+			equalBids = append(equalBids, bids[i])
```
boz, line 80: might be small performance gain by just finding the index and using `bids[:idx]` or even just `idx` later.

x/market/handler/endblock.go
```diff
@@ -55,6 +56,49 @@ func transferFundsForActiveLeases(ctx sdk.Context, keepers Keepers) error {
...
+		}
+
+		// Stringify all of the bids
+		// Strings include Provider+Owner Addresses plus Sequence numbers
+		addrs := make([]byte, 0)
+		for _, bid := range equalBids {
+			addrs = append(addrs, []byte(types.BidIDString(bid.ID()))...)
```
boz, line 87: fwiw I've started to just use `bytes.Buffer` to keep these things easier on the eyes, but...

x/market/handler/endblock.go
```diff
@@ -55,6 +56,49 @@ func transferFundsForActiveLeases(ctx sdk.Context, keepers Keepers) error {
...
+		}
+
+		// Create a numeric hash from the Stringified Bid values
+		h := fnv.New32a()
+		h.Write(addrs)
```
boz, line 92: you can accumulate the hash inside the loop of equal bids instead of writing to a buffer first.

x/market/handler/endblock.go
```diff
@@ -55,6 +56,49 @@ func transferFundsForActiveLeases(ctx sdk.Context, keepers Keepers) error {
...
+		n := int(h.Sum32()) % len(equalBids) // Calculate the remainder to select index of equal bids
```
boz, line 93: :100: great idea.

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,84 @@
...
+	bid1 := createBid(t, uint64(20), int64(5))
+	bid2 := createBid(t, uint64(19), int64(5))
+	bid3 := createBid(t, uint64(20), int64(4))
+	bid4 := createBid(t, uint64(20), int64(5))
+	bid5 := createBid(t, uint64(19), int64(5))
+	bid6 := createBid(t, uint64(18), int64(5))
+	bid7 := createBid(t, uint64(20), int64(8))
+	bid8 := createBid(t, uint64(20), int64(6))
+
+	var winnerTests []winnerTest = []winnerTest{
```
boz, line 31: nit: should just be able to do

```go
tests := []winnerTest{
// ...
}
```

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	for i, test := range winnerTests {
+		f := func(_ *testing.T, test winnerTest) func(t *testing.T) {
+			return func(t *testing.T) {
+				winner := pickBidWinner(test.bids)
+				if !winner.Equals(test.expWinner.ID()) {
+					t.Errorf("unexpected winner: %#v\n%q : %v", winner, types.BidIDString(winner.BidID), winner.Price)
+				}
+			}
+		}
```
boz, line 66: This is making tests too complicated and difficult to read.  I really think it's over-doing it.

I generally add a message to table-based test elements to use in the assertions.  Something to consider:

```go
tests := []struct{
  name string
  val    int
  expect int
}{
  {
     name: "zero-property",
     val: 0,
     expect: 0,
  },
}
for _, test := tests {
  assert.Equal(t,test.expect,foo(test.val),test.name)
}
```

That was before `t.Run()`, and I get that you're hitting lint and semantic issues with it, but there has to be a better way around it.

```go
for i := range winnerTests {
  t.Run(strconv.Itoa(i),func(t *testing.T) {
    assert.Equals(t,winnerTests[i].expWinner.ID(),pickBidWinner(winnerTests[i].bids))
  })

File: x/market/handler/endblock.go (hunk 5 above)
Comment (boz) [reply x/market/handler/endblock.go:92]: in fact you can do it all in the loop that is looking for equal bids.

File: x/market/handler/endblock.go (hunk 3 above)
Comment (Ropes) [reply x/market/handler/endblock.go:80]: That might help solve the simulations index error, good idea.

File: x/market/handler/endblock_test.go (hunk 8 above)
Comment (boz) [reply x/market/handler/endblock_test.go:66]: I mean, is it right?

File: x/market/handler/endblock.go (hunk 2 above)
Comment (Ropes) [reply x/market/handler/endblock.go]: done

**Comment by Ropes** [comment]
Current simulation test failure:
```
Starting the simulation from time Fri Nov 19 14:27:08 UTC 7351 (unixtime 169835840828)
Simulating... block 34/50, operation 0/154. Logs to writing to /home/ropes/.simapp/simulations/2020-06-04_13:46:44.log
    TestAppStateDeterminism: simulate.go:289: error on block  34/50, operation (11/154) from x/market:
        no lease for order: failed to execute message; message index: 0 [cosmos/cosmos-sdk@v0.38.3/baseapp/baseapp.go:663]
        Comment: 
--- FAIL: TestAppStateDeterminism (53.70s)
FAIL
FAIL    github.com/ovrclk/akash/app     53.754s
```

Working on untangling the simulation configuration so I can understand the steps.

File: x/market/handler/endblock.go [hunk 9]
```diff
@@ -78,19 +125,16 @@ func matchOrders(ctx sdk.Context, keepers Keepers) error {
...
-		// open bids; match by lowest price
-		sort.Slice(bids, func(i, j int) bool {
-			// TODO handle same price
-			return bids[i].Price.IsLT(bids[j].Price)
-		})
-
-		winner := bids[0]
+		winner, err := pickBidWinner(bids)
+		if err != nil {
+			return false
```
boz, line 130: should probably panic here :cry:

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,125 @@
...
+	bid4 := createBid(t, uint64(20), int64(5))
+	bid5 := createBid(t, uint64(19), int64(5))
+	bid6 := createBid(t, uint64(18), int64(5))
+	bid7 := createBid(t, uint64(20), int64(8))
+	bid8 := createBid(t, uint64(20), int64(6))
+	bid9 := createBid(t, uint64(20), int64(5))
+
+	var winnerTests = []winnerTest{
+		{
+			desc:      "no bids",
```
boz, line 54: :+1:

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if bytes.Equal(seed0, seed1) {
+		t.Fatalf("random seed bytes the same")
+	}
+	t.Logf("seed data: %q %q", seed0, seed1)
+	return types.Bid{
+		BidID: types.BidID{
+			Owner:    testutil.DeterministicAccAddress(t, seed0),
```
boz, line 116: You need the `OrderID` to be the same for all bids - `Owner`, `DSeq`, `GSeq`, `OSeq`.  You can use `testutil.AccAddress(t)` for that and pass it into this function along with `block`.

This constant seed thing is really unnecessary and bloats the codebase.

x/market/handler/endblock.go
Ropes: So in this case, returning false would block iterating over all the orders if I understand correctly?
https://github.com/ovrclk/akash/blob/2b84c28a07d70653bc35543cf27d90eae0378597/x/market/keeper/keeper.go#L295-L301

Normally I always balk at using panics, but I'm not sure the SDK gives us a better option sadly... :disappointed:

boz: If there's an error when trying to match orders then the state is broken - in cosmos you deal with that by blowing up.

x/market/handler/endblock.go
```diff
@@ -55,6 +57,53 @@ func transferFundsForActiveLeases(ctx sdk.Context, keepers Keepers) error {
...
+	// There are equivalent bid prices; select winner with deterministic
+	// random ordering based on given bids.
+	bidIndex := 0
+	for i := 0; i < len(bids); i++ {
+		if bids[0].Price.IsEqual(bids[i].Price) {
+			bidIndex++
+		} else {
+			break
+		}
+	}
```
boz, line 90:
```go
h := fnv.New32a()
h.Write(bids[bidIndex].Provider)
for ; bidIndex + 1 < len(bids); bidIndex++ {
  if !bids[bidIndex].Price.IsEqual(bids[bidIndex+1].Price) {
    break
  }
  h.Write(bids[bidIndex+1].Provider)
}
n := int(h.Sum32()) % (bidIndex + 1)
return &bids[n], nil
```

x/market/handler/endblock.go
```diff
@@ -55,6 +57,53 @@ func transferFundsForActiveLeases(ctx sdk.Context, keepers Keepers) error {
...
+	// open bids; match by lowest price; sort bids by price
+	sort.Slice(bids, func(i, j int) bool {
+		// AFAICT: the BidID DSeq is pulled from the original OrderID.
+		// So it can't be used to determine who bid first.
+		return bids[i].Price.IsLT(bids[j].Price)
+	})
+	if len(bids) == 0 {
+		// This is a fatal case
+		return nil, errNoBids
+	} else if len(bids) == 1 {
```
boz, line 72: don't really need the `else`.  another way to do it is:

```go
switch len(bids) {
  case 0:
    return nil, errNoBids
  case 1:
    return &bids[0], nil
}
```

x/market/handler/endblock.go
```diff
@@ -78,22 +127,23 @@ func matchOrders(ctx sdk.Context, keepers Keepers) error {
...
 
 		// set winning bid state to matched
-		keepers.Market.OnBidMatched(ctx, winner)
+		keepers.Market.OnBidMatched(ctx, *winner)
 
 		// set losing bids to state lost
-		for _, bid := range bids[1:] {
+		// Set all but winning bid to State: Lost
+		for _, bid := range bids {
+			if winner.Equals(bid.BidID) {
```
boz, line 144: :+1: good catch.

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,133 @@
...
+func TestBidWinner(t *testing.T) {
+	seed := make([]byte, 32)
+	_, err := r.Read(seed)
+	if err != nil {
+		t.Fatal(err)
+	}
+	ownerAddr := testutil.DeterministicAccAddress(t, seed)
+
+	bid0 := createBid(t, int64(3))
+	bid0.Owner = ownerAddr
```
boz, line 48:
```go
  oid := testutil.OrderID(t)
  bid0 := createBid(t,oid,3)
//....

func createBid(t testing.TB, oid types.OrderID, amount int) types.Bid {
  return types.Bid{
    ID:    types.MakeBidID(oid,testutil.AccAddress(t)),
    Price: sdk.NewCoin("akash",int64(amount)),
  }  
}
```

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,133 @@
...
+			desc:      "all the same bid values",
+			bids:      []types.Bid{bid9, bid9, bid9, bid9, bid9},
+			expWinner: &bid9,
+		},
+		{
+			bids:      []types.Bid{bid9, bid3},
+			expWinner: &bid3,
+		},
+		{
+			bids:      []types.Bid{bid1, bid0, bid9, bid2, bid3, bid4},
```
boz, line 100: nit: I think it's clearer to have `makeBid()` here in the table instead having `bid0...bidN`

testutil/ids.go
```diff
@@ -24,6 +27,14 @@ func AccAddress(t testing.TB) sdk.AccAddress {
 	return info.GetAddress()
 }
 
+// DeterministicAccAddress takes a seed []byte paramater for deterministic
+// private key generation.
+func DeterministicAccAddress(t testing.TB, seed []byte) sdk.AccAddress {
+	t.Helper()
+	privKey := ed25519.GenPrivKeyFromSecret(seed)
+	return sdk.AccAddress(privKey.PubKey().Address())
```
boz, line 35: I very much prefer this method to creating a keyring - would be good to re-implement `AccAddress` like this (sans asking the caller to pass a seed in).

x/market/handler/endblock_test.go
Ropes: That's much better, I should have noticed passing the OrderID was an option.

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func TestBidWinner(t *testing.T) {
+	ownerAddr := testutil.AccAddress(t)
+	originOID := types.OrderID{
+		Owner: ownerAddr,
+		DSeq:  uint64(20),
+		GSeq:  uint32(1),
+		OSeq:  uint32(1),
+	}
```
boz, line 46: Please use the generator - it makes modifying things easier in the future and removes the burden of making sure that all fields are populated and populated correctly: https://github.com/ovrclk/akash/blob/b2752f0d280bd5f91ba6f27b9b263588d5c8c29e/testutil/ids.go#L40-L43

x/market/handler/endblock.go
Ropes: Done

x/market/handler/endblock_test.go
Ropes: good call!

testutil/ids.go
Ropes: I added a dedicated method to generate it without the seed. I think it'll be useful in other places.

testutil/ids.go
```diff
@@ -24,6 +27,18 @@ func AccAddress(t testing.TB) sdk.AccAddress {
 	return info.GetAddress()
 }
 
+func AccAddressNonKeyring(t testing.TB) sdk.AccAddress {
```
boz, line 30: just replace `AccAddress` with this.

x/market/handler/endblock.go
boz: update comments

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+	"github.com/ovrclk/akash/testutil"
+	"github.com/ovrclk/akash/x/market/types"
+)
+
+var (
+	// Create and seed the generator.
+	// Typically a non-fixed seed should be used, such as time.Now().UnixNano().
+	// Using a fixed seed will produce the same output on every run.
+	r = rand.New(rand.NewSource(550000))
```
boz, line 19: unused

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,213 @@
...
+	bid0 := createBid(t, originOID, int64(3))
+	bid1 := createBid(t, originOID, int64(5))
+	bid2 := createBid(t, originOID, int64(5))
+	bid3 := createBid(t, originOID, int64(4))
+	bid4 := createBid(t, originOID, int64(5))
+	bid5 := createBid(t, originOID, int64(5))
+	bid6 := createBid(t, originOID, int64(5))
+	bid7 := createBid(t, originOID, int64(8))
+	bid8 := createBid(t, originOID, int64(6))
+	bid9 := createBid(t, originOID, int64(5))
```
boz, line 52: again, would have been more convenient if `createBid()` took an `int` as its third argument.

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,213 @@
...
+	rounds int
+	expErr error
+}
+
+func (td *testDist) testFunc(t *testing.T) {
+	originOID := testutil.OrderID(t)
+
+	distributionSpread := make(map[int]int)
+	for i := 0; i < td.bidNum; i++ {
+		distributionSpread[i] = 0
```
boz, line 122: I don't think that this is necessary - `distributionSpread[hithertoUnusedSlot]` will be `int`'s "zero-value" which is `0`.

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,213 @@
...
+	desc   string
+	bidNum int
+	rounds int
+	expErr error
+}
+
+func (td *testDist) testFunc(t *testing.T) {
+	originOID := testutil.OrderID(t)
+
+	distributionSpread := make(map[int]int)
```
boz, line 120: you know the size of this - `make(map[int]int,t.bidNum)` would work.

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,213 @@
...
+	for i := 0; i < td.bidNum; i++ {
+		distributionSpread[i] = 0
+	}
+
+	for i := 0; i < td.rounds; i++ {
+		winCount := make(map[int]int)
+		bIndex := make(map[string]int)
+		bids := make([]types.Bid, 0)
```
boz, line 128: `make([]types.Bid,0,t.bidNum)`

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,213 @@
...
+		// generate N bids all with the same bidding amount
+		for j := 0; j < td.bidNum; j++ {
+			b := createBid(t, originOID, int64(5))
+			bIndex[b.Provider.String()] = j
+			winCount[j] = 0
```
boz, line 133: not needed.

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,213 @@
...
+			bids = append(bids, b)
+		}
+
+		winner, err := pickBidWinner(bids)
+		if !errors.Is(err, td.expErr) {
+			if !errors.Is(td.expErr, errExpUnevenDistribution) {
```
boz, line 139: this is the only type of error that is expected.

x/market/handler/endblock_test.go
```diff
@@ -0,0 +1,213 @@
...
+		slot := bIndex[winner.Provider.String()]
+		distributionSpread[slot] = distributionSpread[slot] + 1
+	}
+
+	for i, v := range distributionSpread {
+		if v == 0 {
+			if td.expErr != nil && !errors.Is(td.expErr, errExpUnevenDistribution) {
+				t.Errorf("index[%d] had no winners", i)
+			}
+		}
```
boz, line 153:
```go
if v == 0 {
  testutil.Assert(t,td.inconclusive,"no winners",i)
  continue
}

File: x/market/handler/endblock_test.go [hunk 28]
```diff
@@ -0,0 +1,213 @@
...
+		// calculate a reasonable low expectation of wins given the number of test rounds and number of bidders
+		r := float64(td.rounds/td.bidNum) * 0.5
+		if float64(v) < r {
+			if td.expErr != nil && !errors.Is(td.expErr, errExpUnevenDistribution) {
+				t.Errorf("less than %d winnings of [%d]", td.bidNum, i)
+			}
```
boz, line 159:
```go
if float64(v) < r {
  testutil.Assert(t,td.inconclusive,"underperform",td.bidNum,i)
  continue
}
```

x/market/handler/endblock_test.go
Ropes: That's true. I'll change it.

x/market/handler/endblock_test.go
Ropes: Simplified to simply assert that the error matches the testDist type's expErr.

x/market/handler/endblock_test.go
Ropes: logic cleaned up.

x/market/handler/endblock_test.go
Ropes: Removed the unnecessary nesting by utilizing `continue` above.

x/market/handler/endblock_test.go
Ropes: If `createBid()` is called `[]types.Bid{createBid(),...}` style, then it's not possible to assert that a Bid in the set of Bidders was an actual winner. Granted the assertion is just checking the `Provider` address.

x/market/handler/endblock_test.go
boz: @Ropes thanks!  sorry I meant `assert.True` not `testuitl.Assert`.

Try to use `assert`, `require` when you can.  It really makes tests a lot cleaner.

PR #645, opened 2020-06-08
fix: implement antehandlers for market and provider

troian:
ref #618


troian:
provider check logic picked from #646

x/market/ante/ante.go
```diff
@@ -0,0 +1,109 @@
...
+		return types.ErrUnknownOrderForBid
+	}
+
+	if lease.State != types.LeaseActive {
+		return types.ErrLeaseNotActive
+	}
+
+	if bid.State != types.BidMatched {
+		return types.ErrBidNotMatched
+	}
```
boz, line 93: I know that this is the previous logic, but I think that we should allow bids to be closed when they're in state `BidOpen` so that someone can cancel their bid.  They won't be able to create another bid for the same order, tho.

Can discuss this in a separate ticket.

PR #656, opened 2020-06-11
bugfix: fix events

boz:
* `./events`: re-add listening to tx events
* `./x/deployment/types`: fix close deployment tx
* `./provider/cmd/run`: non-colorized logging
* `./testutil`: coin generator.

### DIFF
--- a/x/deployment/types/event.go
+++ b/x/deployment/types/event.go
@@ -134,6 +134,12 @@ func ParseEvent(ev sdkutil.Event) (interface{}, error) {
 			return nil, err
 		}
 		return EventDeploymentUpdate{ID: did}, nil
+	case evActionDeploymentClose:
+		did, err := ParseEVDeploymentID(ev.Attributes)
+		if err != nil {
+			return nil, err
+		}
+		return EventDeploymentClose{ID: did}, nil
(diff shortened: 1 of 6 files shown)


x/deployment/types/event.go
```diff
@@ -134,6 +134,12 @@ func ParseEvent(ev sdkutil.Event) (interface{}, error) {
 			return nil, err
 		}
 		return EventDeploymentUpdate{ID: did}, nil
+	case evActionDeploymentClose:
```
Ropes, line 137: :clap

provider/cluster/service.go
```diff
@@ -172,6 +172,7 @@ loop:
 			break loop
 
 		case ev := <-s.sub.Events():
+			s.log.Debug("event", "ev", ev)
```
Ropes, line 175: Double check: do you want to keep this in there? Might make `debug` very noisy, but it might be helpful.

boz: :shrug: it's not *too* bad when just running the provider.

I think we should add one at the source - in `./event/publish.go` tho so we don't have to do it all over the place.  I'll remove it from here.

Issue #642, opened 2020-06-05
deployments: allow closing groups


boz:
Leases can currently come to an end by the following events:

1. tenant closes the deployment
1. tenant runs out of money
1. tenant cancels order
1. provider cancels bid

For cases 3 and 4 above, a new order is created for the group - there's another round of bids and lease generation.

For case 1, no more orders, bids, leases are created, but it's for all groups in the deployment.

We need a way for the owner to terminate a lease for a single group in a deployment:

## Add `MsgCloseGroup` transaction in `x/deployment`

* [x] propagate changes to related objects - `MsgCloseDeployment` [essentially does this](https://github.com/ovrclk/akash/blob/b2752f0d280bd5f91ba6f27b9b263588d5c8c29e/x/deployment/handler/handler.go#L98-L99) already, but for all groups instead of just one.
* [x] add cli command

PR #657, opened 2020-06-12
feat: Close Group Message functionality

Ropes:
Enable closing a specific group within an existing Akash Deployment.
This will signal the provider to set the Group.State to Closed.

`akashctl tx deployment group close --owner=? --dseq=? --gseq=?` command
added to send the `MsgCloseGroup`.

Simulation tests passing: With a prequisite in SimulateMsgCloseGroup() to skip
sending the MsgGroupClose if the Group.State is already Closed. This is
not the same behavior as the Deployment's Simulation function, but
Simulate...Group throws the handleMsgCloseGroup() ErrGroupClosed when
the simulation runs.

fixes: #642

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,3 +243,5 @@ test-sim-after-import:
 	@echo "Running application simulation-after-import..."
 	go test -mod=readonly -tags=$(BUILD_MAINNET) $(APP_DIR) -run=TestAppSimulationAfterImport -Enabled=true \
 		-NumBlocks=50 -BlockSize=100 -Commit=true -Seed=99 -Period=5 -v -timeout 10m
+
+test-sims: test-sim-fullapp test-sim-nondeterminism test-sim-import-export test-sim-after-import
--- a/x/deployment/client/cli/tx.go
+++ b/x/deployment/client/cli/tx.go
@@ -31,6 +31,7 @@ func GetTxCmd(key string, cdc *codec.Codec) *cobra.Command {
 		cmdCreate(key, cdc),
 		cmdUpdate(key, cdc),
 		cmdClose(key, cdc),
+		cmdGroupClose(key, cdc),
 	)...)
 	return cmd
 }
@@ -133,5 +134,36 @@ func cmdUpdate(key string, cdc *codec.Codec) *cobra.Command {
 		},
 	}
 	AddDeploymentIDFlags(cmd.Flags())
+	return cmd
+}
+
+func cmdGroupClose(_ string, cdc *codec.Codec) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "group-close",
+		Short:   "close a Deployment's specific Group",
+		Example: "akashctl tx deployment group-close --owner=[Account Address] --dseq=[uint64] --gseq=[uint32]",
+		Args:    cobra.ExactArgs(0),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.NewCLIContext().WithCodec(cdc)
+			bldr := auth.NewTxBuilderFromCLI(os.Stdin).WithTxEncoder(utils.GetTxEncoder(cdc))
+			id, err := GroupIDFromFlags(cmd.Flags())
+			if err != nil {
+				return err
+			}
+
+			msg := types.MsgCloseGroup{
+				ID: id,
+			}
+			err = msg.ValidateBasic()
+			if err != nil {
+				return err
+			}
+
+			return utils.GenerateOrBroadcastMsgs(ctx, bldr, []sdk.Msg{msg})
+		},
+	}
+	AddGroupIDFlags(cmd.Flags())
+	MarkReqGroupIDFlags(cmd)
+
 	return cmd
 }
--- a/x/deployment/handler/handler.go
+++ b/x/deployment/handler/handler.go
@@ -19,7 +19,9 @@ func NewHandler(keeper keeper.Keeper, mkeeper MarketKeeper) sdk.Handler {
 		case types.MsgUpdateDeployment:
 			return handleMsgUpdate(ctx, keeper, mkeeper, msg)
 		case types.MsgCloseDeployment:
-			return handleMsgClose(ctx, keeper, mkeeper, msg)
+			return handleMsgCloseDeployment(ctx, keeper, mkeeper, msg)
+		case types.MsgCloseGroup:
+			return handleMsgCloseGroup(ctx, keeper, mkeeper, msg)
 		default:
 			return nil, sdkerrors.ErrUnknownRequest
 		}
@@ -79,7 +81,8 @@ func handleMsgUpdate(ctx sdk.Context, keeper keeper.Keeper, _ MarketKeeper, msg 
 	}, nil
 }
 
-func handleMsgClose(ctx sdk.Context, keeper keeper.Keeper, mkeeper MarketKeeper, msg types.MsgCloseDeployment) (*sdk.Result, error) {
+func handleMsgCloseDeployment(ctx sdk.Context, keeper keeper.Keeper, mkeeper MarketKeeper, msg types.MsgCloseDeployment) (*sdk.Result, error) {
+
 	deployment, found := keeper.GetDeployment(ctx, msg.ID)
 	if !found {
 		return nil, types.ErrDeploymentNotFound
@@ -99,6 +102,30 @@ func handleMsgClose(ctx sdk.Context, keeper keeper.Keeper, mkeeper MarketKeeper,
 		keeper.OnDeploymentClosed(ctx, group)
 		mkeeper.OnGroupClosed(ctx, group.ID())
 	}
+
+	return &sdk.Result{
+		Events: ctx.EventManager().Events(),
+	}, nil
+}
+
+func handleMsgCloseGroup(ctx sdk.Context, keeper keeper.Keeper, mkeeper MarketKeeper, msg types.MsgCloseGroup) (*sdk.Result, error) {
+	group, found := keeper.GetGroup(ctx, msg.ID)
+	if !found {
+		return nil, types.ErrGroupNotFound
+	}
+
+	// if Group already closed; return the validation error
+	err := group.ValidateClosable()
+	if err != nil {
+		return nil, err
+	}
+
+	// Update the Group's state
+	err = keeper.OnCloseGroup(ctx, group)
+	if err != nil {
+		return nil, err
+	}
+	mkeeper.OnGroupClosed(ctx, group.ID())
 
 	return &sdk.Result{
 		Events: ctx.EventManager().Events(),
--- a/x/deployment/keeper/keeper.go
+++ b/x/deployment/keeper/keeper.go
@@ -2,16 +2,10 @@ package keeper
 
 import (
 	"github.com/cosmos/cosmos-sdk/codec"
-	"github.com/pkg/errors"
 
 	"github.com/ovrclk/akash/x/deployment/types"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
-)
-
-var (
-	ErrDeploymentAlreadyExists = errors.New("keeper: deployment already exists")
-	ErrDeploymentNotFound      = errors.New("keeper: deployment not found")
 )
 
 // Keeper of the deployment store
@@ -97,7 +91,7 @@ func (k Keeper) Create(ctx sdk.Context, deployment types.Deployment, groups []ty
 	key := deploymentKey(deployment.ID())
 
 	if store.Has(key) {
-		return ErrDeploymentAlreadyExists
+		return types.ErrDeploymentExists
 	}
 
 	store.Set(key, k.cdc.MustMarshalBinaryBare(deployment))
@@ -123,7 +117,7 @@ func (k Keeper) UpdateDeployment(ctx sdk.Context, deployment types.Deployment) e
 	key := deploymentKey(deployment.ID())
 
 	if !store.Has(key) {
-		return ErrDeploymentNotFound
+		return types.ErrDeploymentNotFound
 	}
 
 	ctx.EventManager().EmitEvent(
@@ -131,6 +125,24 @@ func (k Keeper) UpdateDeployment(ctx sdk.Context, deployment types.Deployment) e
 	)
 
 	store.Set(key, k.cdc.MustMarshalBinaryBare(deployment))
+	return nil
+}
+
+// OnCloseGroup provides shutdown API for a Group
+func (k Keeper) OnCloseGroup(ctx sdk.Context, group types.Group) error {
+	store := ctx.KVStore(k.skey)
+	key := groupKey(group.ID())
+
+	if !store.Has(key) {
+		return types.ErrGroupNotFound
+	}
+	group.State = types.GroupClosed
+
+	ctx.EventManager().EmitEvent(
+		types.EventGroupClose{ID: group.ID()}.ToSDKEvent(),
+	)
+
+	store.Set(key, k.cdc.MustMarshalBinaryBare(group))
 	return nil
 }
 
--- a/x/deployment/keeper/keeper_test.go
+++ b/x/deployment/keeper/keeper_test.go
@@ -221,6 +221,53 @@ func Test_OnDeploymentClosed(t *testing.T) {
 	})
 }
 
+func Test_CloseGroup(t *testing.T) {
+	ctx, keeper := setupKeeper(t)
+	groups := createActiveDeployment(t, ctx, keeper)
+
+	t.Run("assert group 0 state closed", func(t *testing.T) {
+		assert.NoError(t, keeper.OnCloseGroup(ctx, groups[0]))
+		group, ok := keeper.GetGroup(ctx, groups[0].ID())
+		assert.True(t, ok)
+		assert.Equal(t, types.GroupClosed, group.State)
+
+		keeper.OnDeploymentClosed(ctx, group) // coverage: catch an additional code path
+		assert.Equal(t, types.GroupClosed, group.State)
+	})
+	t.Run("group 1 matched-state not-orderable", func(t *testing.T) {
+		group := groups[1]
+		assert.Equal(t, group.ValidateOrderable(), types.ErrGroupNotOpen, group.State.String())
+	})
+}
+
+func Test_CloseOpenGroups(t *testing.T) {
+	ctx, keeper := setupKeeper(t)
+	groups := createDeploymentsWithState(t, ctx, types.GroupOpen, keeper)
+
+	t.Run("assert group 0 state closed", func(t *testing.T) {
+		assert.NoError(t, keeper.OnCloseGroup(ctx, groups[0]))
+		group, ok := keeper.GetGroup(ctx, groups[0].ID())
+		assert.True(t, ok)
+		assert.Equal(t, types.GroupClosed, group.State)
+	})
+	t.Run("group 1 matched-state still orderable", func(t *testing.T) {
+		group := groups[1]
+		assert.Equal(t, group.ValidateOrderable(), nil, group.State.String())
+	})
+}
+
+func Test_Empty_CloseGroup(t *testing.T) {
+	ctx, keeper := setupKeeper(t)
+	group := types.Group{
+		GroupID: testutil.GroupID(t),
+	}
+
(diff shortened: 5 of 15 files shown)


x/deployment/types/msgs.go
```diff
@@ -8,6 +8,8 @@ const (
 	msgTypeCreateDeployment = "create-deployment"
 	msgTypeUpdateDeployment = "update-deployment"
 	msgTypeCloseDeployment  = "close-deployment"
+
```
boz, line 11: :nail_care: no newline so that it's aligned

x/deployment/types/types.go
```diff
@@ -65,6 +65,15 @@ var GroupStateMap = map[string]GroupState{
 	"closed":       GroupClosed,
 }
 
+// GroupStateToReadable returns the adverb of the state
+func GroupStateToReadable(g GroupState) string {
+	m := make(map[GroupState]string, len(GroupStateMap))
+	for k, v := range GroupStateMap {
+		m[v] = k
+	}
+	return m[g]
```
boz, line 74: Let's just use [stringer](https://godoc.org/golang.org/x/tools/cmd/stringer).

with stringer this is moot, but you don't need to make complete copy.

```go
for k, v := range GroupStateMap {
  if v == g {
    return k
  }
}
return "UNKNOWN"
```

x/deployment/handler/handler.go
```diff
@@ -100,3 +102,26 @@ func handleMsgClose(ctx sdk.Context, keeper keeper.Keeper, mkeeper MarketKeeper,
...
 }
+
+func handleMsgCloseGroup(ctx sdk.Context, keeper keeper.Keeper, mkeeper MarketKeeper, msg types.MsgCloseGroup) (*sdk.Result, error) {
+	group, found := keeper.GetGroup(ctx, msg.ID)
+	if !found {
+		return nil, types.ErrGroupNotFound
+	}
+
+	// if Group already closed; return an error
+	if group.State == types.GroupClosed {
```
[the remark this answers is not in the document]
Ropes: Using ValidateOrderable breaks the simulations because it returns ErrGroupClosed for all states other than "open". Checking with the other calls to ValidateOrderable, I don't think it's safe to change behavior of the function and use it in the simulation operation check.

x/deployment/types/types.go
Ropes: Updated in code, an issue created: https://github.com/ovrclk/akash/issues/665

x/deployment/handler/handler.go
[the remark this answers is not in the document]
boz: _make_ a `group.ValidateCloseable()` ;)

x/deployment/keeper/keeper.go
```diff
@@ -97,7 +91,7 @@ func (k Keeper) Create(ctx sdk.Context, deployment types.Deployment, groups []ty
 	key := deploymentKey(deployment.ID())
 
 	if store.Has(key) {
-		return ErrDeploymentAlreadyExists
+		return types.ErrDeploymentExists
```
boz, line 94: :raised_hands:

x/deployment/keeper/keeper_test.go
```diff
@@ -68,6 +68,19 @@ func Test_Create(t *testing.T) {
 		_, ok := keeper.GetGroup(ctx, testutil.GroupID(t))
 		assert.False(t, ok)
 	})
+
+	t.Run("close the first Group", func(t *testing.T) {
+		if len(groups) > 0 {
```
boz, line 73: this is always true.

do a `require.NotEmpty(t,groups)` if you really want to for clarity, it'd the same as the `if {} else {}` logic here.

boz: Actually, I think that since you have `Test_CloseGroup` below, none of these need to be here - leave `Test_Create` as it is.

x/deployment/keeper/keeper_test.go
```diff
@@ -210,6 +223,41 @@ func Test_OnDeploymentClosed(t *testing.T) {
 	})
 }
 
+func Test_CloseGroup(t *testing.T) {
+	ctx, keeper := setupKeeper(t)
+	groups := createActiveDeployment(t, ctx, keeper)
+	gIdx0, gIdx1 := 0, 1
+
+	t.Run("close-group-0", func(t *testing.T) {
+		assert.NoError(t, keeper.OnCloseGroup(ctx, groups[gIdx0]))
```
boz, line 232: This doesn't really need to be in a `t.Run()`.

```go
// do the action
assert.NoError(t, keeper.OnCloseGroup(ctx, groups[gIdx0]))

// assert affects of the action
```
I think you're overly-concerned about annotating failures - line numbers are enough.

x/deployment/keeper/keeper_test.go
```diff
@@ -210,6 +223,41 @@ func Test_OnDeploymentClosed(t *testing.T) {
 	})
 }
 
+func Test_CloseGroup(t *testing.T) {
+	ctx, keeper := setupKeeper(t)
+	groups := createActiveDeployment(t, ctx, keeper)
+	gIdx0, gIdx1 := 0, 1
```
boz, line 229: why?

x/deployment/keeper/keeper_test.go
```diff
@@ -210,6 +223,41 @@ func Test_OnDeploymentClosed(t *testing.T) {
...
+
+	t.Run("close-group-0", func(t *testing.T) {
+		assert.NoError(t, keeper.OnCloseGroup(ctx, groups[gIdx0]))
+	})
+	t.Run("group-1-open", func(t *testing.T) {
+		s := groups[gIdx1]
+		t.Logf("%v", s)
```
boz, line 236: ?

Please just follow the pattern from [`Test_OnLeaseClosed`](https://github.com/ovrclk/akash/blob/f42f3233705f09750195a04a659d97c2ec7e55a2/x/deployment/keeper/keeper_test.go#L173-L191) for this entire test function.

x/deployment/types/types_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"strconv"
+	"testing"
+
+	"github.com/ovrclk/akash/x/deployment/types"
+	"github.com/stretchr/testify/assert"
+)
+
+type GStateTest struct {
```
boz, line 11: :nail_care: no need for `Export`.

x/deployment/types/types_test.go
```diff
@@ -0,0 +1,53 @@
...
+			expStateStr: "closed",
+		},
+		{
+			state:       types.GroupState(99),
+			expStateStr: "UNKNOWN",
+		},
+	}
+
+	for i, test := range tests {
+		t.Run(strconv.Itoa(i), test.Test())
```
boz, line 51: this is super duper unnecessary.  try to not use a bunch of machinery in tests when it can be done in a simpler way.

```go
tests := []struct{
  state types.GroupState
  val   string
} {
  { types.GroupMatched, "matched" },
  { types.GroupOrdered, "ordered" },
  // ...
}

for _, test := range tests {
  assert.Equal(t,t.val,types.GroupStateToReadable(t.state))
}
```

_run/lite/Makefile
```diff
@@ -87,10 +96,16 @@ bid-close:
...
+	--dseq "$(DSEQ)"  \
+	--gseq "$(GSEQ)"
+
+group-close:
+	$(akashctl) tx $(KEY_OPTS) $(CHAIN_OPTS) deployment group close \
+	--owner $(MAIN_ADDR) \
 	--from main \
 	--dseq "$(DSEQ)"  \
 	--gseq "$(GSEQ)"  \
-	--oseq "$(OSEQ)"
```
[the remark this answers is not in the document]
boz: I think the last line of the command still ends in a `\`

Makefile
```diff
@@ -226,3 +226,5 @@ test-sim-after-import:
 	@echo "Running application simulation-after-import..."
 	go test -mod=readonly $(APP_DIR) -run=TestAppSimulationAfterImport -Enabled=true \
 		-NumBlocks=50 -BlockSize=100 -Commit=true -Seed=99 -Period=5 -v -timeout 10m
+
+test-sims: test-sim-fullapp test-sim-nondeterminism test-sim-import-export test-sim-after-import
```
boz, line 230: :100:

x/deployment/handler/handler.go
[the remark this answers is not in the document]
Ropes: I created ValidateClosable().

To that point, should we just make all the fields of Group private? Add getters to the fields?

x/deployment/keeper/keeper_test.go
Ropes: Removed. I was thinking in terms of further referencing, but agree `group := group[0]` is a better way to access for these cases.

x/deployment/types/types_test.go
Ropes: Alright, I added additional assertions to the `gStateTest struct Test()func(t*testing.T)` method, to work with our linter and t.Run. If you want all of the assertions to be with in the `for { }` block, though, then t.Run chokes on passage of the `test` variable. 

So I removed all of that and left it as a single testing.T that iterates and logs initial conditions. That's the only way to have more than one assertion, and not violate the `scopelinter` + `t.Run()` without returning a test function from the struct AFAIK.

_run/lite/Makefile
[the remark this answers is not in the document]
Ropes: https://www.gnu.org/software/make/manual/html_node/Splitting-Recipe-Lines.html#:~:text=5.1.,1%20Splitting%20Recipe%20Lines&text=As%20in%20normal%20makefile%20syntax,be%20invoked%20to%20run%20it.

Makefile docs show the last line doesn't require another newline to be added.

This is the pattern I've always seen:
```
.PHONY: build-k8s
build-k8s: $(K8S_BUILD_DIR)
	@for file in $(K8S_FILES); do \
		mkdir -p `dirname "$(K8S_BUILD_DIR)/$$file"` ; \
		$(SHELL_EXPORT) envsubst <$(K8S_DIR)/$$file >$(K8S_BUILD_DIR)/$$file ;\
	done
```
Last line doesn't need a `\`: https://blog.zikes.me/post/config-k8s-with-make/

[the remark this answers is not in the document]
boz: exactly, the last line shouldn't have a `\`.  it looks like with this change the last line has a `\` - it is

```
   --gseq "$(GSEQ)"  \
```

it only works because there happens to be an empty line below.

x/deployment/types/types.go
```diff
@@ -124,20 +128,31 @@ type Group struct {
...
 	default:
 		return ErrGroupNotOpen
 	}
 }
 
+// ValidateClosable provides error response if group is already closed,
+// and thus should not be closed again, else nil.
+func (g Group) ValidateClosable() error {
+	switch g.State {
+	case GroupClosed:
```
boz, line 149: Can you make a design ticket to decide if `GroupInsufficientFunds` should be included here?

x/deployment/types/types_test.go
```diff
@@ -0,0 +1,71 @@
...
+				GSeq:  id.GSeq,
+			},
+			State: test.state,
+		}
+
+		assert.NotEmpty(t, group.ID().Owner)
+
+		assert.Equal(t, group.ValidateOrderable(), test.expValidateOrderable)
+
+		assert.Equal(t, group.ValidateClosable(), test.expValidateClosable)
```
boz, line 69: Any reason why you don't just use `GroupID: id`?

Look at [`assert.Equal`](https://godoc.org/github.com/stretchr/testify/assert#Equal) and [`assert.Equalf`](https://godoc.org/github.com/stretchr/testify/assert#Equalf).

1. the "expected" value is the second argument.
1. you can add messages to both `Equal` and `Equalf`

for example:

```
assert.Equal(t, test.expValidateOrderable, group.ValidateOrderable(), test.expStateStr)
```

x/deployment/client/cli/tx.go
```diff
@@ -135,3 +136,50 @@ func cmdUpdate(key string, cdc *codec.Codec) *cobra.Command {
...
+	)...)
+	return cmd
+}
+
+func cmdGroupClose(_ string, cdc *codec.Codec) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "close",
+		Short:   "close a Deployment's specific Group",
+		Example: "akashctl tx deployment group close --owner=[Account Address] --dseq=[uint64] --gseq=[uint32]",
+		//Args:    cobra.ExactArgs(1),
```
boz, line 161: `cobra.ExactArgs(0)`

x/deployment/types/types_test.go
boz: Thanks.  Yeah, the `t.Run(name,test.generate())` thing is an overkill here.

x/deployment/handler/handler.go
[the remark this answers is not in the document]
boz: Thanks.  We can't, unfortunately - private fields wouldn't be serialized.

x/deployment/types/types.go
Ropes: https://github.com/ovrclk/akash/issues/684

x/deployment/client/cli/tx.go
```diff
@@ -135,3 +136,50 @@ func cmdUpdate(key string, cdc *codec.Codec) *cobra.Command {
 	AddDeploymentIDFlags(cmd.Flags())
 	return cmd
 }
+
+// getTxGroupCmd returns the Group modification commands
+func getTxGroupCmd(key string, cdc *codec.Codec) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                        "group",
```
boz, line 143: I think we should follow the pattern of the other ones here - `group-close` (one command).

It's 100% debatable whether or not it's the right pattern, but let's at least keep them consistent.

x/deployment/keeper/keeper_test.go
```diff
@@ -211,6 +211,35 @@ func Test_OnDeploymentClosed(t *testing.T) {
...
+	t.Run("assert group 0 state closed", func(t *testing.T) {
+		assert.NoError(t, keeper.OnCloseGroup(ctx, groups[0]))
+		group, ok := keeper.GetGroup(ctx, groups[0].ID())
+		assert.True(t, ok)
+		assert.Equal(t, types.GroupClosed, group.State)
+	})
+	t.Run("group 1 matched-state not-orderable", func(t *testing.T) {
+		group := groups[1]
+		t.Logf("%v", types.GroupStateToReadable(group.State))
+		assert.Equal(t, group.ValidateOrderable(), types.ErrGroupNotOpen)
```
boz, line 227: let's keep the printf debugging out of commits.

I like how this is checking the _behavior_ -- "is it orderable?" -- instead of the precise state.  That said, in this case we want to ensure that the state hasn't changed (checking accidental closing of all groups).  `OnCloseGroup(ctx,groups[0])` doesn't affect whether or not `groups[1]` is orderable (it isn't orderable before and it isn't orderable after).

x/deployment/types/types.go
```diff
@@ -124,20 +128,31 @@ type Group struct {
 }
 
 // ID method returns GroupID details of specific group
-func (d Group) ID() GroupID {
-	return d.GroupID
+func (g Group) ID() GroupID {
```
boz, line 131: :+1:

x/deployment/types/types_test.go
```diff
@@ -0,0 +1,71 @@
...
+			expValidateOrderable: types.ErrGroupNotOpen,
+		},
+	}
+
+	for i, test := range tests {
+		t.Logf("------test-%d: %#v", i, test)
+		assert.Equal(t, types.GroupStateToReadable(test.state), test.expStateStr)
+		id := testutil.GroupID(t)
+		group := types.Group{
+			GroupID: types.GroupID{
```
boz, line 57: `GroupID: testutil.GroupID(t)`

x/deployment/client/cli/tx.go
Ropes: I'm a little confused as to what `group-close` as one command is consistent with? Typically our CLI structure follows the pattern `<uttility group> <object...> <action verb>`. eg `akashctl tx deployment close` 

Or just because `group-close` is tightly coupled to a `deployment`?

x/deployment/keeper/keeper_test.go
Ropes: The reason neither group are 'open', is that `createActiveDeployment()` initializes two groups in state 'matched', so they're not "Orderable" from the start. 

I've create an additional test to assert the case you outlined of closing one group and not affecting sibling groups.

x/deployment/client/cli/tx.go
boz: If by utility group you mean `tx`, `query`, etc... it's:

```
<utility-group> <module> <...>
```

for `query` commands things are mostly as you say but that's not the case right now for `tx` commands:

```
/akash ❯❯❯ ./akashctl help tx market                                                                                                                                                master ✭ ◼
Transaction subcommands

Usage:
  akashctl tx market [flags]
  akashctl tx market [command]

Available Commands:
  bid-close   Close a market bid
  bid-create  Create a market bid
  order-close Close a market order
```

x/deployment/keeper/keeper_test.go
boz: > The reason neither group are 'open', is that createActiveDeployment() initializes two groups in state 'matched', so they're not "Orderable" from the start.

exactly, hence

> it isn't orderable before and it isn't orderable after

x/deployment/client/cli/tx.go
Ropes: ahh okay, thanks. I'll drop the extra command group.

x/deployment/types/types.go
```diff
@@ -67,6 +61,16 @@ var GroupStateMap = map[string]GroupState{
 	"closed":             GroupClosed,
 }
 
+// GroupStateToReadable returns the adverb of the state
+func GroupStateToReadable(g GroupState) string {
```
boz, line 65: this shouldn't be needed anymore - we have the [stringer](https://github.com/ovrclk/akash/blob/8d399ba1e12bd6ae3be55718b6ebb2d91ccce71a/x/deployment/types/autogen_stringer.go#L41) for it.

x/deployment/types/types_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	for i, test := range tests {
+		t.Logf("------test-%d: %#v", i, test)
+		assert.Equal(t, types.GroupStateToReadable(test.state), test.expStateStr)
+		group := types.Group{
+			GroupID: testutil.GroupID(t),
+			State:   test.state,
+		}
+
+		assert.NotEmpty(t, group.ID().Owner)
```
boz, line 60: unnecessary

x/deployment/keeper/keeper_test.go
```diff
@@ -211,6 +211,53 @@ func Test_OnDeploymentClosed(t *testing.T) {
...
+		group, ok := keeper.GetGroup(ctx, groups[0].ID())
+		assert.True(t, ok)
+		assert.Equal(t, types.GroupClosed, group.State)
+
+		keeper.OnDeploymentClosed(ctx, group) // coverage: catch an additional code path
+		assert.Equal(t, types.GroupClosed, group.State)
+	})
+	t.Run("group 1 matched-state not-orderable", func(t *testing.T) {
+		group := groups[1]
+		assert.Equal(t, group.ValidateOrderable(), types.ErrGroupNotOpen, group.State.String())
```
boz, line 229: so, again: this is checking that it's not orderable, but it's not making sure that it's state isn't changing.

https://github.com/ovrclk/akash/pull/657#discussion_r445089848

I see the new test to make sure ancillary groups don't go from orderable -> not-orderable.  this one should be making sure that the state stays in `matched`.

You know what - forget it.  never mind.

x/deployment/types/types.go
```diff
@@ -20,10 +18,6 @@ const (
 	DeploymentClosed // closed
 )
 
-var (
-	ErrGroupNotOpen = errors.New("group not open")
```
boz, line 24: :+1:

x/deployment/types/types.go
Ropes: good point, cleans up the tests too

PR #659, opened 2020-06-12
Modify provider host uri validation

### LINKED ISSUE #629: x/provider: validate `HostURI` - no path allowed
The [`Provider.HostURI`](https://github.com/ovrclk/akash/blob/0dfd6b825feaff2594a9fbf0f14786859ae4beb6/x/provider/types/types.go#L10) field declares the provider's [gateway API](https://github.com/ovrclk/akash/blob/0dfd6b825feaff2594a9fbf0f14786859ae4beb6/provider/gateway/server.go#L12) address.

It should have

* scheme: `http` or `https`
* hostname
* port: optional
* path: empty

akhilkumarpilli:
Fixes #629


x/provider/types/msgs.go
```diff
@@ -111,5 +111,19 @@ func validateProviderURI(val string) error {
...
+	if u.Scheme != "http" && u.Scheme != "https" {
+		return errors.Wrapf(ErrInvalidProviderURI, "scheme in %q should be http or https", val)
+	}
+
+	if u.Host == "" {
+		return errors.Wrapf(ErrInvalidProviderURI, "validating %q for valid host", val)
+	}
+
+	// Considering "/" path as emp&&
+	if u.Path != "" && u.Path != "/" {
```
boz, line 124: I think we should just not allow the `"/"` to keep things easier.

x/provider/types/msgs_test.go
```diff
@@ -24,17 +24,17 @@ func TestConfigPath(t *testing.T) {
...
+		},
 		{
 			path:   "file:///foo.yml",
-			expErr: nil,
+			expErr: ErrInvalidProviderURI,
 		},
 		{
 			path:   "http://localhost:80/foo",
-			expErr: nil,
+			expErr: ErrInvalidProviderURI,
```
boz, line 37: I think we need to add some more tests here.  Can you make sure the relevant entries from the table in [this comment](https://github.com/ovrclk/akash/pull/612#discussion_r428983631) are tested?

troian: Wouldn't regex be easier? https://regex101.com/r/Xc2w0v/2/

PR #666, opened 2020-06-15
Add shellcheck linter

### LINKED ISSUE #641: ci: add `shellcheck` linter
Use the `shellcheck` linter for all scripts in the repo.

* [ ] ensure all scripts end in `.sh` and run `shellcheck` on them in the following:
* [ ] `make` target for running locally (probably use Docker to make install easy)
* [ ] github workflow that runs shellcheck

akhilkumarpilli:
Fixes #641


godownloader.sh
```diff
@@ -343,7 +343,7 @@ EOF
 PROJECT_NAME="akash"
 OWNER=ovrclk
 REPO="akash"
-BINARY=akashctl
+#BINARY=akashctl
```
boz, line 346: I think we should skip this file; it's auto-generated by godownloader.

Makefile
```diff
@@ -62,6 +62,13 @@ release:
 image-minikube:
 	eval $$(minikube docker-env) && make image
 
+shellcheck:
+	docker run --rm --privileged \
```
boz, line 66: does it have to be priviliged?

akhilkumarpilli: No, I just kept it as privileged by looking other commands in Makefile. I will remove it.

PR #669, opened 2020-06-15
fix: reject updates on provider attributes

### LINKED ISSUE #643: design: should provider attribute updates affect current leases?
Currently, if a provider changes their attributes such that they are incompatible with leases that the provider has open, the existing leases are not affected.

We have three options:

1. Status quo
1. Should existing, incompatible leases be closed?
1. Should we disallow incompatible updates?

I prefer the third option because it forces providers to understand what they are doing while at the same time keeping committed state consistent with an invariant that the provider's attributes are compatible with an order's group attributes for all open leases.

troian:
fixes #643

### DIFF
--- a/x/provider/handler/handler.go
+++ b/x/provider/handler/handler.go
@@ -5,18 +5,20 @@ import (
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/pkg/errors"
 
+	mkeeper "github.com/ovrclk/akash/x/market/keeper"
+	mtypes "github.com/ovrclk/akash/x/market/types"
 	"github.com/ovrclk/akash/x/provider/keeper"
 	"github.com/ovrclk/akash/x/provider/types"
 )
 
 // NewHandler returns a handler for "provider" type messages.
-func NewHandler(keeper keeper.Keeper) sdk.Handler {
+func NewHandler(keeper keeper.Keeper, mkeeper mkeeper.Keeper) sdk.Handler {
 	return func(ctx sdk.Context, msg sdk.Msg) (*sdk.Result, error) {
 		switch msg := msg.(type) {
 		case types.MsgCreateProvider:
 			return handleMsgCreate(ctx, keeper, msg)
 		case types.MsgUpdateProvider:
-			return handleMsgUpdate(ctx, keeper, msg)
+			return handleMsgUpdate(ctx, keeper, mkeeper, msg)
 		case types.MsgDeleteProvider:
 			return handleMsgDelete(ctx, keeper, msg)
 		default:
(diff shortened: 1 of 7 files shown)


boz:
@troian I'm sorry this wasn't clear - we want to prohibit attribute changes that are incompatible with active leases that the provider has.

So if a provider p1 has attributes

```
region: us-west
class: b
```

and two leases for groups that have required attributes:

```
- id: 1
  attributes:
    region: us-west
    class: b
- id: 2
  attributes:
    region: us-west
```

the provider would not be able to remove its attribute `class: b` until lease 1 is closed.

does that make sense?

troian:
Gotcha

x/provider/handler/handler.go
```diff
@@ -44,16 +46,35 @@ func handleMsgCreate(ctx sdk.Context, keeper keeper.Keeper, msg types.MsgCreateP
...
+	prov, found := keeper.Get(ctx, msg.Owner)
+	if !found {
 		return nil, errors.Wrapf(types.ErrProviderNotFound, "id: %s", msg.Owner)
 	}
 
+	attributeChangesPermitted := true
+
+	// all filtering code below is madness!. should make an index to not melt the cpu
+	mkeeper.WithLeases(ctx, func(lease mtypes.Lease) bool {
+		if prov.Owner.Equals(lease.Provider) && lease.State != mtypes.LeaseClosed {
```
boz, line 59: I believe that `InsufficientFunds` is a terminal state too.  I think this would be safer:

```go
if !(prov.Owner.Equals(lease.Provider) && lease.State == mtypes.Leaseactive) {
  return false
}
```

x/provider/handler/handler.go
```diff
@@ -44,16 +46,35 @@ func handleMsgCreate(ctx sdk.Context, keeper keeper.Keeper, msg types.MsgCreateP
...
+			// shall we manifest a bit of wonder in case order is not found?
```
boz, line 60: maybe, just to be safe, I guess.

x/provider/types/errors.go
```diff
@@ -31,4 +32,7 @@ var (
 
 	// ErrAttributes error code for provider attribute problems
 	ErrAttributes = sdkerrors.Register(ModuleName, errAttributes, "attribute specification error")
+
+	// ErrAttributesImmutable error code for attributes update
+	ErrAttributesImmutable = sdkerrors.Register(ModuleName, errAttributesImmutable, "attributes cannot be changed")
```
boz, line 37: `ErrIncompatibleAttributes` or something?

x/provider/handler/handler.go
boz: maybe create and use `mkeeper.LeasesForProvider()` so we can better keep track of where indexes are needed and to abstract away the use of those indexes.

troian: Maybe better to keep it as is. actual filtering is provider specific only. It introduces double iteration through leases.

troian: won't work
```go
if !(prov.Owner.Equals(lease.Provider) && lease.State == mtypes.Leaseactive) {
  return false
}
```

this should do
```go
if prov.Owner.Equals(lease.Provider) && (lease.State == mtypes.LeaseActive) {
  return false
}
```

boz: it's just generally nicer to exit early from things and continue on with what you want to process.

```go
// if not interested in this lease, continue on to the next
if !(prov.Owner.Equals(lease.Provider) && lease.State == mtypes.Leaseactive) {
  return false
}
order, found := mkeeper.GetOrder(...)
// ...
return true
```

but all good, just a preference.  it works great as is.

troian: Yah, my bad, should reorder

PR #677, opened 2020-06-18
Improve validation of address with VerifyAddressFormat

### LINKED ISSUE #571: x/: improve validation by parsing addresses

akhilkumarpilli:
Fixes #571

### DIFF
--- a/x/deployment/query/types.go
+++ b/x/deployment/query/types.go
@@ -18,9 +18,11 @@ type DeploymentFilters struct {
 	State types.DeploymentState
 }
 
+// Accept returns whether deployment filters valid or not
 func (filters DeploymentFilters) Accept(obj types.Deployment, isValidState bool) bool {
-	if (filters.Owner.Empty() && !isValidState) ||
-		(filters.Owner.Empty() && (obj.State == filters.State)) ||
+	err := sdk.VerifyAddressFormat(filters.Owner)
+	if (err != nil && !isValidState) ||
+		(err != nil && (obj.State == filters.State)) ||
 		(!isValidState && (obj.DeploymentID.Owner.Equals(filters.Owner))) ||
 		(obj.DeploymentID.Owner.Equals(filters.Owner) && obj.State == filters.State) {
 		return true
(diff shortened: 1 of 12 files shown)


x/deployment/query/types.go
```diff
@@ -18,9 +18,11 @@ type DeploymentFilters struct {
 	State types.DeploymentState
 }
 
+// Accept returns whether deployment filters valid or not
 func (filters DeploymentFilters) Accept(obj types.Deployment, isValidState bool) bool {
-	if (filters.Owner.Empty() && !isValidState) ||
-		(filters.Owner.Empty() && (obj.State == filters.State)) ||
+	err := sdk.VerifyAddressFormat(filters.Owner)
```
boz, line 23: we need to test all of these query filters thoroughly; maybe leave it out of this PR.

I think that this isn't the semantics we want - if it's an empty address then don't filter on the address, if it's a _bad_ address then return error.

akhilkumarpilli:
> Looking great, thank you.
> 
> Can you make sure that there's adequate test coverage for the message validations?
> 
> See my note about the filters - I think that they can probably be left out of this PR. The semantics of querying should be:
> 
>     1. if the address is empty, don't filter on address
> 
>     2. if the address is non-empty and invalid, return an error
> 
>     3. if the address is non-empty and valid, filter on the address

Added few more cli tests for testing filters. Also verifying given owner address and if invalid, throwing error. is

[This](https://github.com/ovrclk/akash/blob/master/x/deployment/query/types.go#L21) will be same as I am validating given owner address before coming to this function.

boz:
> [This](https://github.com/ovrclk/akash/blob/master/x/deployment/query/types.go#L21) function will not change as I am validating given owner address before coming to this function.

Perfect, thank you.

PR #685, opened 2020-06-19
feat: _run/lite - remove "stake" token

### LINKED ISSUE #678: remove stake token from _run/lite
Will need to replace `stake` with `akash` in the genesis during bootstrap.

boz:
fixes #678

### DIFF
--- a/_run/lite/Makefile
+++ b/_run/lite/Makefile
@@ -3,24 +3,30 @@ include ../common.mk
 akashctl       = $(AKASHCTL) --home $(CLIENT_HOME) 
 akashd         = $(AKASHD) --home $(NODE_HOME)
 KEY_OPTS       = --keyring-backend=test
-CHAIN_NAME     = local
-CHAIN_OPTS     = --chain-id $(CHAIN_NAME)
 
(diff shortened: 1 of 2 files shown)


_run/lite/Makefile
```diff
@@ -3,24 +3,30 @@ include ../common.mk
 akashctl       = $(AKASHCTL) --home $(CLIENT_HOME) 
 akashd         = $(AKASHD) --home $(NODE_HOME)
 KEY_OPTS       = --keyring-backend=test
-CHAIN_NAME     = local
-CHAIN_OPTS     = --chain-id $(CHAIN_NAME)
 
 init: bins init-keys init-genesis init-gentx init-collect-gentx init-config
```
troian, line 7: add `init-node`?

boz: probably but I'm not sure.  have to reconcile w/ other `_run/` dirs.

boz:
@Ropes recipe for removing stake here.  `sed -i.orig '/"stake"/"akash"/g' genesis.gson` or whatever would probably be fine too.  The only real trick is that you have to override the default `--amount` in `gentx`.

PR #688, opened 2020-06-20
Use stringer for static type conversions

### LINKED ISSUE #665: feat: Use Stringer code generator for static type conversions
There are multiple type mappings where a string is converted into a static type or converse. 

Example of mapping strings to types in one function, and then needing to have inverse functionality. https://github.com/ovrclk/akash/pull/657#discussion_r439154127

Generating the inverse lookup code statically with [stringer](https://godoc.org/golang.org/x/tools/cmd/stringer).

### TODO
* [ ] Add stringer to build tools
  * Vendor vs local install?
* [ ] Mark codepaths needing to be generated
* [ ] Tooling to generate static code

akhilkumarpilli:
Fixes #665


x/deployment/types/types.go
```diff
@@ -45,15 +47,15 @@ type GroupState uint8
...
+	GroupOpen GroupState = iota + 1 // open
 	// GroupOrdered is used when state of group is ordered
-	GroupOrdered
+	GroupOrdered // ordered
 	// GroupMatched is used when state of group is matched
-	GroupMatched
+	GroupMatched // matched
 	// GroupInsufficientFunds is used when group has insufficient funds
-	GroupInsufficientFunds
+	GroupInsufficientFunds // insufficient
```
boz, line 56: can you make it `// insufficient-funds`?

Makefile
```diff
@@ -24,6 +24,12 @@ bins: $(BINS)
 build:
 	$(GO) build ./...
 
+generate: stringer
+	$(GO) generate ./...
+
+stringer:
+	$(GO) install golang.org/x/tools/cmd/stringer
```
boz, line 31: can you add this to [`deveps-install`](https://github.com/ovrclk/akash/blob/2859bc805774ddfded6336d8871bb03505832079/Makefile#L135-L138) instead of making a new target?

Also please add `golang.org/x/tools/cmd/stringer` to [`tools.go`](https://github.com/ovrclk/akash/blob/2859bc805774ddfded6336d8871bb03505832079/tools.go#L7-L9)

x/market/types/state.go
```diff
@@ -0,0 +1,67 @@
+// Code generated by "stringer -linecomment -output=state.go -type=OrderState,BidState,LeaseState"; DO NOT EDIT.
```
boz, line 1: can you make the file name `autogen_stringer.go` for all of the generated files here?  it makes it easier to quickly know what is auto-generated and what isn't for both us reading it and for doing things like skipping these files during linting.

x/market/types/types.go
```diff
@@ -145,11 +147,11 @@ type LeaseState uint8
 
 const (
 	// LeaseActive is used when state of lease is active
-	LeaseActive LeaseState = iota + 1
+	LeaseActive LeaseState = iota + 1 // active
 	// LeaseInsufficientFunds is used when lease has insufficient funds
-	LeaseInsufficientFunds
+	LeaseInsufficientFunds // insufficient
```
boz, line 152: similar - `insufficient-funds`

boz:
I'd like to merge this shortly @Ropes @troian; please take a look if you can.

PR #700, opened 2020-06-23
configure mainnet using build tags

### LINKED ISSUE #683: mainnet branch: possible to do with build tags?
I think we may be able to forgo the headache of a "mainnet" branch by using build tags - split mainnet-speciality into a separate file.

rough illustration:

`app/app_configure_mainnet.go`:
```go
// +build mainnet

func addAkashModules(...) {
  // probably skip things
}
```

`app/app_configure.go`:
```go
// +build !mainnet

func addAkashModules(...) {
  // add our modules
}
```

akhilkumarpilli:
Fixes #683


akhilkumarpilli:
@anilCSE @boz , seperated mainnet using build tags. Integration tests are running fine when using `mainnet` tag. Able to build and install akash binaries with `mainnet` tag. But sim tests and unit tests are not working when running with `mainnet` tag. I am getting this error for those tests:

```
can't load package: package .: build constraints exclude all Go files in ~/go/src/github.com/ovrclk/akash
Makefile:222: recipe for target 'test-sim-fullapp' failed
make: *** [test-sim-fullapp] Error 1
```

anilCSE:
@akhilkumarpilli looks good, file namings are bit odd. Lets think of different namings. How about `config/default.go` and `config/mainnet.go`?

.github/workflows/integration.yaml
```diff
@@ -14,4 +14,10 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
       - run: make test-integration
+  integration-mainnet:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+      - run: export MAINNET=true && make test-integration
```
boz, line 22: :nail_care: you can just do `MAINNET=true make test-integration`

Makefile
```diff
@@ -5,11 +5,18 @@ APP_DIR := ./app
 GO := GO111MODULE=on go
 GOBIN := $(shell go env GOPATH)/bin
 
+# Setting mainnet flag based on env value
+# export MAINNET=true to set build tag mainnet
+BUILD_MAINNET = !mainnet
```
boz, line 10: I think this can just be empty - you don't need `!mainnet`.

Makefile
```diff
@@ -143,7 +150,7 @@ devdeps-install:
 
 test-integration: $(BINS)
 	cp akashctl akashd ./_build
-	go test -mod=readonly -p 4 -tags=integration -v ./integration/...
+	go test -mod=readonly -p 4 -tags=integration,$(BUILD_MAINNET) -v ./integration/...
```
boz, line 153: can probably quote and separate with a space like on [line 19](https://github.com/ovrclk/akash/pull/700/files#diff-b67911656ef5d18c4ae36cb6741b7965R19) so that you don't have to worry about the `,`.

app/app.go
```diff
@@ -97,203 +96,14 @@ func NewApp(
 		tkeys:          tkeys,
 	}
 
-	app.keeper.params = params.NewKeeper(
```
boz, line 100: is it possible to keep the common parts here and use the functions to add the tag-specific values?

app/app.go
```diff
@@ -97,203 +96,14 @@ func NewApp(
...
-
-		market.NewAppModule(
-			app.keeper.market,
-			app.keeper.deployment,
-			app.keeper.provider,
-			app.keeper.bank,
-		),
-
-		provider.NewAppModule(app.keeper.provider, app.keeper.bank, app.keeper.market),
-	)
```
boz, line 254: same here.  for instance, does `ModuleManager` have an `AppendModule()` function or something that can be used in the tag-specific functions?

akhilkumarpilli:
> @akhilkumarpilli looks good, file namings are bit odd. Lets think of different namings. How about `config/default.go` and `config/mainnet.go`?

If we rename files to `config/default.go` and `config/mainnet.go`, then package will change. Then I think it will be quite difficult to understand code.

app/app.go
akhilkumarpilli: I am not able to find a way to append modules in ModuleManager(). Tried to reduce the amount of duplicate code and left with these.

app/app.go
```diff
@@ -195,105 +193,22 @@ func NewApp(
...
-		supply.NewAppModule(app.keeper.supply, app.keeper.acct),
-		mint.NewAppModule(app.keeper.mint),
-		staking.NewAppModule(app.keeper.staking, app.keeper.acct, app.keeper.supply),
-		distr.NewAppModule(app.keeper.distr, app.keeper.acct, app.keeper.supply, app.keeper.staking),
-		slashing.NewAppModule(app.keeper.slashing, app.keeper.acct, app.keeper.staking),
-		params.NewAppModule(), // NOTE: only used for simulation to generate randomized param change proposals
-		deployment.NewAppModuleSimulation(app.keeper.deployment, app.keeper.acct),
-		market.NewAppModuleSimulation(app.keeper.market, app.keeper.acct, app.keeper.deployment,
-			app.keeper.provider, app.keeper.bank),
-		provider.NewAppModuleSimulation(app.keeper.provider, app.keeper.acct),
```
boz, line 295: and here

boz:
sorry @akhilkumarpilli I forgot to merge this.

PR #703, opened 2020-06-24
devenv: add browseable option to _run/kind

boz:
Add an option to proxy localhost:80 to your app
to play nice with kubernetes ingresses.

### DIFF
--- a/_run/kube/Makefile
+++ b/_run/kube/Makefile
@@ -17,8 +17,8 @@ provider-run:
 provider-status:
 	curl -s "$(GATEWAY_ENDPOINT)/status" | jq
 
-.PHONY: provider-send-manifest
-provider-send-manifest:
+.PHONY: send-manifest
+send-manifest:
 	$(AKASHCTL) provider send-manifest "$(SDL_PATH)" \
 		--owner     "$(KEY_ADDRESS)" \
 		--dseq      "$(DSEQ)"        \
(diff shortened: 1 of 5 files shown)


boz:
@jackzampolin 

Screen Shot 2020-06-23 at 10 17 09 PM

_run/kube/Makefile
```diff
@@ -17,8 +17,8 @@ provider-run:
 provider-status:
 	curl -s "$(GATEWAY_ENDPOINT)/status" | jq
 
-.PHONY: provider-send-manifest
-provider-send-manifest:
+.PHONY: deployment-send-manifest
+deployment-send-manifest:
```
Ropes, line 21: Just to reduce overloading "deployment", perhaps `send-manifest-to-provider`, or keeping it short `send-manifest` imparts the action and subject of transfer.

_run/kube/README.md
```diff
@@ -66,13 +66,35 @@ make clean kind-cluster-clean init
...
 
 __t1__
 ```sh
 make kind-cluster-create
 ```
 
+#### Map local port 80 to to port 80 of your workload
+
+If anything else is listening on port 80 (any other web server), this method
+may fail.  If it does succeed, you will be able to browse your app from the browser.
```
Ropes, line 92: To help manage expectations: `... will fail.` If it can't bind to the port, it won't work.

_run/kube/Makefile
boz: I'll go with `send-manifest`; good call.

The command eventually needs to send the manifest to all providers of a deployment, hence `deployment-`.  There's a few of these that are hard calls tho.

PR #704, opened 2020-06-24
fix: adjust MaxGroupMemPrice

### LINKED ISSUE #640: provider bid pricing: adjust bid price config
The [current](https://github.com/ovrclk/akash/blob/21f9cb626a4cc2b490d9f9d245ed5742dd33ba57/provider/bidengine/pricing.go#L12) bid price calculation was formulated for our first testnet in 2018.  It chooses a random value between a min and max price that is determined by a configurable price per gigabyte range.

For some reason, [`calculatePrice()`](https://github.com/ovrclk/akash/blob/21f9cb626a4cc2b490d9f9d245ed5742dd33ba57/provider/bidengine/pricing.go#L12) is returning a very small range (`[0,1]`).  This might be config or it might be that the deployment price is set too low (we should use higher prices in our examples).

troian:
if total amount of memory in group less then 1G
calculatePriceRange will always return range 0..1.

fixes #640 ?

### DIFF
--- a/validation/config.go
+++ b/validation/config.go
@@ -26,7 +26,7 @@ type ValConfig struct {
 	MaxGroupStorage int64 `env:"AKASH_MAX_GROUP_STORAGE"   envDefault:"5368709120"` // 5Gi
 
 	MinGroupMemPrice int64 `env:"AKASH_MEM_PRICE_MIN" envDefault:"50"`
-	MaxGroupMemPrice int64 `env:"AKASH_MEM_PRICE_MAX" envDefault:"150"`
+	MaxGroupMemPrice int64 `env:"AKASH_MEM_PRICE_MAX" envDefault:"1048576"`
 }
 
 var defaultConfig = ValConfig{}


validation/config.go
```diff
@@ -35,6 +39,10 @@ func init() {
 	if err := env.Parse(&defaultConfig); err != nil {
 		panic(err)
 	}
+
+	if defaultConfig.MaxGroupMemPrice < minDefaultMaxGroupMemPrice {
+		panic("invalid max group memory price")
+	}
```
boz, line 45: let's not do this.  i agree with the sentiment but we should handle it more gracefully.  probably need to refactor the way this validation works - it was added in a rush to make a (2018) testnet more user-friendly.

just skip it for now

troian: `fmt.Println` instead of panic?

boz: nah.

boz: everything still works fine with a lower value, right?  it's just that the prices are lower than we'd like.

troian: kk. makes sense

PR #707, opened 2020-06-24
fix: check kind installed

troian:


_run/kube/kind.mk
```diff
@@ -12,6 +13,10 @@ PROVIDER_ENDPOINT ?= http://$(PROVIDER_HOST)
 
 INGRESS_CONFIG_PATH ?= https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/kind/deploy.yaml
 
+ifeq (, $(shell which kind))
+$(error "No kind in $(PATH), consider install?")
```
boz, line 17: "run 'make devdeps-install' in the top-level directory to install"

troian: maybe put `devdeps-install` into some mk and include in any makefile that needs it?

boz: it only really needs to be run once.  i don't care all that much.

boz:
The instructions now don't point to the version pinned by `tools.go` and `go.sum`.

PR #708, opened 2020-06-24
_run/kube documentation proposal

Ropes:
Updates to the documentation to hopefully improve readability and
explicit steps needed to achieve operation. Goal was to improve
readbility as raw markdown and rendered document.

* Trimmed up some duplicated text and typos.
* Steps which execute commands vs status check commands have titles
which inform `run step: {N}` and `status`.
    * Status commands are indented.
* Header sizes are increased for sections.
* First `kube-cluster-create` command options are formatted into a
table.


_run/kube/README.md
```diff
@@ -6,7 +6,7 @@ The _Kube_ dev environment builds:
 * An Akash Provider Services Daemon (PSD) for bidding and running workloads.
 * A Kubernetes cluster for the PSD to run workloads on.
 
-The [instructions](#runbook) below will illustrate how to:
+The [instructions](#runbook) below will illustrate how to run a network with a single, local node and execute workloads in Minikube:
```
boz, line 9: ~Minikube~ `[kind](https://kind.sigs.k8s.io/)`

_run/kube/README.md
```diff
@@ -49,22 +41,23 @@ this example.
...
 Each command is marked __t1__-__t3__ to indicate a suggested terminal number.
 
 If at any time you'd like to start over with a fresh chain, simply run:
 
-__t1__
+__t1 run__
 ```sh
-make clean kind-cluster-clean init
+make clean kind-cluster-clean 
+make init
```
boz, line 57: curious: is this for readability, consistency w/ later steps, or something else?

_run/kube/README.md
```diff
@@ -73,97 +66,85 @@ environments.  Two options are offered below - the first (random port) is less e
...
 make node-run
 ```
 
 You can check the status of the network with:
 
-__t1__
+__t1 status__
 ```sh
-make node-status
+    make node-status
```
boz, line 99: accidental tab here & in many pre-formatted sections below?

boz: I see now that this was intentional.

I *think* this is so you can differentiate status commands from non-status commands, I'm just not sure if that is helpful at all.  Have you seen this elsewhere?

Ropes: I intentionally added the tab as a way to separate status commands from commands which execute a change. 

It's just an additional idea to split the two types of commands. The two are also delineated by the title `__t* run step...__`, so it's a bit redundant.

_run/kube/README.md
Ropes: If there is an error in the `kind-cluster-clean` step; is, it breaks, and then won't run the `init` component. So I split it into two lines so the init will run afterwards.

_run/kube/README.md
boz: I see.  cool idea.  I was open to it but after looking through it I think it just makes things more confusing.

Every `##` block has one "change" command followed by query commands.  Maybe we just prefix each `##` header with a step number or something.

Ropes: How does `### __t1 Step: 2__` for commands which execute changes and must be run look? eg:

### __t1 Step: 2__
```sh
echo "run me"
```

Issue #604, opened 2020-05-20
db performance: indexes


boz:
We need indexes for fast look-ups of often-used queries.

Some that come to mind:

* [active orders](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/x/deployment/handler/endblock.go#L16-L19)
* [active leases](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/x/market/handler/endblock.go#L28-L30)
* [unmatched orders](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/x/market/handler/endblock.go#L62-L64)
* [leases for provider](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/client/extra.go#L9-L11) indexing bids might be better than leases.

Edit:

One option is to use prefixes to represent various states.  See [kava](https://github.com/Kava-Labs/kava) for an example

* [key](https://github.com/Kava-Labs/kava/blob/467e6f7d8b48cea93f6d6d6f1b26675e4332aed9/x/auction/types/keys.go#L30)
* [keeper](https://github.com/Kava-Labs/kava/blob/467e6f7d8b48cea93f6d6d6f1b26675e4332aed9/x/auction/keeper/keeper.go#L91-L103)

Ropes:
Re:discovered the need for order index.

Similar to our previous indexing of data which needs quick iteration in EndBlock calls; `market` module needs an index on Orders in the Open state.

The Market module's endblock calls [WithOrders](https://github.com/ovrclk/akash/blob/98d799978399b9d2591145d0d7495d2580cc51f1/x/market/handler/endblock.go#L105). Which will iterate over all Orders, including historical ones which have long since been closed. While it has a quick exit for closed orders, it will waste a lot of time iterating in a mature blockchain.

Ropes:
Started on the Orders Open Index, for next week:
    TODO:
    * Add/Created Open Orders to the new Index
    * Remove them when state changes
    * Implement testing

boz:
Where are we at with this @Ropes ?

PR #714, opened 2020-06-26
Deployment EndBlock testing(prep for Indexing)

Ropes:
* WithDeploymentsActive() added to do basic filtering of all
deployments which are in active state.
    * Used in Deployment's EndBlock function.
    * Tests on EndBlock attempting to assert expected behavior.
    * TODO: Assert EndBlock error cases.
* testutil.CoinDenom added to enable passing created coins into testutil
methods which accept sdk.Coin's with consistent denoms.
    * Updated all tests which fit this case.

related: #604

### DIFF
--- a/testutil/base.go
+++ b/testutil/base.go
@@ -11,6 +11,10 @@ import (
 	dtypes "github.com/ovrclk/akash/x/deployment/types"
 )
 
+// CoinDenom provides ability to create coins in test functions and
+// pass them into testutil functionality.
+const CoinDenom = "akash"
+
 // Name generates a random name with the given prefix
 func Name(_ testing.TB, prefix string) string {
 	return fmt.Sprintf("%s-%v", prefix, rand.Uint64())
@@ -40,12 +44,15 @@ func Attributes(t testing.TB) []sdk.Attribute {
 
 }
 
+// Resources produces an attribute list for populating a Group's
+// 'Resources' fields.
 func Resources(t testing.TB) []dtypes.Resource {
 	t.Helper()
 	count := rand.Intn(10) + 1
 
 	vals := make([]dtypes.Resource, 0, count)
 	for i := 0; i < count; i++ {
+		coin := sdk.NewCoin(CoinDenom, sdk.NewInt(rand.Int63n(9999)))
 		res := dtypes.Resource{
 			Unit: types.Unit{
 				CPU:     100,
@@ -53,7 +60,7 @@ func Resources(t testing.TB) []dtypes.Resource {
 				Storage: 10,
 			},
 			Count: 1,
-			Price: sdk.NewCoin("thepricedcoin", sdk.NewInt(int64(rand.Uint16()))),
+			Price: coin,
 		}
 		vals = append(vals, res)
 	}
--- a/x/deployment/handler/endblock_test.go
+++ b/x/deployment/handler/endblock_test.go
@@ -1,0 +1,73 @@
+package handler_test
+
+import (
+	"testing"
+
+	"github.com/ovrclk/akash/testutil"
+	"github.com/ovrclk/akash/x/deployment/handler"
+	"github.com/ovrclk/akash/x/deployment/types"
+	mtypes "github.com/ovrclk/akash/x/market/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestEndBlock(t *testing.T) {
+	suite := setupTestSuite(t)
+	d0 := testutil.Deployment(suite.t)
+	g0 := testutil.DeploymentGroup(suite.t, d0.DeploymentID, uint32(5))
+	g1 := testutil.DeploymentGroup(suite.t, d0.DeploymentID, uint32(100))
+	g1.State = types.GroupClosed
+
+	d1 := testutil.Deployment(suite.t)
+	d1.State = types.DeploymentClosed
+	g2 := testutil.DeploymentGroup(suite.t, d1.DeploymentID, uint32(8))
+
+	// create deployments in storage
+	df := func(s *testSuite, d types.Deployment, groups ...types.Group) {
+		grps := make([]types.GroupSpec, 0, len(groups))
+		for _, g := range groups {
+			grps = append(grps, g.GroupSpec)
+		}
+		m := types.MsgCreateDeployment{
+			ID:     d.ID(),
+			Groups: grps,
+		}
+		_, err := s.handler(s.ctx, m)
+		assert.NoError(s.t, err)
+
+		if d.State == types.DeploymentClosed {
+			m := types.MsgCloseDeployment{
+				ID: d.ID(),
+			}
+			_, err := s.handler(s.ctx, m)
+			assert.NoError(s.t, err)
+		}
+	}
+	df(suite, d0, g0, g1)
+	df(suite, d1, g2)
+
+	// Execute EndBlock method
+	handler.OnEndBlock(suite.ctx, suite.dkeeper, suite.mkeeper)
+
+	// Check results of EndBlock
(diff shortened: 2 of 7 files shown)


testutil/base.go
```diff
@@ -40,20 +44,27 @@ func Attributes(t testing.TB) []sdk.Attribute {
...
 	t.Helper()
 	count := rand.Intn(10) + 1
 
 	vals := make([]dtypes.Resource, 0, count)
 	for i := 0; i < count; i++ {
+		coin := sdk.NewCoin(CoinDenom, sdk.NewInt(int64(rand.Uint16())))
+		if coin.Amount.Int64() > int64(10000) {
+			limited := coin.Amount.Int64() % int64(1000)
+			coin = sdk.NewCoin(CoinDenom, sdk.NewInt(limited))
+		}
```
boz, line 59: Is this to address the fluctuating coverage?

I think that the action item on those is to make tests that target the sometimes-covered lines.

Reading through this, it's unclear what the point of capping it at 1,000 if it's over 10,000 is

x/deployment/handler/endblock_test.go
```diff
@@ -0,0 +1,78 @@
...
+		if d.State == types.DeploymentClosed {
+			m := types.MsgCloseDeployment{
+				ID: d.ID(),
+			}
+			_, err := suite.handler(suite.ctx, m)
+			assert.NoError(suite.t, err)
+		}
+	}
+	df(suite, d0, g0, g1)
+	df(suite, d1, g2)
```
boz, line 46: love it

x/deployment/handler/endblock_test.go
```diff
@@ -0,0 +1,78 @@
...
+	g1 := testutil.DeploymentGroup(suite.t, d0.DeploymentID, uint32(100))
+	g1.State = types.GroupClosed
+
+	d1 := testutil.Deployment(suite.t)
+	d1.State = types.DeploymentClosed
+	g2 := testutil.DeploymentGroup(suite.t, d1.DeploymentID, uint32(8))
+
+	// create deployments in storage
+	df := func(s *testSuite, d types.Deployment, groups ...types.Group) {
+		grps := make([]types.GroupSpec, 0)
```
boz, line 26: wonder if the linter will :rotating_light: here

```go
grps := make([]types.GroupSpec, 0,len(groups))
```

x/deployment/handler/endblock_test.go
```diff
@@ -0,0 +1,78 @@
...
+	df(suite, d1, g2)
+
+	// Execute EndBlock method
+	handler.OnEndBlock(suite.ctx, suite.dkeeper, suite.mkeeper)
+
+	// Check results of EndBlock
+	gx := suite.dkeeper.GetGroups(suite.ctx, d0.ID())
+	if len(gx) == 0 {
+		t.Error("no groups returned from keeper")
+	}
```
boz, line 55: `assert.NotEmpty(suite.t,gx)`

x/deployment/handler/endblock_test.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, g := range gx {
+		orderCreated := false
+		suite.mkeeper.WithOrdersForGroup(suite.ctx, g.ID(), func(o mtypes.Order) bool {
+			suite.t.Logf("Order for group: %#v found", o.GroupID())
+			orderCreated = true
+			return true
+		})
+		if !orderCreated {
+			suite.t.Error("order was not created for a group")
+		}
```
boz, line 65: `assert.True(suite.t,orderCreated)`

x/deployment/handler/endblock_test.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	// d1 is a closed group, assert no orders created
+	gy := suite.dkeeper.GetGroups(suite.ctx, d1.ID())
```
boz, line 69: `assert.Len(suite.t,gy,1)`

x/deployment/keeper/keeper.go
```diff
@@ -147,6 +147,24 @@ func (k Keeper) WithDeployments(ctx sdk.Context, fn func(types.Deployment) bool)
 	}
 }
 
+// WithDeploymentsActive filters to only those with State: Active
+func (k Keeper) WithDeploymentsActive(ctx sdk.Context, fn func(types.Deployment) bool) {
+	store := ctx.KVStore(k.skey)
+	// Question: is it possible to search by a single attribute across all Deployments
+	// with only a prefix iterator?
```
boz, line 154: Need to make a new index for it `(index-id,state,deployment-id)`.

If we only need to index active deployments then only active deployments need to go into that index.

x/deployment/keeper/keeper.go
```diff
@@ -147,6 +147,24 @@ func (k Keeper) WithDeployments(ctx sdk.Context, fn func(types.Deployment) bool)
 	}
 }
 
+// WithDeploymentsActive filters to only those with State: Active
+func (k Keeper) WithDeploymentsActive(ctx sdk.Context, fn func(types.Deployment) bool) {
```
boz, line 151: `WithActiveDeployments()` rolls off the tongue better.  Or make the state a parameter `WithDeploymentsInState(ctx,state,fn)`

x/market/handler/handler_test.go
```diff
@@ -90,7 +90,7 @@ func TestCreateBidValid(t *testing.T) {
 	msg := types.MsgCreateBid{
 		Order:    order.ID(),
 		Provider: provider,
-		Price:    sdk.NewCoin("thepricedcoin", sdk.NewInt(1)),
+		Price:    sdk.NewCoin(testutil.CoinDenom, sdk.NewInt(1)),
```
boz, line 93: :+1:

x/deployment/keeper/key.go
```diff
@@ -19,6 +19,12 @@ func deploymentKey(id types.DeploymentID) []byte {
 	return buf.Bytes()
 }
 
+func deploymentStateKey(d types.Deployment) []byte {
```
boz: probably leave this out if it's unused.

testutil/base.go
Ropes: There's an error which pops up if the value is over 10k by some tests, I'm not exactly sure the reasoning.

So if it's over 10k, it gets remaindered by 1000 to make sure it's below 10k.

boz: why not just use `rand.Intn(10000)` initially?

x/deployment/keeper/keeper.go
Ropes: Yeah I talked with Jack a little about that, I started down that progress on another branch since it'll be more substantial. I'll remove the question since answered.

testutil/base.go
Ropes: That seems like the obvious solution :)  Done.

testutil/base.go
```diff
@@ -40,20 +44,23 @@ func Attributes(t testing.TB) []sdk.Attribute {
...
 
+// Resources produces an attribute list for populating a Group's
+// 'Resources' fields.
 func Resources(t testing.TB) []dtypes.Resource {
 	t.Helper()
 	count := rand.Intn(10) + 1
 
 	vals := make([]dtypes.Resource, 0, count)
 	for i := 0; i < count; i++ {
+		coin := sdk.NewCoin(CoinDenom, sdk.NewInt(rand.Int63n(9999)))
```
boz, line 55: It doesn't really matter much but `Intn(x)` returns something in the range of `[0,x)` (`[0,x-1]`); hence the `Intn(10) + 1` above - I wanted the range `[1,10]`.

PR #721, opened 2020-06-29
feat: implement service log stream

troian:


provider/gateway/client.go
```diff
@@ -144,5 +149,88 @@ func makeURI(host string, path string) (string, error) {
...
+				if e = conn.WriteMessage(websocket.PongMessage, []byte{}); e != nil {
+					return
+				}
+			case websocket.TextMessage:
+				var logLine cluster.ServiceLogMessage
+				if e = json.Unmarshal(msg, &logLine); e != nil {
+					return
+				}
+
+				logs.Stream <- logLine
```
boz, line 229: make ~`logs.Stream`~ `streamch` a buffered channel and do a non-blocking send here to not seize up the websocket?

```go
 // or a param or whatever
const streamchBufsiz = 64
// ...
streamch := make(chan cluster.ServiceLogMessage,streamchBufsiz)
// ...

select {
  case streamch <- logLine:
  default:
    // error out or something.  optional for now.
}
```

provider/gateway/client.go
```diff
@@ -144,5 +149,88 @@ func makeURI(host string, path string) (string, error) {
...
+			default:
+			}
+		}
```
boz, line 232: I think this can be simplified a lot.  Let's forget the pings and read deadlines.

```go
loop:
for {
  var logLine cluster.ServiceLogMessage
  if err := conn.ReadJSON(&logLine); err != nil {
    break
  }
  select {
    case streamch <- logLine:
    default:
      break loop
  }
}
conn.Close() // this isn't done right now
```

provider/gateway/client.go
```diff
@@ -144,5 +149,88 @@ func makeURI(host string, path string) (string, error) {
...
+		if i == 0 {
+			uri += "?"
+		} else {
+			uri += "&"
+		}
+
+		uri += fmt.Sprintf("%s=%s", url.QueryEscape(kv.k), url.QueryEscape(kv.v))
+	}
+
+	conn, _, err := websocket.DefaultDialer.DialContext(ctx, uri, nil)
```
boz, line 197: :thinking: will cancelling this context shut down the connection, or is it only for dialing?

provider/gateway/middleware.go
```diff
@@ -73,3 +85,48 @@ func parseLeaseID(req *http.Request) (mtypes.LeaseID, error) {
...
+
+			val := vars.Get("follow")
+			if val == "" {
+				follow = true
+			} else {
+				follow, err = strconv.ParseBool(val)
+				if err != nil {
+					return
+				}
+			}
```
boz, line 114: hmm let's be strict at this point; the client should send well-formed requests.

```go
follow := false
if val := vars.Get("follow"); val != "" {
  if follow, err = strconv.ParseBool(val); err != nil {
    return
  }
}
```

provider/gateway/middleware.go
```diff
@@ -73,3 +85,48 @@ func parseLeaseID(req *http.Request) (mtypes.LeaseID, error) {
...
+
+			if val = vars.Get("tail"); val != "" {
+				vl := new(int64)
+				*vl, err = strconv.ParseInt(val, 10, 32)
+				if err != nil {
+					return
+				}
+
+				tailLines = vl
```
boz, line 123: `vl` unneeded again I think.

provider/gateway/router.go
```diff
@@ -116,9 +150,152 @@ func leaseServiceStatusHandler(log log.Logger, cclient cluster.ReadClient) http.
...
 
 func leaseServiceLogsHandler(log log.Logger, cclient cluster.ReadClient) http.HandlerFunc {
-	return func(w http.ResponseWriter, req *http.Request) {
-		http.Error(w, "unimplemented", http.StatusNotImplemented)
+	return func(w http.ResponseWriter, r *http.Request) {
+		service := mux.Vars(r)["serviceName"]
+		if service == "" {
+			http.Error(w, "invalid request", http.StatusBadRequest)
+			return
+		}
```
boz, line 158: let's handle this in the same way the other params are handled.  for better or for worse.

provider/gateway/router.go
```diff
@@ -116,9 +150,152 @@ func leaseServiceStatusHandler(log log.Logger, cclient cluster.ReadClient) http.
...
+
+		go wsLogWriter(ws, wsLogsConfig{
+			lid:       requestLeaseID(r),
+			service:   service,
+			follow:    requestLogFollow(r),
+			tailLines: requestLogTailLines(r),
+			log:       log,
+			client:    cclient,
+		})
```
boz, line 180: I think that this handler should block until everything is done.  does `wsLogWriter()` need to be in a `go`?  I think maybe not.

provider/gateway/router.go
```diff
@@ -116,9 +150,152 @@ func leaseServiceStatusHandler(log log.Logger, cclient cluster.ReadClient) http.
...
+	}
+}
+
+func wsLogWriter(ws *websocket.Conn, cfg wsLogsConfig) {
+	pingTicker := time.NewTicker(pingPeriod)
+
+	ctx, cancel := context.WithCancel(context.Background())
```
boz, line 187: we got a context with the request, should use that instead of `context.Background()`.  Maybe it's in the `ws`, but you might have to pass it in.

provider/gateway/router.go
```diff
@@ -116,9 +150,152 @@ func leaseServiceStatusHandler(log log.Logger, cclient cluster.ReadClient) http.
...
+	// drain logs channel in separate goroutine to unblock seeders waiting for write space
+	scannersStarted.Add(1)
+	go func() {
+		defer scannersStarted.Done()
+		for range logCh {
+		}
+	}()
+
+	close(logCh)
+	scannersStarted.Wait()
```
boz, line 298: Again let's forgo the ping stuff for now.

1. I don't think we need `scannersStarted`.
1. `logCh` doesn't need to be closed.
1. drain loop seems off; the writers are already in their own threads.
1. this should probably be in the same thread as the initial http handler.

```go
for _, lg := range logs {
  go func(name string, scan *bufio.Scanner) {
    defer scanners.Done()
    defer scan.Close()
    for scan.Scan() && ctx.Err() == nil {
       // ...
    }
  }()
}

donech := make(chan struct{})
go func() {
  scanners.Wait()
  close(donech)
}()
// ...

for alive {
  select {
    case line := <- logCh:
      ws.WriteJSON(line)
      // if err alive = false
    case <-donech:
      alive = false
  }
}

// tell all the log readers to stop.
cancel()

ws.WriteMessage(CloseMessage,...)

drain:
for {
  select {
    case <- donech:
      break drain
    case <- logCh:
  }
}
```

provider/cmd/serviceLogs.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	switch uri.Scheme {
+	case "ws", "http", "":
+		uri.Scheme = "ws"
+	case "wss", "https":
+		uri.Scheme = "wss"
+	default:
+		return errors.Errorf("invalid uri scheme \"%s\"", uri.Scheme)
+	}
```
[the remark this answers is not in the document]
troian: good point

provider/cmd/serviceLogs.go
```diff
@@ -0,0 +1,123 @@
...
+
+	var tailLines *int64
+
+	if cmd.Flags().Changed("tail") {
+		vl := new(int64)
+		if *vl, err = cmd.Flags().GetInt64("tail"); err != nil {
+			return err
+		}
+
+		tailLines = vl
```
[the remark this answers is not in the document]
boz: I take it back.  I was thinking of [`kingpin.Int()`](https://godoc.org/gopkg.in/alecthomas/kingpin.v2#FlagClause.Int) for some reason last night :shrug:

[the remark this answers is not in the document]
troian: I refactored it `kubectl` way. it is reasonable to set default flags value to `-1` and use it as value to read from the beginning

provider/gateway/router.go
boz: you were right to close the `*bufio.Scanner`s in the main/outer thread - `cancel()` won't shut down the logs coming from kube, you have to `Close()` the bufio.

boz: oh wait that's why i added the `ctx.Err() == nil` condition.  never mind.

_run/kube/README.md
```diff
@@ -208,6 +208,20 @@ __t1 status__
...
+__t1 service status__
+```sh
+make provider-service-status
+```
+
+Fetch logs from deployed service (all pods)
+
+__t1 service logs__
+```sh
+make provider-service-logs
```
boz, line 222: :+1:

provider/cluster/kube/k8s_integration_test.go
```diff
@@ -89,6 +92,39 @@ func TestNewClient(t *testing.T) {
 	assert.NoError(t, err)
 	assert.NotEqual(t, maxtries, tries)
 
+	logs, err := client.ServiceLogs(ctx, lid, svcname, true, nil)
+	require.NoError(t, err)
+	require.Equal(t, int(sstat.AvailableReplicas), len(logs))
+
+	log := make(chan string)
+	defer close(log)
```
boz, line 100: not really necessary, but it will cause a panic if `log` is written to after it's closed, which is nice for the tests

provider/cluster/kube/k8s_integration_test.go
```diff
@@ -89,6 +92,39 @@ func TestNewClient(t *testing.T) {
...
+
+	var wgFinished sync.WaitGroup
+	wgFinished.Add(1)
+	go func(scan *bufio.Scanner) {
+		defer wgFinished.Done()
+
+		for scan.Scan() {
+			lg := scan.Text()
+			fmt.Println(lg)
```
boz, line 109: :nail_care: remove printf debugging

provider/cluster/kube/k8s_integration_test.go
```diff
@@ -89,6 +92,39 @@ func TestNewClient(t *testing.T) {
...
+			log <- lg
+			break
+		}
+	}(logs[0].Scanner)
+
+	select {
+	case line := <-log:
+		assert.NotEmpty(t, line)
+	case <-time.After(10 * time.Second):
+		_ = logs[0].Stream.Close()
```
boz, line 119: it's closed below, not necessary here.

provider/cluster/kube/k8s_integration_test.go
```diff
@@ -89,6 +92,39 @@ func TestNewClient(t *testing.T) {
...
+		assert.Fail(t, "timed out waiting for logs")
+	}
+
+	for _, lg := range logs {
+		assert.NoError(t, lg.Stream.Close(), lg.Name)
+	}
+
+	// logs[0].Stream.
```
boz, line 127: The [same](https://github.com/ovrclk/akash/pull/721#discussion_r448777841) race condition is still here - `log <- lg` can get stuck.

provider/cluster/types.go
```diff
@@ -29,8 +29,13 @@ type ServiceStatus struct {
...
 
+type ServiceLogMessage struct {
+	Name    string `json:"name"`
+	Message string `json:"message"`
+}
+
 type ServiceLogs struct {
-	Name string `json:"name"`
+	Stream chan ServiceLogMessage
 }
```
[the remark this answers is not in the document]
boz: :point_up:

provider/cmd/serviceLogs.go
```diff
@@ -0,0 +1,118 @@
...
+
+	result, err := gclient.ServiceLogs(context.Background(), provider.HostURI, lid, svcName, follow, tailLines)
+	if err != nil {
+		return err
+	}
+
+	printFn := func(msg cluster.ServiceLogMessage) error {
+		fmt.Printf("[%s] %s\n", msg.Name, msg.Message)
+		return nil
+	}
```
boz, line 99: :100:

provider/gateway/client.go
```diff
@@ -110,6 +115,8 @@ func (c *client) ServiceStatus(ctx context.Context, host string, id mtypes.Lease
 		return nil, err
 	}
 
+	fmt.Printf("status %s\n", uri)
```
boz, line 118: :nail_care: printf debug

provider/gateway/client.go
```diff
@@ -144,5 +149,88 @@ func makeURI(host string, path string) (string, error) {
...
+
+	for i, kv := range query {
+		if i == 0 {
+			uri += "?"
+		} else {
+			uri += "&"
+		}
+
+		uri += fmt.Sprintf("%s=%s", url.QueryEscape(kv.k), url.QueryEscape(kv.v))
+	}
```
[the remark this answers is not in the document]
boz: use [`url.Values`](https://golang.org/pkg/net/url/#Values) instead of rolling all of this by hand.

provider/gateway/client.go
```diff
@@ -144,5 +149,88 @@ func makeURI(host string, path string) (string, error) {
...
+	conn, _, err := websocket.DefaultDialer.DialContext(ctx, uri, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// todo check status?
+
+	logs := &cluster.ServiceLogs{
+		Stream: make(chan cluster.ServiceLogMessage),
+	}
```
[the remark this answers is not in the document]
boz: :point_up:

provider/gateway/client.go
```diff
@@ -144,5 +151,98 @@ func makeURI(host string, path string) (string, error) {
...
+
+	go func(conn *websocket.Conn) {
+		defer close(logs.Stream)
```
boz, line 221: Should it close the connection?
```go
defer func(){
   _ = conn.Close()
}()
```

provider/gateway/middleware.go
```diff
@@ -55,6 +73,24 @@ func requireLeaseID(log log.Logger) mux.MiddlewareFunc {
...
+			if svc == "" {
+				http.Error(w, "empty service name", http.StatusBadRequest)
+				return
+			}
+
+			fmt.Printf("param service %s\n", svc)
+			context.Set(req, serviceContextKey, svc)
+			next.ServeHTTP(w, req)
+		})
+	}
```
boz, line 91: :pray:

provider/gateway/middleware.go
boz: :point_up:

provider/gateway/middleware.go
boz: I thought that with the default of `-1` we don't have to deal with all of this.

provider/gateway/router.go
```diff
@@ -39,13 +70,19 @@ func newRouter(log log.Logger, pclient provider.Client) *mux.Router {
 		leaseStatusHandler(log, pclient.Cluster())).
 		Methods("GET")
 
+	srouter := lrouter.PathPrefix("/service/{serviceName}").Subrouter()
+	srouter.Use(requireService())
+
 	// GET /lease/<lease-id>/service/<service-name>/status
-	lrouter.HandleFunc("/service/{serviceName}/status",
+	srouter.HandleFunc("/status",
```
boz, line 77: :pray:

provider/gateway/router.go
```diff
@@ -101,12 +138,7 @@ func leaseStatusHandler(log log.Logger, cclient cluster.ReadClient) http.Handler
 
 func leaseServiceStatusHandler(log log.Logger, cclient cluster.ReadClient) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
-		service := mux.Vars(req)["serviceName"]
-		if service == "" {
-			http.Error(w, "invalid request", http.StatusBadRequest)
-			return
-		}
-		status, err := cclient.ServiceStatus(req.Context(), requestLeaseID(req), service)
```
boz, line 109: :pray:

provider/gateway/router.go
```diff
@@ -116,9 +148,121 @@ func leaseServiceStatusHandler(log log.Logger, cclient cluster.ReadClient) http.
...
+	}
+
+	// drain logs channel in separate goroutine to unblock seeders waiting for write space
+	go func() {
+		for range logCh {
+		}
+	}()
+
+	close(logCh)
+	scanners.Wait()
```
boz, line 265: There are still problems here.

1. There's no guarantee that all of the writes to `logCh` are done before it's closed.
1. There's no way to know that the scanners have finished in the `for alive {` loop.

Suggestion from [earlier](https://github.com/ovrclk/akash/pull/721#discussion_r448813951):

* worker: `scan.Close()` in the before `scanner.Done()` so you don't have to deal with it elsewhere.
* worker: guard the loop with `scan.Scan() && ctx.Err() == nil` so that it stops when `cancel()` is called.
* mainloop: kick off a go routine that does `scanner.Wait()` then can signal that to `select` by closing a channel.
* mainloop: call `cancel()` to begin shutdown after `for alive {` loop
* mainloop: properly drain `logCh` by selecting in a loop along with the channel that signals that the workers are done.  this should not be in a separate thread.

testutil/manifest_app.go
```diff
@@ -34,7 +34,7 @@ func (mg manifestGeneratorApp) Service(t testing.TB) manifest.Service {
 	t.Helper()
 	return manifest.Service{
 		Name:  "demo",
-		Image: "quay.io/ovrclk/demo-app",
+		Image: "chentex/random-logger:latest", // todo does image change introduces side-effects?
```
boz, line 37: nah.  it's fine.

provider/gateway/router.go
troian: right `scanners.Wait` should be this way to prevent write into closed channel
```go
scanners.Wait()
close(logCh)
```

- scanner does not implement `Close()` that why have to close it from the reader. 
- `Scan()` can be blocked by underlying `Read()` call. to properly abort the call `Close()` should be called from another goroutine

```
	for _, lg := range logs {
		go func(name string, scan *bufio.Scanner) {
			defer scanners.Done()

			for scan.Scan() {
				logCh <- cluster.ServiceLogMessage{
					Name:    name,
					Message: scan.Text(),
				}
			}
		}(lg.Name, lg.Scanner)
	}

	alive := true

	donech := make(chan struct{})

	go func() {
		scanners.Wait()
		close(donech)
	}()

	for alive {
		select {
		case line := <-logCh:
			if err = ws.WriteJSON(line); err != nil {
				alive = false
			}
		case <-pingTicker.C:
			if err = ws.WriteMessage(websocket.PingMessage, []byte{}); err != nil {
				alive = false
			}
		case <-donech:
			alive = false
		}
	}

	cancel()

	for i := range logs {
		_ = logs[i].Stream.Close()
	}

	// drain logs channel in separate goroutine to unblock seeders waiting for write space
	go func() {
		for range logCh {
		}
	}()

	close(logCh)
```

troian: no, example above is wrong. it will deadlock

troian: looks like original approach was right. `for alive {` does not need to know about streams state. any first failure from `ws.WriteJSON(line)` or writing ping message, or timeout on waiting for `pong` break the `alive` loop.
then all streams are forced to close
```go
	for i := range logs {
		_ = logs[i].Stream.Close()
	}
```

by draining `logch` we ensure all scanners waiting on space in `logch` write and exit. when `logch` is closed all scanners finished

boz: > does not need to know about streams state. any first failure from ws.WriteJSON(line) ...

I'm talking about the logs coming from kube - `scan.Scan()` having returned false for all of them.

> by draining logch we ensure all scanners waiting on space in logch write and exit. when logch is closed all scanners finished

When do you know that you should close `logCh`?  Why do you need to close `logCh`?

provider/cluster/kube/k8s_integration_test.go
```diff
@@ -89,6 +92,41 @@ func TestNewClient(t *testing.T) {
...
+	case <-log:
+		if !timer.Stop() {
+			<-timer.C
+		}
+	case <-timer.C:
+		_ = logs[0].Stream.Close()
+		t.Errorf("timed out waiting for logs from pod [%s]", logs[0].Name)
+	}
+
+	wgFinished.Wait()
```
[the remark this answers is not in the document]
troian: or `log := make(chan string, 1)` 🤣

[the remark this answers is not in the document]
boz: 100% right thing to do

provider/gateway/router.go
troian: sorted in slack

provider/gateway/router.go
```diff
@@ -116,8 +148,130 @@ func leaseServiceStatusHandler(log log.Logger, cclient cluster.ReadClient) http.
...
+	}()
+
+	alive := true
+
+	for alive {
+		select {
+		case line := <-logch:
+			if err = ws.WriteJSON(line); err != nil {
+				alive = false
+			}
```
boz, line 253:
```go
case <- donech:
  alive = false
```

troian: thanks!

provider/cluster/kube/k8s_integration_test.go
```diff
@@ -89,6 +91,31 @@ func TestNewClient(t *testing.T) {
 	assert.NoError(t, err)
 	assert.NotEqual(t, maxtries, tries)
 
+	logs, err := client.ServiceLogs(ctx, lid, svcname, true, nil)
+	require.NoError(t, err)
+	require.Equal(t, int(sstat.AvailableReplicas), len(logs))
+
+	log := make(chan string)
+	defer close(log)
```
boz, line 99: `log := make(chan string,1)` like you mentioned, and don't `close(log)`.

provider/gateway/client.go
```diff
@@ -144,5 +158,79 @@ func makeURI(host string, path string) (string, error) {
...
+	default:
+		return nil, errors.Errorf("invalid uri scheme \"%s\"", endpoint.Scheme)
+	}
+
+	query := url.Values{}
+
+	query.Set("follow", strconv.FormatBool(follow))
+	query.Set("tail", strconv.FormatInt(tailLines, 10))
+
+	endpoint.RawQuery = query.Encode()
```
boz, line 191: :100:

provider/gateway/client.go
```diff
@@ -144,5 +158,79 @@ func makeURI(host string, path string) (string, error) {
...
+				if e = conn.WriteMessage(websocket.PongMessage, []byte{}); e != nil {
+					return
+				}
+			case websocket.TextMessage:
+				var logLine ServiceLogMessage
+				if e = json.Unmarshal(msg, &logLine); e != nil {
+					return
+				}
+
+				streamch <- logLine
```
boz, line 229: :thinking: I wonder if we should handle the `websocket.CloseMessage`?  we can add it later if so.

provider/gateway/router.go
```diff
@@ -116,8 +148,132 @@ func leaseServiceStatusHandler(log log.Logger, cclient cluster.ReadClient) http.
...
+		}
+	}
+
+	cancel()
+
+	for i := range logs {
+		_ = logs[i].Stream.Close()
+	}
+
+	// drain logs channel in separate goroutine to unblock seeders waiting for write space
```
boz, line 269: :nail_care: ~separate goroutine~

provider/gateway/client.go
troian: good point

provider/gateway/router.go
troian: oh yeah!

PR #735, opened 2020-07-06
fix: reject provider messages with duplicated attributes

### LINKED ISSUE #728: provider attributes: don't allow duplicate keys
Right now, providers are allowed to have duplicate attributes.  This doesn't make sense because if either value doesn't match the required attribute then [`MatchAttributes()`](https://github.com/ovrclk/akash/blob/3310cc668be406a19029ce458831a53f16345198/x/deployment/types/types.go#L93-L104) will fail.

troian:
fixes #728


troian:
just after first push realized test weren't committed

PR #736, opened 2020-07-07
feat:604 Active Lease indexing

Ropes:
Equivalent to the [Open Group indexing](https://github.com/ovrclk/akash/pull/729) queries.

Creates a second index in the `market` module which does not store data,
only tracks active Leases. Once a lease is moved out of active state the
key is removed from the index.

Testing improvement is not significant, mostly depending on existing
tests. More will come down with the Open Bids index.

refs: #604

### DIFF
--- a/x/market/keeper/keeper.go
+++ b/x/market/keeper/keeper.go
@@ -95,7 +95,8 @@ func (k Keeper) CreateBid(ctx sdk.Context, oid types.OrderID, provider sdk.AccAd
 	return bid, nil
 }
 
-// CreateLease creates lease for bid with given bidID
+// CreateLease creates lease for bid with given bidID.
+// Should only be called by the EndBlock handler or unit tests.
 func (k Keeper) CreateLease(ctx sdk.Context, bid types.Bid) {
 	store := ctx.KVStore(k.skey)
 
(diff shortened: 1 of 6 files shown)


x/market/keeper/keeper.go
```diff
@@ -106,8 +106,19 @@ func (k Keeper) CreateLease(ctx sdk.Context, bid types.Bid) {
...
 
-	// XXX TODO: check not overwrite
+	// assert no overwrite
+	found := store.Get(key)
+	if found != nil { // value already exists
+		var val types.Lease
+		k.cdc.MustUnmarshalBinaryBare(found, &val)
+		k.updateActiveLeaseIndex(store, val)
+		return // do not overwrite
+	}
```
boz, line 116: let's not do this, the semantics are too confusing.  return an error if you want.

I think `CreateLease` is only used in the end blocker, where this case is checked already.

x/market/keeper/key.go
```diff
@@ -43,6 +44,25 @@ func leaseKey(id types.LeaseID) []byte {
...
+	binary.Write(buf, binary.BigEndian, id.DSeq)
+	binary.Write(buf, binary.BigEndian, id.GSeq)
+	binary.Write(buf, binary.BigEndian, id.OSeq)
+	buf.Write(id.Provider.Bytes())
+	return buf.Bytes()
+}
+
+func convertLeaseActiveKey(activeKey []byte) ([]byte, error) {
+	buf := bytes.NewBuffer(leasePrefix)
+	_, err := buf.Write(activeKey[2:])
```
boz, line 59: i think here and also in the active group index it might be best to use the length of the key.

```go
	_, err := buf.Write(activeKey[len(leaseActivePrefix):])
```

Ropes: Oh I like that better!

x/market/keeper/keeper.go
Ropes: Okay I removed it, left a comment. I didn't really see a great way to handle/enforce usage otherwise.

PR #737, opened 2020-07-07
fix: commands to accept common sdk params

### LINKED ISSUE #730: `akashctl events`: not accepting common sdk params
`akashctl events` doesn't accept `--node`, `--home` which it needs to connect to the right node.

As a workaround, `AKASHCTL_NODE`, `AKASHCTL_HOME` environment variables are honored.

I think that #700 _might_ be the source of this issue.

troian:
fixes #730 issue with viper when setting flags in cobra commands from env

### DIFF
--- a/cmd/akashctl/main.go
+++ b/cmd/akashctl/main.go
@@ -2,8 +2,6 @@ package main
 
 import (
 	"net/http"
-	"os"
-	"path"
 
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/flags"
@@ -16,12 +14,13 @@ import (
 	bankcmd "github.com/cosmos/cosmos-sdk/x/bank/client/cli"
 	"github.com/rakyll/statik/fs"
 
-	"github.com/ovrclk/akash/app"
-	"github.com/ovrclk/akash/cmd/common"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"github.com/tendermint/go-amino"
 	"github.com/tendermint/tendermint/libs/cli"
+
+	"github.com/ovrclk/akash/app"
+	"github.com/ovrclk/akash/cmd/common"
 
 	// unnamed import of statik for swagger UI support
 	_ "github.com/ovrclk/akash/cmd/statik"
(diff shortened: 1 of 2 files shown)


troian:
looks like it was made to bind `cosmos` flags

boz:
`--home` and things like `--node` all come from cosmos tho.  basically everything but what's in `x/*/client/cli/flags.go` is from cosmos-sdk.

cmd/akashctl/main.go
```diff
@@ -41,8 +40,28 @@ func main() {
...
+
+		if err := viper.BindPFlag(cli.EncodingFlag, root.PersistentFlags().Lookup(cli.EncodingFlag)); err != nil {
+			return err
+		}
+
+		if err := viper.BindPFlag(cli.OutputFlag, root.PersistentFlags().Lookup(cli.OutputFlag)); err != nil {
+			return err
+		}
+
+		if err := viper.BindPFlag(flags.FlagNode, root.Flags().Lookup(flags.FlagNode)); err != nil {
```
boz, line 60: are all of these flags always present?  `akashctl keys list` for instance might not take `--node`.

It's fine to have here so long as it doesn't crash when `root.Flags().Lookup(flags.FlagNode)` returns `nil`

troian: it does indeed

cmd/akashctl/main.go
```diff
@@ -41,8 +40,25 @@ func main() {
 
 	// Add --chain-id to persistent flags and mark it required
 	root.PersistentFlags().String(flags.FlagChainID, "", "Chain ID of tendermint node")
+	root.PersistentFlags().String(flags.FlagNode, "", "The node address")
+	_ = viper.BindPFlag(flags.FlagNode, root.PersistentFlags().Lookup(flags.FlagNode))
```
boz, line 44: Why is this different than the others?

FWIW the `--node` param usually comes from `flags.GetCommands()` and `flags.PostCommands()` in `cosmos/cosmos-sdk/client/flags/flags.go`.  See `x/deployment/client/cli/{query,tx}.go`.

The `events` command is not wrapped in that; I think you should just add it in `events/cmd/root.go`.

```go
		c.Flags().String(FlagNode, "tcp://localhost:26657", "<host>:<port> to Tendermint RPC interface for this chain")
		viper.BindPFlag(FlagNode, c.Flags().Lookup(FlagNode))
```

events/cmd/root.go
```diff
@@ -18,6 +20,14 @@ func EventCmd(cdc *codec.Codec) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "events",
 		Short: "Prints out akash events in real time",
+		PreRunE: func(cmd *cobra.Command, _ []string) error {
+			cmd.Flags().String(flags.FlagNode, "", "The node address")
```
boz, line 24: i think you want to do this outside of `PreRunE`?

troian: right, that was rather stupid

cmd/akashctl/main.go
```diff
@@ -41,8 +40,23 @@ func main() {
 
 	// Add --chain-id to persistent flags and mark it required
 	root.PersistentFlags().String(flags.FlagChainID, "", "Chain ID of tendermint node")
+
 	root.PersistentPreRunE = func(_ *cobra.Command, _ []string) error {
-		return initConfig(root)
+		// viper bindings below should be applied to root command rather then
+		// to an argument from this function. otherwise viper won't able to find them
```
Ropes, line 46: That is annoying, but if it works :ship:

events/cmd/root.go
```diff
@@ -25,6 +27,9 @@ func EventCmd(cdc *codec.Codec) *cobra.Command {
 		},
 	}
 
+	cmd.Flags().String(flags.FlagNode, "", "The node address")
+	_ = viper.BindPFlag(flags.FlagNode, cmd.Flags().Lookup(flags.FlagNode))
```
Ropes, line 31: Curious: Is this FlagNode variable related to the other changes or just a fix while editing things?

troian: that ensures flag `node` processed for `events` command

Issue #715, opened 2020-06-27
feat: provider config - required attributes


boz:
We need a provider services setting that limits what orders it will bid on.  It will currently bid on any order whose required attributes match it.  For example, if the order has zero attributes then all providers will bid on it, regardless of their attributes.

Motivation: I want to set up a small provider that bids on only *my* orders.  I'd put an attribute like `owner: abozanich-c0ff33` on both the order and the provider.  But the provider will still bid on orders that don't have the `owner` attribute.

In other words, we need to add the inverse of [this](https://github.com/ovrclk/akash/blob/84b62e3327fa615fcae4d231b5eb17c42bfd1dad/provider/bidengine/order.go#L233-L237).

PR #744, opened 2020-07-09
Add required attributes to provider config

akhilkumarpilli:
Fixes: #715

### DIFF
--- a/provider/bidengine/order.go
+++ b/provider/bidengine/order.go
@@ -185,9 +185,10 @@ loop:
 			// Begin submitting fulfillment
 			bidch = runner.Do(func() runner.Result {
 				return runner.NewResult(nil, o.session.Client().Tx().Broadcast(&mtypes.MsgCreateBid{
-					Order:    o.order,
-					Provider: o.session.Provider().Address(),
-					Price:    price,
+					Order:         o.order,
+					Provider:      o.session.Provider().Address(),
+					Price:         price,
+					RequiredAttrs: o.session.RequiredAttributes(),
 				}))
 			})
 
@@ -230,13 +231,13 @@ loop:
 
 func (o *order) shouldBid(group *dquery.Group) bool {
 
-	if !o.session.Provider().MatchReqAttributes(group.Requirements) {
+	if !o.session.MatchAttributes(group.Requirements) {
 		o.log.Debug("unable to fulfill: required attributes not matched")
 		return false
 	}
 
 	// compares group requirements with provider attributes and required attributes
-	if !group.MatchAttributes(o.session.Provider().GetAllAttributes()) {
+	if !group.MatchAttributes(append(o.session.Provider().Attributes, o.session.RequiredAttributes()...)) {
 		o.log.Debug("unable to fulfill: incompatible attributes")
 		return false
 	}
--- a/provider/bidengine/order.go
+++ b/provider/bidengine/order.go
@@ -230,8 +230,13 @@ loop:
 
 func (o *order) shouldBid(group *dquery.Group) bool {
 
-	// does provider have required attributes?
-	if !group.MatchAttributes(o.session.Provider().Attributes) {
+	if !o.session.Provider().MatchReqAttributes(group.Requirements) {
+		o.log.Debug("unable to fulfill: required attributes not matched")
+		return false
+	}
+
+	// compares group requirements with provider attributes and required attributes
+	if !group.MatchAttributes(o.session.Provider().GetAllAttributes()) {
 		o.log.Debug("unable to fulfill: incompatible attributes")
 		return false
 	}
--- a/x/market/types/msgs.go
+++ b/x/market/types/msgs.go
@@ -12,9 +12,10 @@ const (
 
 // MsgCreateBid defines an SDK message for creating Bid
 type MsgCreateBid struct {
-	Order    OrderID        `json:"order"`
-	Provider sdk.AccAddress `json:"owner"`
-	Price    sdk.Coin       `json:"price"`
+	Order         OrderID         `json:"order"`
+	Provider      sdk.AccAddress  `json:"owner"`
+	Price         sdk.Coin        `json:"price"`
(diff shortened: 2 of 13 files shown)


x/provider/query/types.go
```diff
@@ -44,3 +45,13 @@ func (obj Providers) String() string {
 func (p *Provider) Address() sdk.AccAddress {
 	return p.Owner
 }
+
+// MatchReqAttributes method compares provided attributes with specific provider require attributes
+func (p Provider) MatchReqAttributes(attrs []sdk.Attribute) bool {
```
troian, line 50: Should it be renamed to `MatchRequiredAttributes` ? `Req` is usually shortcut for `request`

x/provider/types/types.go
```diff
@@ -8,9 +8,29 @@ type Attributes []sdk.Attribute
...
+	Owner         sdk.AccAddress `json:"owner"`
+	HostURI       string         `json:"host-uri"`
+	Attributes    Attributes     `json:"attributes"`
+	ReqAttributes Attributes     `json:"req-attributes" yaml:"req-attributes"`
+}
+
+// MatchReqAttributes method compares provided attributes with specific provider require attributes
+func (p Provider) MatchReqAttributes(attrs []sdk.Attribute) bool {
+loop:
+	for _, req := range p.ReqAttributes {
```
boz, line 20: Let's only do this in one place - move the current matching routine to a utility package if you have to.

There are assumptions baked into the existing algorithm and it's surprisingly easy to get things different.  `match(a,b)` isn't the same as `match(b,a)` for instance.

akhilkumarpilli:
@boz we will have a issue if we configure it only on provider daemon.
[Here]( https://github.com/ovrclk/akash/blob/master/x/market/handler/handler.go#L50) bid creation will fail when order is configured with required attributes but provider will not have any required attributes as we get provider details from store.

boz:
That's fine.  It's just a configuration setting for the provider daemon to filter what it bids on.

boz:
If I'm running a demo on testnet I don't want the daemon to bid on orders that have no attributes, for instance.  This would solve that case.

akhilkumarpilli:
No, bid creation will not work at all if both order and your provider daemon configured with same required attributes

boz:
My provider attributes has region:us-west.  1000 orders come in that have have region: us-west.  I have capacity for 10 orders right now and I only want to bid on orders that have team-name: abozanich

boz:
An order with zero attributes matches all providers.  I don't want to bid on all orders.  This solves that.

akhilkumarpilli:
For suppose, a order is created with attributes `owner: only-akash` and even provider daemon also configured required attribute with same `owner: only-akash`. `shouldBid` function will be valid as both attributes are same. So `MsgCreateBid` will be broadcasted and process will move to `handler.go`. 

When process come to [this](https://github.com/ovrclk/akash/blob/master/x/market/handler/handler.go#L50), this will throw error as order attributes wiill not match with provider attributes. This is because `owner: only-akash` attribute is not present in provider of chain data.

akhilkumarpilli:
> For suppose, a order is created with attributes `owner: only-akash` and even provider daemon also configured required attribute with same `owner: only-akash`. `shouldBid` function will be valid as both attributes are same. So `MsgCreateBid` will be broadcasted and process will move to `handler.go`.
> 
> When process come to [this](https://github.com/ovrclk/akash/blob/master/x/market/handler/handler.go#L50), this will throw error as order attributes wiill not match with provider attributes. This is because `owner: only-akash` attribute is not present in provider of chain data.

@boz , configured from provider daemon. But to fix this issue, we need to add `RequiredAttrs` field in `MsgCreateBid` struct.

boz:
@akhilkumarpilli this is for an _additional_ filter - it would happen on the provider side _after_ the current order/provider matching.  If the order doesn't match the provider then this additional filter is irrelevant.

Think of it as being similar to a provider daemon that is configured to not bid lower than `100akash` on an order - if an order comes in with a max-price of `50akash`, then the provider won't bid on it even when the order matches the provider.

anilCSE:
> additional

True @boz  but the problem is with https://github.com/ovrclk/akash/blob/0a664fcbf64c3d6a0865d680191efcdcff06d59d/x/market/handler/handler.go#L50

If we need this filter, we need to enable this, we need to have this information in order attributes. But, order attributes are all required (for now) and if we don't store this optional attribute on-chain, even our provider will not have the attribute required. So, it won't bid.

One possible solution is, introducing optional attributes on both `order` and bidding config for the `provider`

x/market/types/msgs.go
```diff
@@ -12,9 +12,10 @@ const (
 
 // MsgCreateBid defines an SDK message for creating Bid
 type MsgCreateBid struct {
-	Order    OrderID        `json:"order"`
-	Provider sdk.AccAddress `json:"owner"`
-	Price    sdk.Coin       `json:"price"`
+	Order         OrderID         `json:"order"`
+	Provider      sdk.AccAddress  `json:"owner"`
+	Price         sdk.Coin        `json:"price"`
+	RequiredAttrs []sdk.Attribute `json:"attributes"`
```
boz, line 18: this shouldn't be here.  the provider just chooses to not bid if these new attributes don't match.

again, it's similar to the price being lower than what the provider is willing to bid on.

provider/bidengine/order.go
```diff
@@ -230,8 +231,13 @@ loop:
...
 
-	// does provider have required attributes?
-	if !group.MatchAttributes(o.session.Provider().Attributes) {
+	if !o.session.MatchAttributes(group.Requirements) {
+		o.log.Debug("unable to fulfill: required attributes not matched")
+		return false
+	}
+
+	// compares group requirements with provider attributes and required attributes
+	if !group.MatchAttributes(append(o.session.Provider().Attributes, o.session.RequiredAttributes()...)) {
```
boz, line 240: I don't think that this works.  this is `match(a,b+c) = match(a,b) && match(a,c)`, but we want `match(a,b) && match(c,a)`.

Issue #727, opened 2020-07-01
ids: more friendly string representation


boz:
For all of the IDs, make a more friendly string representation.  Most important is to have the account address format rather than hex.

For example, logs say things like this:

```
I[2020-07-01|21:30:37.501] shutting down module=bidengine-order order="{Owner:9F97543245F06F406D82D0D92D3D74337484A143 DSeq:14690 GSeq:1 OSeq:1}"
```

The `Owner` field above should be represented as `akash1n7t4gvj97ph5qmvz6rvj60t5xd6gfg2rm5dh6x`.

Secondly, it would be better to have things in [path](https://github.com/ovrclk/akash/blob/3310cc668be406a19029ce458831a53f16345198/x/market/query/path.go#L36) format:

`"akash1n7t4gvj97ph5qmvz6rvj60t5xd6gfg2rm5dh6x/14690/1/1"`

Ropes:
Implemented a pretty printer in the `sdkutil` package, everything looks nice in the tests when the global config is set via:
```go
	config := sdk.GetConfig()
	config.SetBech32PrefixForAccount("akash", "akash")
	config.Seal()
```
However when tests are run outside of that package, they'll show `cosmos` as the address denom. So tests will look a little funky for all types which get the stringer implementation.

Still need to implement a lot of String()'er interfaces.

boz:
Can you call [`common.InitSDKConfig()`](https://github.com/ovrclk/akash/blob/c5cf6e255ca506a88fb387c4a65ff87c5ee4dbf1/cmd/common/sdk.go#L19) from `testutil.init()`?  It might cause circular dependencies, not sure.

It's really not important, tho - the bech32 prefixes don't matter in tests AFAIK.

Ropes:
How to best edit the global config for testing is still a little bit of a head scratcher, but logs now look like this:
```
I[2020-07-08|17:03:30.204] group fetched                                module=bidengine-order order=akash1un2hh75y0nqrg0fz035cg3spcphe78jhrwmr68/1/1/1
D[2020-07-08|17:03:30.204] reservation requested                        module=provider-cluster cmp=service cmp=inventory-service order=akash1un2hh75y0nqrg0fz035cg3spcphe78jhrwmr68/1/1/1 resources=akash1un2hh75y0nqrg0fz035cg3spcphe78jhrwmr68/1/1
[...]

PR #745, opened 2020-07-09
feat:727 Friendly stringer of Block ID types

Ropes:
Provide pretty printer for Block address types with AccAddress, DSeq,
GSeq, and OSeq fields which will be printed to logs and command
responses.

`sdkutil` package provides helper methods to format and parse values
from the human readable format.

```
Format: [Owner AccAddress]/[DSeq]/[GSeq]/[OSeq]/[Provider AccAddress]
eg: akash1vgv30hr8lel8r7270wywmak5c82es3njws3u4z/795423625/1/1/akash1pxqksfr60zc6uadht6300uavwl3s58ct0apt9w
```

fixes: #727

### DIFF
--- a/sdkutil/event.go
+++ b/sdkutil/event.go
@@ -1,10 +1,10 @@
 package sdkutil
 
 import (
-	"errors"
 	"strconv"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/pkg/errors"
 )
 
 const (
@@ -23,6 +23,10 @@ var (
 	ErrUnknownModule = errors.New("Unknown module")
 	// ErrUnknownAction is the error with message "Unknown action"
 	ErrUnknownAction = errors.New("Unknown action")
+	// ErrParsingBlockID indicates one of the uint parsers failed to convert a value.
+	ErrParsingBlockID = errors.New("error parsing block id values")
+	// ErrInvalidParseBlockIDInput indicates the splitting of block path failed.
+	ErrInvalidParseBlockIDInput = errors.New("error parsing block id input string")
 )
 
 type BaseModuleEvent struct {
--- a/sdkutil/object_id.go
+++ b/sdkutil/object_id.go
@@ -1,0 +1,182 @@
+package sdkutil
+
+import (
+	"path"
+	"path/filepath"
+	"reflect"
+	"strconv"
+	"strings"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/pkg/errors"
+)
+
+const (
+	blockIDOwnerPosition   = 0
+	blockIDDSeqPosition    = 1
+	blockIDGSeqPosition    = 2
+	blockIDOSeqPosition    = 3
+	blockIDProviderPostion = 4
+)
+
+// BlockID provides type to properly format print blockchain ID types.
+type BlockID struct {
+	Owner    *sdk.AccAddress
+	DSeq     *uint64
+	GSeq     *uint32
+	OSeq     *uint32
+	Provider *sdk.AccAddress
+}
+
+// NewBlockID initializes the BlockID struct, passing nil values is acceptable
+// for fields which are not used.
+func NewBlockID(owner *sdk.AccAddress, dseq *uint64, gseq *uint32, oseq *uint32, provider *sdk.AccAddress) BlockID {
+	return BlockID{
+		Owner:    owner,
+		DSeq:     dseq,
+		GSeq:     gseq,
+		OSeq:     oseq,
+		Provider: provider,
+	}
+}
+
+// String method provides the full formating of all the BlockID fields which are not nil.
+// Format: [Owner AccAddress]/[DSeq]/[GSeq]/[OSeq]/[Provider AccAddress]
+// eg: akash1vgv30hr8lel8r7270wywmak5c82es3njws3u4z/795423625/1/1/akash1pxqksfr60zc6uadht6300uavwl3s58ct0apt9w
+func (b BlockID) String() string {
+	parts := make([]string, 0)
+	if b.Owner != nil {
+		parts = append(parts, b.Owner.String())
+	}
+	if b.DSeq != nil {
+		parts = append(parts, strconv.FormatUint(*b.DSeq, 10))
+	}
+	if b.GSeq != nil {
+		parts = append(parts, strconv.FormatUint(uint64(*b.GSeq), 10))
+	}
+	if b.OSeq != nil {
+		parts = append(parts, strconv.FormatUint(uint64(*b.OSeq), 10))
+	}
+	if b.Provider != nil {
+		parts = append(parts, b.Provider.String())
+	}
+	return path.Join(parts...)
+}
+
+// ReflectBlockID accepts an ID struct and unpacks the block identifying fields.
+// Useful for passing *ID types and getting consistent output based on the consistent
+// field names.
+func ReflectBlockID(id interface{}) *BlockID {
+	b := BlockID{}
+
+	val := reflect.ValueOf(id)
+	typeOfID := val.Type()
+
+	for i := 0; i < val.NumField(); i++ {
+		field := val.Field(i)
+		switch typeOfID.Field(i).Name {
+		case "Owner":
+			if field.Type() == reflect.TypeOf(sdk.AccAddress{}) {
+				x, ok := field.Interface().(sdk.AccAddress)
+				if ok && len(x) == sdk.AddrLen {
+					b.Owner = &x
+				}
+			}
+		case "DSeq":
+			if field.Type() == reflect.TypeOf(uint64(0)) {
+				x, ok := field.Interface().(uint64)
+				if ok {
+					b.DSeq = &x
+				}
+			}
+		case "GSeq":
+			if field.Type() == reflect.TypeOf(uint32(0)) {
+				x, ok := field.Interface().(uint32)
+				if ok {
+					b.GSeq = &x
+				}
+			}
+		case "OSeq":
+			if field.Type() == reflect.TypeOf(uint32(0)) {
+				x, ok := field.Interface().(uint32)
+				if ok {
+					b.OSeq = &x
+				}
+			}
+		case "Provider":
+			if field.Type() == reflect.TypeOf(sdk.AccAddress{}) {
+				x, ok := field.Interface().(sdk.AccAddress)
+				if ok && len(x) == sdk.AddrLen {
+					b.Provider = &x
+				}
+			}
+		default:
+			// Skip this field
+		}
+	}
+	return &b
+}
+
+// FmtBlockID provides a human readable representation of the block chain ID fields.
+func FmtBlockID(owner *sdk.AccAddress, dseq *uint64, gseq *uint32, oseq *uint32, provider *sdk.AccAddress) string {
+	return BlockID{
+		Owner:    owner,
+		DSeq:     dseq,
+		GSeq:     gseq,
+		OSeq:     oseq,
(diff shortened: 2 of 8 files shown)


sdkutil/event.go
```diff
@@ -87,3 +95,150 @@ func GetString(attrs []sdk.Attribute, key string) (string, error) {
 	}
 	return "", ErrNotFound
 }
+
+// BlockIDer returns a type which supports the String'er interface.
+type BlockIDer interface {
```
Ropes, line 100: Delete

x/provider/handler/handler.go
```diff
@@ -63,6 +63,7 @@ func handleMsgUpdate(ctx sdk.Context, keeper keeper.Keeper, mkeeper mkeeper.Keep
 	var err error
 
 	// all filtering code below is madness!. should make an index to not melt the cpu
+	// TODO: use WithActiveLeases, filter by lease.Provider
```
boz, line 66: :+1:

sdkutil/event.go
```diff
@@ -1,17 +1,30 @@
...
 )
 
 const (
 	akashEventMessageV1 = "akash.v1"
 
 	// EventTypeMessage defines the Akash message string
 	EventTypeMessage = akashEventMessageV1
+
+	akashHexLen    = 20
+	akashBech32Len = 44
```
boz, line 21: are any components of these defined in sdk or tm?

sdkutil/event.go
```diff
@@ -87,3 +104,172 @@ func GetString(attrs []sdk.Attribute, key string) (string, error) {
...
 }
+
+// BlockID provides type to properly format print blockchain ID types.
+type BlockID struct {
+	Owner    *sdk.AccAddress
+	DSeq     *uint64
+	GSeq     *uint32
+	OSeq     *uint32
+	Provider *sdk.AccAddress
+}
```
boz, line 115: cool.  i think that this can maybe go in a package of its own, but for now put it in its own file - `object_id.go` or something.

sdkutil/event.go
```diff
@@ -87,3 +104,172 @@ func GetString(attrs []sdk.Attribute, key string) (string, error) {
...
+// Format: [Owner AccAddress]/[DSeq]/[GSeq]/[OSeq]/[Provider AccAddress]
+// eg: akash1vgv30hr8lel8r7270wywmak5c82es3njws3u4z/795423625/1/1/akash1pxqksfr60zc6uadht6300uavwl3s58ct0apt9w
+func (b BlockID) String() string {
+	parts := make([]string, 0)
+	if b.Owner != nil {
+		parts = append(parts, b.Owner.String())
+	}
+	if b.DSeq != nil {
+		parts = append(parts, strconv.FormatUint(*b.DSeq, 10))
+	}
```
boz, line 139: if `b.Owner == nil` then `b.DSeq` should be nil, etc...

:thinking: probably not important to enforce here tho.

sdkutil/event.go
```diff
@@ -87,3 +104,172 @@ func GetString(attrs []sdk.Attribute, key string) (string, error) {
...
+// ReflectBlockID accepts an ID struct and unpacks the block identifying fields.
+// Useful for passing *ID types and getting consistent output based on the consistent
+// field names.
+func ReflectBlockID(id interface{}) *BlockID {
+	b := BlockID{}
+
+	val := reflect.ValueOf(id)
+	typeOfID := val.Type()
+
+	for i := 0; i < val.NumField(); i++ {
```
boz, line 161: if the `i`th value is `nil` then the `i+1`th value should be `nil`.

again not sure if we should enforce here.

sdkutil/event.go
```diff
@@ -87,3 +104,172 @@ func GetString(attrs []sdk.Attribute, key string) (string, error) {
...
+	if len(parts[blockIDOwnerPosition]) == akashBech32Len {
+		aa, err := sdk.AccAddressFromBech32(parts[blockIDOwnerPosition])
+		if err != nil {
+			return nil, errors.Wrap(ErrParsingBlockID, err.Error())
+		} else if len(aa) > 0 {
+			b.Owner = &aa
+		}
+	} else {
+		return nil, errors.Wrap(ErrInvalidParseBlockIDInput, "owner address bytes invalid")
+	}
```
boz, line 235: :nail_care: invert these and return early instead of nesting the "happy path"

```go
if len(x) != expected {
  return err
}
// ...
```

basically remove all `else` and `else if`s.

also, I believe that `sdk.AccAddressFromBech32` checks the length, we don't need to do it again.

sdkutil/event.go
```diff
@@ -87,3 +104,172 @@ func GetString(attrs []sdk.Attribute, key string) (string, error) {
...
+		}
+
+		if len(parts[blockIDOSeqPosition]) > 0 {
+			oseq, err := strconv.ParseUint(parts[3], 10, 64)
+			if err != nil {
+				return nil, errors.Wrap(ErrParsingBlockID, err.Error())
+			}
+			x := uint32(oseq)
+			b.OSeq = &x
+		}
```
boz, line 264: what if `len(parts) == 3`?

All of the `seqs` should never be zero, you probably don't need to check the string length - I think these should all be guarded by `len(parts) > blockFooPosition` rather than `len(parts[blockFooPosition])`.

sdkutil/event_test.go
```diff
@@ -0,0 +1,177 @@
...
+		b, err := sdkutil.ParseBlockID(errStr)
+		assert.Error(t, err)
+		assert.Nil(t, b)
+	})
+	t.Run("error parsing provider", func(t *testing.T) {
+		errStr := "akash19jeunqhqy04wj5ee2t3kq2l6uee53fqkmp878d/1478600512/2170091657/2036973947/akash18u664a4m558u79vecdne93fwd80nhfgd9svxxx"
+		b, err := sdkutil.ParseBlockID(errStr)
+		assert.Error(t, err)
+		assert.Nil(t, b)
+	})
```
boz, line 142: use table-driven tests for these

sdkutil/event_test.go
```diff
@@ -0,0 +1,177 @@
...
+	t.Logf("block: %q", block)
+
+	bid := testutil.BidID(t)
+	block = sdkutil.ReflectBlockID(bid)
+	t.Logf("bidID: %q", block)
+
+	str := block.String()
+	t.Logf("%q", str)
+	_, err := sdkutil.ParseBlockID(str)
+	assert.NoError(t, err)
```
boz, line 157: going to add assertions, yes?  make sure to get all of the id types.

testutil/base.go
```diff
@@ -11,6 +11,12 @@ import (
 	dtypes "github.com/ovrclk/akash/x/deployment/types"
 )
 
+func init() {
+	config := sdk.GetConfig()
+	config.SetBech32PrefixForAccount(CoinDenom, CoinDenom)
+	config.Seal()
```
boz, line 17: Let's see how this does.  The race detector may complain about it at some point.

Ropes: Race detector is happy so far.

sdkutil/event.go
Ropes: That's a better way to check; done!

sdkutil/event.go
Ropes: I'm also not sure if the ordering of val.NumField is consistent with how the struct 'looks', so I wrote this to be agnostic of any ordering, which I think is more robust.

sdkutil/event.go
Ropes: `akashBech32Len` is custom due to `akash` prefix to the string, the AddrLen can be pulled from `sdk`. I've updated the const to match sdk value, could just replace it.

sdkutil/event.go
Ropes: I think I've improved the if/elses in the latest changes.

sdkutil/object_id.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// Owner Address field always expected
+	aa, err := sdk.AccAddressFromBech32(parts[blockIDOwnerPosition])
+	if err != nil {
+		return nil, errors.Wrap(ErrParsingBlockID, err.Error())
+	} else if len(aa) > 0 {
+		b.Owner = &aa
+	}
+
+	if len(parts) > 1 && len(parts) < 6 {
```
boz, line 148: unnecessary check.

sdkutil/object_id_test.go
```diff
@@ -0,0 +1,188 @@
...
+	str := block.String()
+	t.Logf("%q", str)
+	parsed, err := sdkutil.ParseBlockID(str)
+	assert.NoError(t, err)
+
+	assert.Equal(t, *block.Owner, *parsed.Owner)
+	assert.Equal(t, *block.DSeq, *parsed.DSeq)
+	assert.Equal(t, *block.GSeq, *parsed.GSeq)
+	assert.Equal(t, *block.OSeq, *parsed.OSeq)
+	assert.Equal(t, *block.Provider, *parsed.Provider)
```
boz, line 158: Do this for all ID types:

```
id := testutil.FooID(t)
block := sdkutil.ReflectBlockID(id)
assert.Equal(t,id.Foo,*block.Foo)
assert.Equal(t,id.Bar,*block.Bar)
// ...
```

sdkutil/object_id_test.go
```diff
@@ -0,0 +1,188 @@
...
+	if len(split) != 3 {
+		t.Error("expected returned path have three elements")
+	}
+}
+
+func TestOrderID(t *testing.T) {
+	oid := testutil.OrderID(t)
+
+	block := sdkutil.ReflectBlockID(oid)
+	t.Logf("%s", block)
```
boz, line 175: ?

Ropes: Assertions added.

sdkutil/object_id_test.go
Ropes: That is clearer; updated!

sdkutil/object_id_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+func TestOrderID(t *testing.T) {
+	oid := testutil.OrderID(t)
+	block := sdkutil.ReflectBlockID(oid)
+
+	assert.Equal(t, oid.Owner, *block.Owner)
+	assert.Equal(t, oid.DSeq, *block.DSeq)
+	assert.Equal(t, oid.GSeq, *block.GSeq)
+	assert.Equal(t, oid.OSeq, *block.OSeq)
+}
```
boz, line 171: `LeaseID`?  `DeploymentID`?

sdkutil/object_id_test.go
```diff
@@ -0,0 +1,182 @@
...
+)
+
+const akashHexLen = 20
+const akashBech32Len = 44
+
+func accAddress(t testing.TB) sdk.AccAddress {
+	t.Helper()
+	privKey := ed25519.GenPrivKey()
+	return sdk.AccAddress(privKey.PubKey().Address())
+}
```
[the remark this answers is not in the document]
Ropes: :facepalm: that was definitely leftover prior to testutil getting the method.

sdkutil/object_id_test.go
Ropes: I think I got them all. :heavy_check_mark:

testutil/base.go
boz, line 17: this is only going to cause problems.  let's not be hesitant to back it out.

Ropes: Agreed, I'm not sure what the better solution is though.

boz: In principal the tests shouldn't care about the coin denomination.

PR #746, opened 2020-07-10
Command to listen for lease creation and send manifest automatically

jackzampolin:
I'm having an issue seeing the events. Nothing is printing after transaction gets sent. Repro for testing (from /_run/kube):

T1:  KIND_CONFIG=kind-config-80.yaml make kind-cluster-delete kind-cluster-create
T2: make clean init node-run
T3: make provider-create provider-run
T4: make bins && ../../akashctl --home ./cache/client/ tx deployment make-it-so deployment.yaml --from main

### DIFF
--- a/deploy/deploy.go
+++ b/deploy/deploy.go
@@ -1,18 +1,20 @@
 package deploy
 
 import (
-	ccontext "context"
+	"context"
 	"fmt"
 	"os"
 
-	"github.com/cosmos/cosmos-sdk/client/context"
-	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/spf13/cobra"
 	"golang.org/x/sync/errgroup"
 
+	clientCtx "github.com/cosmos/cosmos-sdk/client/context"
+	"github.com/cosmos/cosmos-sdk/client/flags"
 	"github.com/cosmos/cosmos-sdk/codec"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/x/auth"
-	"github.com/cosmos/cosmos-sdk/x/auth/client/utils"
+	authUtils "github.com/cosmos/cosmos-sdk/x/auth/client/utils"
+
 	"github.com/ovrclk/akash/events"
 	"github.com/ovrclk/akash/provider/gateway"
 	"github.com/ovrclk/akash/provider/manifest"
@@ -32,9 +34,10 @@ func CMD(cdc *codec.Codec) *cobra.Command {
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// Setup CLI context and transaction builder
-			ctx := context.NewCLIContext().WithCodec(cdc)
-			bldr := auth.NewTxBuilderFromCLI(os.Stdin).WithTxEncoder(utils.GetTxEncoder(cdc))
-			if err := ctx.Client.Start(); err != nil {
+			cliCtx := clientCtx.NewCLIContext().WithCodec(cdc)
+
+			// Start the tendermint client
+			if err := cliCtx.Client.Start(); err != nil {
 				return err
 			}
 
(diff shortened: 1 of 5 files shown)


x/deployment/client/cli/tx.go
```diff
@@ -32,10 +40,152 @@ func GetTxCmd(key string, cdc *codec.Codec) *cobra.Command {
...
+								&manifest.SubmitRequest{
+									Deployment: msg.ID.DeploymentID(),
+									Manifest:   mani,
+								},
+							); err != nil {
+								return err
+							}
+
+							if leases = leases - 1; leases == 0 {
+								fmt.Printf("Leases left %d...\n", leases)
```
jackzampolin, line 148: Planning on exiting the command here, but unable to make it exit cleanly with the current code.

jackzampolin: have this exiting cleanly!

x/deployment/client/cli/tx.go
```diff
@@ -32,10 +40,154 @@ func GetTxCmd(key string, cdc *codec.Codec) *cobra.Command {
...
+			leases := len(groups)
+			group.Go(func() error {
+				for {
+					select {
+					case <-sub.Done():
+						return nil
+					case ev := <-sub.Events():
+						switch msg := ev.(type) {
+						case types.EventDeploymentCreated:
+							fmt.Printf("Deployment %d created...\n", msg.ID.DSeq)
```
Ropes, line 121: Could we use the ctx logger instead of fmt.Print?

x/deployment/client/cli/tx.go
```diff
@@ -32,10 +40,154 @@ func GetTxCmd(key string, cdc *codec.Codec) *cobra.Command {
...
+					return err
+				}
+			}
+
+			// create a new pubsub bus to handle chain events
+			bus := pubsub.NewBus()
+			defer bus.Close()
+
+			// Create an error group to handle chain listener
+			group, cctx := errgroup.WithContext(ccontext.Background())
```
Ropes, line 97: :heart:

x/deployment/client/cli/tx.go
```diff
@@ -32,10 +40,154 @@ func GetTxCmd(key string, cdc *codec.Codec) *cobra.Command {
...
+			if err != nil {
+				return err
+			}
+
+			// Default DSeq to the current block height
+			if id.DSeq == 0 {
+				if id.DSeq, err = currentBlockHeight(ctx); err != nil {
+					return err
+				}
+			}
```
Ropes, line 90: We should wrap all the Msg creation including the 
```
			msg := types.MsgCreateDeployment{
				ID: id,
				// TODO: Version:  []byte{0x1, 0x2},
				Groups: make([]types.GroupSpec, 0, len(groups)),
			}
```
 into a utility function. It's also called in normal Create, and it'd be good not to have drift.

x/deployment/client/cli/tx.go
```diff
@@ -32,10 +40,154 @@ func GetTxCmd(key string, cdc *codec.Codec) *cobra.Command {
...
 		cmdGroupClose(key, cdc),
+		makeItSoCommand(key, cdc),
 	)...)
 	return cmd
 }
 
+func makeItSoCommand(key string, cdc *codec.Codec) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "make-it-so [sdl-file]",
+		Short: fmt.Sprintf("Create %s", key),
```
Ropes, line 51: Still `Create deployment`, might want to add a little explanation here for the command.

jackzampolin:
Naming could def use some work 🤣 Also based on the conversation earlier today I think functionality similar to this could be blown out into a whole seperate `client-services` binary.

x/deployment/client/cli/tx.go
jackzampolin: Thats a much better idea.

x/deployment/client/cli/tx.go
jackzampolin: Agree, I'll add this.

x/deployment/client/cli/tx.go
jackzampolin: So turns out you can't get a logger from the `ctx` 🤦

boz: the sdk client context - `ctx.PrintOutput()` maybe - it formats in json or yaml depending on flags or whatever.

jackzampolin: Yeah I want this to look more "log"-y. Think I'll emulate whats in the provider daemon.

deploy/deploy.go
```diff
@@ -128,7 +128,7 @@ func CMD(cdc *codec.Codec) *cobra.Command {
 								return err
 							}
 
-							if leases = leases - 1; leases == 0 {
+							if leases--; leases == 0 {
```
troian, line 131: 😍

jackzampolin:
Built this as a separate binary too if you want to give that a try: https://github.com/jackzampolin/deploy

Issue #565, opened 2020-05-12
deployment version


boz:
A deployment has two components: the on-chain resource model used for the auction, and the off-chain description of _what_ is to be deployed called the _manifest_.

The manifest is sent to providers, and providers need a way to verify that the manifest they sent.

The "deployment version" is a hash of the manifest and resides in the on-chain `deployment.Version`.

* [x] `x/deployment/types`: require version in `MsgCreate`.
* [x] `x/deployment/client/cli`: calculate deployment version and include in `MsgCreate`.
* [x] `x/deployment/handler`: consume version from `MsgCreate`.

To do this correctly we need to have a stable serialization of the manifest.  This means sorting all components consistently.  This is currently done in `sdl/v1.go` when parsing an sdl file; we should probably move that functionality to an isolated package to be used elsewhere.

Ropes:
Version/Hashing of the SDL sorted though somewhat inefficient, it's not in a critical compute bottleneck path.

Ropes:
Unit tests showing functional passing of Version from Create to Update messages *in keepers*. Still plumbing the `akashctl` commands.

Some was referencing the `Version` field type as an sdk.AccAddress. From reading some commented code, seems like there was an initial idea to integrate the owner address into the version, but this implementation just uses the sha256 of the SDL struct.

PR #748, opened 2020-07-14
feat:565 Deployment Version Field

Ropes:
Implementation of passing a Deployment manifest's Version from Create and Update commands and messages.

* `sdl.VersionSha` package produces deterministic version hash from SDL struct.
* `akashctl tx deployment [create|update]` generates and sets Version field.
* `testutil` package updated to provide a consistent Deployment.Version value.
  * `testutil.DeploymentVersion()` provides random version sums.
* Assertions/test that the Version is passed to the keeper via
`MsgUpdateDeployment` & `MsgCreateDeployment`.
  * MsgUpdateDeployment passes GroupSpec to update image tags
* Test integration tests for `deployment update`.
  * New deployment-v2.yaml with tag change.
  * Assert's Version field is changed post update.
* Tests to validate different deployment Versions
  * Compare differing deployment yaml sha256 sums in tests.
* Deployment* SDKEvent types carry Version field.
  * `EventDeploymentCreated` & `EventDeploymentUpdated`
  * Structs and constructors updated with Version field.
  * Tests updated to create and pass Versions for Create & Update.
* Test Simulations execute Update Deployments with different manifest
file.

fixes: #565

### DIFF
--- a/sdl/sdl.go
+++ b/sdl/sdl.go
@@ -1,12 +1,17 @@
 package sdl
 
 import (
+	"crypto/sha256"
+	"encoding/json"
 	"io/ioutil"
+
+	yaml "gopkg.in/yaml.v2"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	"github.com/ovrclk/akash/manifest"
 	"github.com/ovrclk/akash/validation"
 	dtypes "github.com/ovrclk/akash/x/deployment/types"
-	yaml "gopkg.in/yaml.v2"
 )
 
 // SDL is the interface which wraps Validate, Deployment and Manifest methods
@@ -27,7 +32,7 @@ func ReadFile(path string) (SDL, error) {
 
 // Read reads buffer data and returns SDL instance
 func Read(buf []byte) (SDL, error) {
-	// TODO: handle versions
+	// TODO: handle versions; read 'version' field and switch
 	obj := &v1{}
 	if err := yaml.Unmarshal(buf, obj); err != nil {
 		return nil, err
@@ -55,6 +60,8 @@ func Read(buf []byte) (SDL, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// TODO: Determine if worth repairing ValidateManifest; functionality is commented out
 	if err := validation.ValidateManifest(m); err != nil {
 		return nil, err
 	}
@@ -63,4 +70,26 @@ func Read(buf []byte) (SDL, error) {
 	// }
 
 	return obj, nil
+}
+
+// Version creates the deterministic Deployment Version hash from the SDL.
+// Sha256 returns 32 byte  sum of the SDL.
+func Version(s SDL) ([]byte, error) {
+	manifest, err := s.Manifest()
+	if err != nil {
+		return nil, err
+	}
+
+	m, err := json.Marshal(manifest)
+	if err != nil {
+		return nil, err
+	}
+
+	sortedBytes, err := sdk.SortJSON(m)
+	if err != nil {
+		return nil, err
+	}
+
+	sum := sha256.Sum256(sortedBytes)
+	return []byte(sum[:]), nil
 }
--- a/sdl/v1.go
+++ b/sdl/v1.go
@@ -17,8 +17,10 @@ var (
 	allowedVersion = semver.MustParse("1.5.0")
 )
 
+var _ SDL = (*v1)(nil) // Assert interface implementation
+
 type v1 struct {
-	Version  string
+	Version  string   `yaml:"version"`
 	Include  []string `yaml:",omitempty"`
 	Services map[string]v1Service
 	Profiles v1Profiles
--- a/sdl/v1_test.go
+++ b/sdl/v1_test.go
@@ -3,7 +3,7 @@ package sdl_test
 import (
 	"testing"
 
-	"github.com/ovrclk/akash/sdl"
+	sdlv1 "github.com/ovrclk/akash/sdl"
 	"github.com/ovrclk/akash/types"
 	"github.com/ovrclk/akash/types/unit"
 	"github.com/stretchr/testify/assert"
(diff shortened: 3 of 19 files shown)


sdl/sdl.go
```diff
@@ -55,6 +60,8 @@ func Read(buf []byte) (SDL, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// TODO: Determine if worth repairing ValidateManifest; functionality is commented out
```
boz, line 64: yes, we have to.  doesn't need to be in this PR.

sdl/sdl.go
```diff
@@ -64,3 +71,20 @@ func Read(buf []byte) (SDL, error) {
 
 	return obj, nil
 }
+
+// VersionSha creates the deterministic Deployment Version hash from the SDL.
+// Sha256 returns 32 byte  sum of the SDL.
+func VersionSha(s SDL) ([]byte, error) {
```
boz, line 77: `Version` is sufficient.

sdl/v1.go
```diff
@@ -17,8 +17,10 @@ var (
 	allowedVersion = semver.MustParse("1.5.0")
 )
 
+var _ SDL = (*v1)(nil) // Assert interface implementation
+
 type v1 struct {
-	Version  string
+	Version  string   `yaml:"version"`
```
boz, line 23: :+1:

sdl/v1_test.go
```diff
@@ -16,14 +16,30 @@ const (
 	randStorage uint64 = 1 * unit.Gi
 )
 
-func Test_v1_Parse_docs(t *testing.T) {
-	sdl, err := sdl.ReadFile("../x/deployment/testdata/deployment.yaml")
+func Test_v1_Parse_Deployments(t *testing.T) {
+	SDLv1, err := sdl.ReadFile("../x/deployment/testdata/deployment.yaml")
```
boz, line 20: lowercase if you must.  maybe better to use aliased import instead - `import sdlv1 "github.com/ovrclk/akash/sdl/v1"`

x/deployment/client/cli/tx.go
```diff
@@ -126,8 +131,32 @@ func cmdUpdate(key string, cdc *codec.Codec) *cobra.Command {
...
+			}
+
+			version, err := sdl.VersionSha(sdlManifest)
+			if err != nil {
+				return err
+			}
+			_, err = fmt.Fprint(ctx.Output, version)
+			if err != nil {
+				return err
+			}
```
boz, line 150: :scissors:

x/deployment/handler/handler_test.go
```diff
@@ -145,8 +148,17 @@ func TestUpdateDeploymentExisting(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, res)
 
+	t.Run("assert deployment version", func(t *testing.T) {
+		d, ok := suite.dkeeper.GetDeployment(suite.ctx, deployment.DeploymentID)
+		require.True(t, ok)
+		assert.Equal(t, d.Version, testutil.Sha256Nonce[:])
+	})
+
+	depSum := sha256.Sum256([]byte("ohai"))
```
boz, line 157: can you just use `testutil.Sha256Nonce`?  also, call it something else - `tetsutil.DefaultVersion`.

x/deployment/simulation/operations.go
```diff
@@ -98,15 +98,19 @@ func SimulateMsgCreateDeployment(ak govtypes.AccountKeeper, k keeper.Keeper) sim
 			return simulation.NoOpMsg(types.ModuleName), nil, nil
 		}
 
-		sdl, readError := sdl.ReadFile("../x/deployment/testdata/deployment.yaml")
+		SDL, readError := sdl.ReadFile("../x/deployment/testdata/deployment.yaml")
```
boz, line 101: same - import as `sdlv1` or just keep it as-is.

x/deployment/handler/handler_test.go
Ropes: The objective of the test was to run an update and confirm that the Version was updated with the new value. Thus the new Sum.

I set the second value to be the hash of the `DefaultDeploymentVersion` value, seem reasonable?

sdl/v1_test.go
Ropes: I agree, it makes sense to split apart `sdl` package into `sdl/v1` `sdl/vN` etc. It's imported in a lot of places though, so I'd rather not refactor all of that for this PR. I've updated the import alias and reset the variable used to `sdl`.

Shall I open up an issue for splitting up versioned parts of `sdl` into `sdl/v...` packages?

sdl/sdl.go
```diff
@@ -64,3 +71,20 @@ func Read(buf []byte) (SDL, error) {
 
 	return obj, nil
 }
+
+// Version creates the deterministic Deployment Version hash from the SDL.
+// Sha256 returns 32 byte  sum of the SDL.
+func Version(s SDL) ([]byte, error) {
+	m, err := json.Marshal(s)
```
boz, line 78: Sorry, I didn't communicate this well - it should be a hash of the manifest -- returned by `s.Manifest()` -- because that is the struct that we send to the providers.

The struct that implements the `SDL` interface is too implementation-dependent - it's only really meant for parsing our particular file, but it should be possible to construct a manifest in many different ways, and the hash of that manifest will be the version regardless of how it was built.

Additionally, since the server has the manifest and not the sdl-parsing object, it can't verify the object we send it.

x/deployment/types/event.go
```diff
@@ -39,6 +42,7 @@ func (ev EventDeploymentCreated) ToSDKEvent() sdk.Event {
 		append([]sdk.Attribute{
 			sdk.NewAttribute(sdk.AttributeKeyModule, ModuleName),
 			sdk.NewAttribute(sdk.AttributeKeyAction, evActionDeploymentCreated),
+			sdk.NewAttribute(evVersionKey, string(ev.Version)),
```
boz, line 45: this is going to be quite ugly.  let's make a type that (un)marshals hex or base36 or something - or do that by hand here.

x/deployment/types/msgs.go
```diff
@@ -64,7 +67,7 @@ func (msg MsgUpdateDeployment) ValidateBasic() error {
 		return err
 	}
 
-	if err := sdk.VerifyAddressFormat(msg.Version); err != nil {
+	if len(msg.Version) == 0 {
```
boz, line 70: this should be 224/8 bytes right?   maybe define a constant somewhere, maybe defined from the package that the hash method comes from.

x/deployment/types/event.go
Ropes: Updated to encode data as hex before writing as attribute.

x/deployment/types/msgs.go
Ropes: I modified this to match the length of the version, it lead down a rabbit hole. Thankfully saved by the integration tests.

There is some place where this check runs and it seems to silently stop the update.  Created an issue to tackle this when it's more important: https://github.com/ovrclk/akash/issues/749

Issue #740, opened 2020-07-08
order max-wait duration


boz:
Orders currently have a `StartAt` which dictates what block order matching will begin.  We need another parameter that dictates how long the order is open for matching before it's closed.

Rational: If misconfigured or otherwise unmatched orders linger in the unmatched state, it clogs the matching routine.  Also, these orphaned orders are likely no longer managed by the tenant, and if a provider comes along, say, a year later, that provider can get paid for the lease even tho they don't do anything.

Like `StartAt`, this can be hard-coded for now; we will parameterize it in another issue.

Ropes:
From @jackzampolin: Register a variable in the Genesis, then it will be able to be manged by governance params.

The final form should be the User configuring the wait period parameter for orders to be bid on.

jackzampolin:
Well in the user configurable case, it would be set either in `MsgCreateDeployment` or in the SDL. The gov param would be if it's a chain wide parameter.

PR #750, opened 2020-07-17
feat:740 Order Bidding TTL

Ropes:
Provide an automatic cancellation of Market Orders. A maximum Order
duration is specifed as a

Current Order bidding duration is 1 day. Assuming 7s/block written; this
totals to 12342 blocks for one day. If the chain height is beyond that,
then the order is signaled to be closed.

fixes: #740


x/market/types/errors.go
```diff
@@ -68,4 +71,10 @@ var (
 	ErrOrderMatched = sdkerrors.New(ModuleName, errCodeOrderMatched, "order matched")
 	// ErrOrderClosed order closed
 	ErrOrderClosed = sdkerrors.New(ModuleName, errCodeOrderClosed, "order closed")
+	// ErrOrderExists indicates a new order was proposed overwrite the existing store key
+	ErrOrderExists = sdkerrors.New(ModuleName, errCodeOrderExists, "order already exists in store")
+	// ErrOrderTooEarly to match bid
+	ErrOrderTooEarly = sdkerrors.New(ModuleName, errCodeOrderTooEarly, "order: chain height to low for bidding")
+	// ErrOrderDurationExceeded order should be closed
+	ErrOrderDurationExceeded = sdkerrors.New(ModuleName, errCodeOrderDurationExceeded, "order duration has exceeded the bidding duration")
```
Ropes, line 79: Error naming needs revisit..

x/market/keeper/keeper.go
```diff
@@ -52,14 +52,18 @@ func (k Keeper) CreateOrder(ctx sdk.Context, gid dtypes.GroupID, spec dtypes.Gro
...
 		Spec:    spec,
 		State:   types.OrderOpen,
-		StartAt: ctx.BlockHeight() + orderTTL, // TODO: check overflow
+		StartAt: ctx.BlockHeight() + orderTTL,                         // TODO: check overflow
+		CloseAt: ctx.BlockHeight() + orderTTL + spec.OrderBidDuration, // TODO: check overflow, set via parameter
 	}
 
 	key := orderKey(order.ID())
+	// check not overwrite
+	expNil := store.Get(key)
```
boz, line 61: `if store.Has(key)`

x/market/types/errors.go
```diff
@@ -68,4 +71,10 @@ var (
 	ErrOrderMatched = sdkerrors.New(ModuleName, errCodeOrderMatched, "order matched")
 	// ErrOrderClosed order closed
 	ErrOrderClosed = sdkerrors.New(ModuleName, errCodeOrderClosed, "order closed")
+	// ErrOrderExists indicates a new order was proposed overwrite the existing store key
+	ErrOrderExists = sdkerrors.New(ModuleName, errCodeOrderExists, "order already exists in store")
+	// ErrOrderTooEarly to match bid
+	ErrOrderTooEarly = sdkerrors.New(ModuleName, errCodeOrderTooEarly, "order: chain height to low for bidding")
```
boz, line 77: it's not an error to bid before matching begins.

Ropes: Well this raises the debate on using error types as signals for responses to functionality. However `ValidateCanMatch() error` is using that pattern(from before this change). [Returning an error](https://github.com/ovrclk/akash/blob/bd1c6bf74e89d9227ad8879c27df09d031ef12d5/x/market/handler/endblock.go#L105) to indicate skipping this Order in the potential for OnEndBlock Order bidding loop. 

I agree it would be cleaner to signal skipping the Order with a bool/signal struct rather than error value. Shall we start to shift to that pattern?

x/market/types/types_test.go
```diff
@@ -0,0 +1,120 @@
...
+				State:   types.OrderOpen,
+				StartAt: baseHeight,
+				CloseAt: baseHeight + boostHeight,
+			},
+			height: int64(90),
+			expErr: types.ErrOrderTooEarly,
+		},
+	}
+	for _, test := range tests {
+		t.Logf("%s", test.desc)
```
boz, line 112: :hankey:

x/deployment/types/types.go
```diff
@@ -54,9 +54,10 @@ const (
 
 // GroupSpec stores group specifications
 type GroupSpec struct {
-	Name         string          `json:"name"`
-	Requirements []sdk.Attribute `json:"requirements"`
-	Resources    []Resource      `json:"resources"`
+	Name             string          `json:"name"`
+	Requirements     []sdk.Attribute `json:"requirements"`
+	Resources        []Resource      `json:"resources"`
+	OrderBidDuration int64           `json:"bid-duration"`
```
boz, line 60: add validation to `ValidateBasic()` - it should not be less than... I don't know... `1`.

x/market/types/types.go
```diff
@@ -85,10 +89,24 @@ func (o Order) ValidateCanMatch(height int64) error {
 		return err
 	}
 
-	if o.StartAt > height {
-		return errors.Errorf("too early to match order (%v > %v)", o.StartAt, height)
+	switch {
+	case height < o.StartAt:
+		return errors.Wrap(ErrOrderTooEarly, fmt.Sprintf("(%v > %v)", o.StartAt, height))
+	case height > o.CloseAt:
+		if o.CloseAt == int64(0) { // zero-value CloseAt value; invalid if unset
```
boz, line 96: don't worry about backwards-compatibility right now.

please remember to return early for simple cases.

x/market/types/errors.go
boz: i see.  it's ok, forget it.

x/market/types/types_test.go
Ropes: I need to make a regex sublinter for this.. sorry.

x/market/types/types.go
Ropes: removed compatibility case, short circuited normal case

x/market/types/types_test.go
Ropes: removed all `t.Logf()` from repository.

x/market/keeper/keeper.go
```diff
@@ -52,14 +52,16 @@ func (k Keeper) CreateOrder(ctx sdk.Context, gid dtypes.GroupID, spec dtypes.Gro
 		OrderID: types.MakeOrderID(gid, oseq),
 		Spec:    spec,
 		State:   types.OrderOpen,
-		StartAt: ctx.BlockHeight() + orderTTL, // TODO: check overflow
+		StartAt: ctx.BlockHeight() + orderTTL,                         // TODO: check overflow
+		CloseAt: ctx.BlockHeight() + orderTTL + spec.OrderBidDuration, // TODO: check overflow, set via parameter
```
boz, line 56: This is a critical issue now because `spec.OrderBidDuration` is supplied by the user instead of being a constant (`5`).

x/market/types/types.go
```diff
@@ -85,10 +88,17 @@ func (o Order) ValidateCanMatch(height int64) error {
...
-		return errors.Errorf("too early to match order (%v > %v)", o.StartAt, height)
+	switch {
+	case height > o.StartAt && height < o.CloseAt: // valid to match short circuit
+		return nil
+	case height < o.StartAt:
+		return errors.Wrap(ErrOrderTooEarly, fmt.Sprintf("(%v > %v)", o.StartAt, height))
+	case height > o.CloseAt:
+		// Close Open Order if it have surpassed the CloseAt block height.
+		return ErrOrderDurationExceeded
+	default:
```
boz, line 99: dude.  no.

if this was about the "return early" comment from a previous revision, i was talking about taking care of simple cases first instead of having a bunch of nesting.

```go
if x == 0 {
  if y > 0 {
    foo()
    return x
  }
  return y
}
return z
```
is better to write as

```go
if x != 0 {
  return z
}
if y == 0 {
  return y
}
foo()
return x
```

PR #751, opened 2020-07-21
feat: 604 Open Orders Index/Filter

Ropes:
To avoid iterating over all opened orders on each market module's
EndBlock method; added an index to Orders which will filter to only the
ones in Open state.

* Add/Created Open Orders to the new Index
* Update Open Order index key
  * Single method to handle all state changes
* Testing for OnEndBlock -> matchOrders()
  * Extracted @troian's suiteTest structure into importable struct to
  facilitate full chain/block testing.
* BidID String()'er fix

refs: #604


x/market/handler/endblock_external_test.go
```diff
@@ -0,0 +1,131 @@
...
+		require.NoError(t, err)
+	})
+
+	t.Run("assert still open orders", func(t *testing.T) {
+		count := 0
+		suite.MarketKeeper().WithOpenOrders(suite.Context(), func(o types.Order) bool {
+			count++
+			return false
+		})
+		assert.Len(t, openGroups, count)
```
boz, line 100: :nail_care: error message might be a bit off.  `assert.Equal(t,len(openGroups),count)` -- done in lease count assertion -- conveys what you're testing a bit more IMHO.

x/market/keeper/keeper_test.go
```diff
@@ -54,6 +56,19 @@ func Test_WithOrders(t *testing.T) {
...
 
 	assert.Equal(t, 1, count)
+
+	t.Run("open orders", func(t *testing.T) {
+		openCount := 0
+		keeper.WithOpenOrders(ctx, func(result types.Order) bool {
+			if assert.Equal(t, order.ID(), result.ID()) {
+				openCount++
+			} else {
+				t.Logf("different order ID??: %q", result.ID())
```
boz, line 66: :question: this will already be logged by `assert.Equal`.

x/market/types/id.go
```diff
@@ -101,7 +101,7 @@ func (id BidID) OrderID() OrderID {
 
 // String method for consistent output.
 func (id BidID) String() string {
-	return sdkutil.FmtBlockID(&id.Owner, &id.DSeq, &id.GSeq, &id.OSeq, nil)
+	return sdkutil.FmtBlockID(&id.Owner, &id.DSeq, &id.GSeq, &id.OSeq, &id.Provider)
```
boz, line 104: did this not affect any tests?  seems like `parse(id.String()).Provider == id.Provider` should have failed.

testutil/state/suite.go
```diff
@@ -0,0 +1,104 @@
...
+
+// ProviderKeeper key store
+func (ts *TestSuite) ProviderKeeper() pkeeper.Keeper {
+	return ts.pkeeper
+}
+
+// BankKeeper key store
+func (ts *TestSuite) BankKeeper() bank.Keeper {
+	return ts.bkeeper
+}
```
boz, line 104: :+1: 

can it be re-used in existing tests?

x/market/handler/endblock_external_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/ovrclk/akash/testutil"
+	"github.com/ovrclk/akash/testutil/state"
+	"github.com/ovrclk/akash/x/market/handler"
+	"github.com/ovrclk/akash/x/market/types"
+)
+
+func TestMatchOrders(t *testing.T) {
+	codec := app.MakeCodec()
+	suite := state.SetupTestSuite(t, codec)
+	t.Logf("block height: %d", suite.Context().BlockHeight())
```
boz, line 19: Remove debug logging before opening a PR.  I'm tired of having to point this out.

x/market/types/id.go
Ropes: I'm a little confused about that too, tests against strings weren't 100% though.

testutil/sdk.go
```diff
@@ -7,6 +7,14 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-func Coin(_ testing.TB) sdk.Coin {
+func Coin(t testing.TB) sdk.Coin {
+	t.Helper()
```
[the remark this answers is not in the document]
Ropes: I cargo-culted it for consistency with all of the other small test type builder functions which call `t.Helper()`, and also don't have any `t` assertions. https://github.com/ovrclk/akash/blob/26e321d3d2e282a437b382d4b2fc725b8a3e5ef2/testutil/ids.go#L25-L66

After looking at the `t.Helper()` docs and discussion, I agree it doesn't add much value here. Shall we just decide to remove it from most of `testutil` package then?

x/market/handler/endblock_external_test.go
Ropes: Good point!

testutil/state/suite.go
Ropes: For the most part yes! However in the case of tests within the `handler` package, it can't because of the `app` package import. That causes an import cycle.

So it can replace testing of package's public API, but for internal methods(such as `matchOrders()` directly) it has to be a local copy-pasta.

testutil/sdk.go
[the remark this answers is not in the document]
boz: `DeploymentID` passes it into `AccAddress`.  It's fine to leave as an argument to make a testutil functions consistent IMHO.

testutil/state/suite.go
```diff
@@ -0,0 +1,104 @@
...
+	ms      sdk.CommitMultiStore
+	ctx     sdk.Context
+	mkeeper keeper.Keeper
+	dkeeper dkeeper.Keeper
+	pkeeper pkeeper.Keeper
+	bkeeper bank.Keeper
+}
+
+// T returns the testing.T struct of the suite.
+func (ts *TestSuite) T() *testing.T {
```
boz, line 36: can you just return `testing.TB`?

x/market/types/id.go
boz: The tests were supposed to have a "`decode(encode(x)) == x`" test for all types.  They clearly don't or are broken in some other way.  Fix the test for this type and look through the tests for the other types for similar problems.

testutil/state/suite.go
Ropes: I agree that'd be preferable for consistency, however this is necessary because `assert` package only accepts `testing.T` :disappointed:

x/market/types/id.go
Ropes: created an issue to track: https://github.com/ovrclk/akash/issues/752

testutil/state/suite.go
boz: which assert package?  `stretchr/testify/{assert,require}` both work with `testing.TB`.  `git grep require -- testutil`

x/market/types/id.go
boz: It's because you're not testing `id.String()` [here](https://github.com/ovrclk/akash/blob/26e321d3d2e282a437b382d4b2fc725b8a3e5ef2/sdkutil/object_id_test.go#L231)

Ropes: great catch, updated test and referenced new issue to take a refreshed look at it.

PR #756, opened 2020-07-23
WIP feat: implement querying pagination

### LINKED ISSUE #726: querying: add pagination
Add pagination to all query list requests and responses.

server/node side:
* [ ] implement in `x/*/query`.
* [ ] server: define a max page size.
* [ ] accept `after` as bool (default: `false`) and `page` (default: `1`).

client side:
* [ ] use common flag utils for all `x/*/client/cli` packages.

troian:
fixes #726

### DIFF
--- a/x/deployment/query/querier.go
+++ b/x/deployment/query/querier.go
@@ -3,10 +3,15 @@ package query
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	abci "github.com/tendermint/tendermint/abci/types"
+
 	"github.com/ovrclk/akash/sdkutil"
 	"github.com/ovrclk/akash/x/deployment/keeper"
 	"github.com/ovrclk/akash/x/deployment/types"
-	abci "github.com/tendermint/tendermint/abci/types"
+)
+
+const (
+	queryLimit = 10
 )
 
 // NewQuerier creates and returns a new deployment querier instance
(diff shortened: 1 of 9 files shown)


x/deployment/query/querier.go
```diff
@@ -24,31 +29,46 @@ func NewQuerier(keeper keeper.Keeper) sdk.Querier {
...
+
+	if !filters.After {
+		k.WithDeployments(ctx, func(deployment types.Deployment) bool {
+			if filters.Accept(deployment, isValidState) {
+				value := Deployment{
+					Deployment: deployment,
+					Groups:     k.GetGroups(ctx, deployment.ID()),
+				}
+				values = append(values, value)
+			}
```
boz, line 49: I think we want to use filters in both cases.  Ideally we wouldn't have to do this same chunk of code twice.

x/deployment/query/querier.go
```diff
@@ -24,31 +29,46 @@ func NewQuerier(keeper keeper.Keeper) sdk.Querier {
...
+			return false
+		})
+	} else {
+		limit := queryLimit
+		k.WithDeploymentsAfter(ctx, keeper.OwnerKey(filters.Owner), func(deployment types.Deployment) bool {
 			value := Deployment{
 				Deployment: deployment,
-				Groups:     keeper.GetGroups(ctx, deployment.ID()),
+				Groups:     k.GetGroups(ctx, deployment.ID()),
```
boz, line 57: :roll_eyes:

x/types/types.go
```diff
@@ -0,0 +1,19 @@
...
+	cflags "github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+const (
+	flagAfter = "after"
+)
+
+func AddPaginationFlags(flags *pflag.FlagSet) {
```
boz, line 13: :+1:

Issue #752, opened 2020-07-21
Stringer testing improvements and re-evaluation


Ropes:
Work on #604 exposed that the BidID `String()`'er method was not correctly formatting all of the struct's fields. This issue is to re-analyze all of the Stringer methods on types, add more tests to assert that `decode(encode(x)) == <T>ID` string values map back to original type as expected.

Originating discussion: https://github.com/ovrclk/akash/pull/751#discussion_r458325500

Original issue: https://github.com/ovrclk/akash/issues/727

PR #758, opened 2020-07-23
fix:752 Stringer Test Assertions

Ropes:
Improved testing around formatting *ID types to their string
representation, and parsing string data back into original type.
(consume generated dog food)

Tests now run full circle from BlockID.String(), and reflected
BlockID.String() matches the original IDs.String().

fixes: #752

Played around with default `%s` formatting of data types; eg `types.Deployment`. As is currently an reference to the wrapping type of an *ID type, will return the stringer of the ID type, which at least does provide clarity into what ID is being referenced. It would be nice to come up with a consistent String() convention for types with `state`, version, prices.

### DIFF
--- a/sdkutil/object_id_test.go
+++ b/sdkutil/object_id_test.go
@@ -133,7 +133,7 @@ func TestBidID(t *testing.T) {
 	block := sdkutil.ReflectBlockID(bid)
 
 	str := block.String()
-	block, err := sdkutil.ParseBlockID(str)
+	sdkBlock, err := sdkutil.ParseBlockID(str)
 	assert.NoError(t, err)
 
 	assert.Equal(t, bid.Owner, *block.Owner)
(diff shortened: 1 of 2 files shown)


sdkutil/object_id_test.go
```diff
@@ -156,86 +156,112 @@ func TestGroupIDString(t *testing.T) {
 func TestBlockReflectionID(t *testing.T) {
 	t.Run("deployment-id", func(t *testing.T) {
 		id := testutil.DeploymentID(t)
-		block := sdkutil.ReflectBlockID(id)
 
-		assert.Equal(t, id.Owner, *block.Owner)
-		assert.Equal(t, id.DSeq, *block.DSeq)
+		t.Run("reflect", func(t *testing.T) {
+			block := sdkutil.ReflectBlockID(id)
```
troian, line 161: result of `sdkutil.ReflectBlockID` is always not nil. Does it make sense to return result as value? Will help a bit for escape analysis

Ropes: Done! Tests updated.

Issue #755, opened 2020-07-23
Protobuf migration


anilCSE:
# Upgrade SDK to 0.40 and use protobuf encoding for state, tx, CLI and querier.

## Bump SDK version to 0.40
Waiting for the release/rc. Use master for now

## State Encoding
As detailed in [ADR#19](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-019-protobuf-state-encoding.md), migrate state encoding to use proto encoding. Upstream tracking PR for reference: https://github.com/cosmos/cosmos-sdk/issues/5444
- [x] x/deployment (#767)
- [x] x/market (#766)
- [x] x/provider (#768)

## Query Encoding
Reference: [ADR#21](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-021-protobuf-query-encoding.md) and https://github.com/cosmos/cosmos-sdk/issues/5921

Following the specification outlined in [ADR#21](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-021-protobuf-query-encoding.md), for each module:
- Implement a Query gRPC service definition along with google.api.http REST routes in types.proto
- Implement the QueryServer and wire it up to the app module
Use efficient pagination as described in https://github.com/cosmos/cosmos-sdk/issues/6191

### CLI query commands
- [ ] x/deployment
- [ ] x/market
- [ ] x/provider
### Query services
- [x] x/deployment (#759)
- [x] x/market (#760)
- [x] x/provider (#761)
### Remove/Deprecate Rest querier (?)
- [ ] x/deployment
- [ ] x/market
- [ ] x/provider

## Transaction Encoding
Reference: [ADR#20](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-020-protobuf-transaction-encoding.md) and https://github.com/cosmos/cosmos-sdk/issues/6213

PR #760, opened 2020-07-24
WIP: x/market: grpc query service

akhilkumarpilli:
ref: #755


x/market/keeper/grpc_query.go
```diff
@@ -0,0 +1,211 @@
...
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkquery "github.com/cosmos/cosmos-sdk/types/query"
+	"github.com/ovrclk/akash/x/market/types"
+)
+
+// Querier is used as Keeper will have duplicate methods if used directly, and gRPC names take precedence over keeper
+type Querier struct {
+	Keeper
+}
```
boz, line 18: Can you make `Querier` an interface, with a private struct implementation for all of these?

I'd like to do the same for `Keeper` as well.

x/market/types/types.go
```diff
@@ -100,28 +122,58 @@ func (o Order) validateMatchableState() error {
...
+
+	// Checking oseq filter
+	if filters.OSeq != 0 && filters.OSeq != obj.OrderID.OSeq {
+		return false
+	}
+
+	// Checking state filter
+	if filters.State != 0 && filters.State != obj.State {
+		return false
+	}
```
boz, line 150: :pray: thank you

x/market/types/types.go
```diff
@@ -164,3 +270,57 @@ var LeaseStateMap = map[string]LeaseState{
...
+	var out string
+	for _, order := range l {
+		out += order.String() + "\n"
+	}
+
+	return strings.TrimSpace(out)
+}
+
+// Accept returns whether lease filters valid or not
+func (filters LeaseFilters) Accept(obj Lease) bool {
```
boz, line 294: OT: It is reasonable to put these here, but they are pretty independent and I'd like to move them out of `types` at some point.

PR #761, opened 2020-07-24
WIP: x/provider: grpc query service

akhilkumarpilli:
ref: #755


proto/akash/market/lease.proto
```diff
@@ -0,0 +1,62 @@
...
+message Lease {
+    option (gogoproto.equal) = false;
+
+    LeaseID lease_id = 1 [
+        (gogoproto.customname) = "LeaseID",
+        (gogoproto.nullable)   = false,
+        (gogoproto.jsontag)    = "id",
+        (gogoproto.moretags)   = "yaml:\"id\""
+    ];
+    uint32 state = 2 [
```
boz, line 51: can we use an enum for these?

cmd/akashd/genaccounts.go
```diff
@@ -113,45 +114,57 @@ contain valid denominations. Accounts may optionally be supplied with vesting pa
...
-				return errors.Wrap(err, "failed to unmarshal genesis state")
+				return fmt.Errorf("failed to unmarshal genesis state: %w", err)
 			}
 
-			authGenState := auth.GetGenesisStateFromAppState(cdc, appState)
+			authGenState := authtypes.GetGenesisStateFromAppState(cdc, appState)
 
 			if authGenState.Accounts.Contains(addr) {
-				return errors.Errorf("cannot add account at existing address %s", addr)
+				return fmt.Errorf("cannot add account at existing address %s", addr)
```
anilCSE, line 125: can we use `sdkerrors` throughout for custom errors?

proto/akash/market/lease.proto
anilCSE: Yes, makes sense. Let's use enum for these

Makefile
```diff
@@ -244,4 +244,56 @@ test-sim-after-import:
 	go test -mod=readonly -tags=$(BUILD_MAINNET) $(APP_DIR) -run=TestAppSimulationAfterImport -Enabled=true \
 		-NumBlocks=50 -BlockSize=100 -Commit=true -Seed=99 -Period=5 -v -timeout 10m
 
-test-sims: test-sim-fullapp test-sim-nondeterminism test-sim-import-export test-sim-after-import
\ No newline at end of file
+test-sims: test-sim-fullapp test-sim-nondeterminism test-sim-import-export test-sim-after-import
+
+proto-gen:
+	@./script/protocgen.sh
```
boz, line 250: extremely small nitpick: I prefer to see the command that is executing; do you mind not using `@`?

PR #763, opened 2020-07-28
Base Proto PR with sdk proto files

akhilkumarpilli:


anilCSE:
@akhilkumarpilli let's fix simapp tests before merging this. wdyt?

anilCSE:
> @akhilkumarpilli let's fix simapp tests before merging this. wdyt?

Oh, there are a lot of dependencies to make this work. May be we can fix these simapp tests onto master

anilCSE:
@boz I suggest to merge this PR with tests failing, otherwise we would need to address a lot stuff in single PR itself. WDYT? P.S: This is not onto `master`

boz:
@anilCSE agreed.  definitely go for it.

Issue #762, opened 2020-07-25
Provider Updates Deployments on Version change


Ropes:
Now that the Deployment's Manifest version is tracked(#565 : #748); the Provider service needs to be updated to actually update their leased Groups with the new Deployment spec. Only a few attributes of the SDL Manifest can be updated, but when a new Version is set on the chain, the Provider needs to await manifest changes to be sent.

### Provider changes needed(WIP)
* ManifestService needs to be updated to properly [handleUpdate](https://github.com/ovrclk/akash/blob/de2bca2ea03a8129c85020e4f8f44636c3851b24/provider/manifest/manager.go#L107) to use the Version field.
  * Await for Owner to send new SDL.
  * Update k8s with allowed changes from validated SDL.
  * Downstream message to inform of successful update?

(Still investigating provider codebase, will add implementation details)

Issue #725, opened 2020-07-01
`akashctl`: audit query filter params


boz:
For all of the `akashctl query [deployment|market|provider] query * list` commands, ensure that we can filter by all of the fields of the listed object's ID.

PR #765, opened 2020-07-28
feat:762 Deployment Version Updating validation

Ropes:
* ~~Performance improvements to Active Lease lookup~~
* Code Documentation in the `provider` codebase
* Validation of manifest version to blockchain Deployment version
  * Asserts that manifest object provided is correct.
* ManifestVersion() method to extract Version from Manifest type.
* Makefile `deployment-update` command added
* Document Deployment update and (re)sending manifest.

fixes: #762

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -45,7 +45,7 @@ type Client interface {
 	Tx() TxClient
 }
 
-// NewClient creates new client instance
+// NewClient creates new client instance to interface with terndermint.
 func NewClient(
 	log log.Logger,
 	cctx ccontext.CLIContext,
--- a/provider/cluster/service.go
+++ b/provider/cluster/service.go
@@ -212,15 +212,13 @@ loop:
 			}
 
 		case dm := <-s.managerch:
-
 			s.log.Debug("manager done", "lease", dm.lease)
 
+			// unreserve resources
 			if _, err := s.inventory.unreserve(dm.lease.OrderID(), dm.mgroup); err != nil {
 				s.log.Error("unreserving inventory", "err", err,
 					"lease", dm.lease, "group-name", dm.mgroup.Name)
 			}
-
-			// todo: unreserve resources
 
 			delete(s.managers, mquery.LeasePath(dm.lease))
 		}
--- a/provider/manifest/manager.go
+++ b/provider/manifest/manager.go
@@ -1,6 +1,7 @@
 package manifest
 
 import (
+	"bytes"
 	"context"
 	"time"
 
@@ -12,6 +13,7 @@ import (
 	"github.com/ovrclk/akash/provider/event"
 	"github.com/ovrclk/akash/provider/session"
 	"github.com/ovrclk/akash/pubsub"
+	"github.com/ovrclk/akash/sdl"
 	"github.com/ovrclk/akash/util/runner"
 	"github.com/ovrclk/akash/validation"
 	dquery "github.com/ovrclk/akash/x/deployment/query"
@@ -20,7 +22,11 @@ import (
 )
 
 var (
+	// ErrShutdownTimerExpired for a terminating deployment
 	ErrShutdownTimerExpired = errors.New("shutdown timer expired")
+	// ErrManifestVersion indicates that the given manifest's version does not
+	// match the blockchain Version value.
+	ErrManifestVersion = errors.New("manifest version validation failed")
 )
 
 func newManager(h *service, daddr dtypes.DeploymentID) (*manager, error) {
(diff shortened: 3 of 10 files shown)


Ropes:
Codecov isn't happy because no tests were added to the small-ish piece which does the validation check. 

Question to discuss: how should we test provider functionality? `integration/` tests don't go far enough to actually validate the provider's executing functionality. Extending the `TestSuite` would enable sharring the data stores between separate components.

client/extra.go
```diff
@@ -6,14 +6,22 @@ import (
...
 
-// TODO: implement with search parameters
-
+// ActiveLeasesForProvider filter the Lease search to the provider's ID AccAddress
+// paramenter and 'active' state.
 func (c *qclient) ActiveLeasesForProvider(id sdk.AccAddress) (mquery.Leases, error) {
-	leases, err := c.Leases(mquery.LeaseFilters{})
+	// Establish Provider's search
+	lf := mquery.LeaseFilters{
+		Owner:        id,
```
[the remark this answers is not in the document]
Ropes: :grimacing: yeah... so this is the filter struct:
```go
// LeaseFilters defines flags for lease list filter
type LeaseFilters struct {
	Owner sdk.AccAddress
	// State flag value given
	StateFlagVal string
	// Actual state value decoded from LeaseStateMap
	State types.LeaseState
}
```

After diving into query code, looks like we don't have the actual `Provider` address filtering functionality built in yet. However I'm also rather unsure how this raw ABCI query is actually being interpreted. I'd like to get some tests around it.. however if we want to get the version fix out, I reverted this change and we can get the Version update out.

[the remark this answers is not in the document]
boz: right.  it's not ideal.  it shouldn't be using a string value, etc...

all of the ids are additive and the `Owner` field comes from the `DeploymentID`

client/client.go
```diff
@@ -31,7 +31,7 @@ type QueryClient interface {
 	pquery.Client
 
 	// TODO: implement with search parameters
-	ActiveLeasesForProvider(id sdk.AccAddress) (mquery.Leases, error)
+	ActiveLeasesForProvider(provider sdk.AccAddress) (mquery.Leases, error)
```
boz, line 34: eh

provider/manifest/manager.go
```diff
@@ -301,7 +308,16 @@ func (m *manager) validateRequests() {
 }
 
 func (m *manager) validateRequest(req manifestRequest) error {
-	// TODO: hash(manifest) == m.data.Version
+	// ensure that an uploaded manifest matches the hash declared on
+	// the Akash Deployment.Version
+	manifestVersion, err := sdl.ManifestVersion(req.value.Manifest)
```
boz, line 313: :nail_care: too verbose.  `version` is fine.

client/extra.go
```diff
@@ -6,17 +6,22 @@ import (
...
+	leases, err := c.Leases(lf)
 	if err != nil {
 		return nil, err
 	}
 
+	// TODO: Secondary filtering *should* not be necessary
 	var filtered mquery.Leases
 	for _, lease := range leases {
-		if lease.Provider.Equals(id) && lease.State == types.LeaseActive {
+		if lease.Provider.Equals(provider) && lease.State == types.LeaseActive {
```
boz: just leave this stuff as it was - these changes don't really add anything.

/bump #725 #604

provider/cluster/service.go
```diff
@@ -212,16 +212,14 @@ loop:
...
-
 			s.log.Debug("manager done", "lease", dm.lease)
 
+			// unreserve resources
 			if _, err := s.inventory.unreserve(dm.lease.OrderID(), dm.mgroup); err != nil {
 				s.log.Error("unreserving inventory", "err", err,
 					"lease", dm.lease, "group-name", dm.mgroup.Name)
 			}
 
-			// todo: unreserve resources
```
boz, line 223: :+1:

client/extra.go
```diff
@@ -6,14 +6,19 @@ import (
...
+	// NOTE: Currently unable to filter by Provider in tendermint scan
+	lf := mquery.LeaseFilters{
+		// TODO: Add filters when tests prove functionality
+	}
+	leases, err := c.Leases(lf)
 	if err != nil {
 		return nil, err
 	}
 
+	// TODO: Secondary filtering *should* not be necessary
```
boz, line 21: Just leave this stuff as it was before - these changes don't really add anything.

refs #604 #725

Ropes: :heavy_check_mark:

Issue #777, opened 2020-08-04
testing: E2E validation of app run by Provider


Ropes:
**Goal**: Single test execution which validates that our all components of the Akash stack are functioning correctly and running an example App is successful. To detect when a change causes a break in E2E functionality; every component should provide useful logging to make the cause of the failure easily identifiable between each service. 

## Integration Testing

Currently we have basic blockchain testing of `akashctl` and `akashd` in the `integration/` package. This ~~could theoretically be~~ has been extended to test running the `provider`, however to provide full E2E running App validation some minified k8s API would need to support the provider process.(currently running against the same `kind` env that we use in `_run/kube`).

## Service Testing

Testing via the indipendent services orchestrated via containers or bash scripts. Ultimately a close approximation to what will be running, and a true test of the services. The downside is that it requires a full orchestration system which also requires maintenance and sorting through the processes output can waste an incredible amount of time.

## Integration Testing Implemented

Thus far work done on expanding the capability of our `integration/` testing package. 

* `provider` is now able to run with equivalent configuration to how it's handled in `_run/kube`
  * Re-using the `integration` package's Deployment test code, the `provider` is bidding on Orders.
* Had to hack up the `cosmos-sdk/tests` package to support processes where EnvVars needed to be set.
  * Since all this code is going to need a refactor, I've done a quick copy-hack in the `integration/` package.
* Various improvements to cleanup and common `test_helpers.go`

### TODO:
* [ ] Manage `kind` environment for integration tests. Potentially with `integration` package helpers.
* [ ] Send manifest to Provider
* [ ] Assert that application is deployed
* [ ] Assert that the application is available and responds
* [ ] Teardown

boz:
There are existing `kind` setup scripts, used [here](https://github.com/ovrclk/akash/blob/faf26a8704ecf39730597254b9633d3da6542ca7/.github/workflows/k8s-integration.yaml#L16-L20) in CI and also in `_run/single`, `_run/kube`.

Ropes:
Hopefully will reuse `_run/kube` and the CI tooling!

As is right now I've got the `integration` package bootstrapping the Provider, creating Deployments, validating won Leases, and sending manifest to the provider which does create the `web` container. However running into issues with the spec, or tied to the issue we have with how `kind` got created.(Still de-:bug: 'ing)

```                                                                                                            
[...]

PR #780, opened 2020-08-07
feat: E2E provider integration test

Ropes:
Functional integration test of the Provider running against a KinD cluster locally or in GH Actions. Uses previous `integration` test functionality of bootstraping a blockchain, data, keys, and runs an Akash node. A KinD k8s cluster is created, and a `provider` service is connected to it and run.

A tenant creates a Akash demo-app Deployment, and sends the manifest to Provider. The Provider manages the deployment and runs it in KinD. The integration test then requests the static web page from the k8s external IP:Port and validates that the App is running.
    
fixes: #777

### DIFF
--- a/.github/workflows/k8s-integration.yaml
+++ b/.github/workflows/k8s-integration.yaml
@@ -10,13 +10,24 @@ on:
       - provider/cluster/kube/*
 
 jobs:
-  crd:
+  crd-e2e:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
       - uses: engineerd/setup-kind@v0.4.0
-      - name: Setup K8S
-        run: ./script/setup-kind.sh
-      - name: Run Tests
+        with:
+            config: ./_run/kube/kind-config.yaml
+      - name: Setup Ingress K8S
+        run: make -s -C _run/kube kind-ingress-setup
+      - name: k8s-ingress
+        run: make -s -C _run/kube kind-k8s-ip
+      - name: Kube Environment
+        run: |
+          kubectl config view
+          kubectl cluster-info
+          kubectl get pods,ingress,svc -A
+      - name: Run E2E Tests
+        run: make test-e2e-integration
+      - name: Run K8s Tests
         run: make test-k8s-integration
--- a/.github/workflows/k8s-integration.yaml
+++ b/.github/workflows/k8s-integration.yaml
@@ -11,13 +11,17 @@ on:
 
 jobs:
   crd-e2e:
+    env:
+      KIND_NAME: kind
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
       - uses: engineerd/setup-kind@v0.4.0
         with:
-            config: ./_run/kube/kind-config.yaml
+          config: ./_run/kube/kind-config.yaml
+      - name: Docker Status
+        run: docker ps -a
       - name: Setup Ingress K8S
         run: make -s -C _run/kube kind-ingress-setup
       - name: k8s-ingress
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,9 @@ APP_DIR := ./app
 GO := GO111MODULE=on go
 GOBIN := $(shell go env GOPATH)/bin
 
-KIND_API_IP ?= $(shell make -sC _run/kube kind-k8s-ip)
-KIND_HTTP_PORT ?= $(shell make -sC _run/kube app-http-port)
-KIND_VARS ?= KIND_APP_PORT="$(KIND_HTTP_PORT)" KIND_K8S_IP="$(KIND_API_IP)"
+KIND_APP_IP ?= $(shell make -sC _run/kube kind-k8s-ip)
+KIND_APP_PORT ?= $(shell make -sC _run/kube app-http-port)
+KIND_VARS ?= KIND_APP_IP="$(KIND_APP_IP)" KIND_APP_PORT="$(KIND_APP_PORT)"
 
 # Setting mainnet flag based on env value
 # export MAINNET=true to set build tag mainnet
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,10 @@ APP_DIR := ./app
 GO := GO111MODULE=on go
 GOBIN := $(shell go env GOPATH)/bin
 
+KIND_API_IP ?= $(shell make -sC _run/kube kind-k8s-ip)
+KIND_HTTP_PORT ?= $(shell make -sC _run/kube app-http-port)
+KIND_VARS ?= KIND_APP_PORT="$(KIND_HTTP_PORT)" KIND_K8S_IP="$(KIND_API_IP)"
+
 # Setting mainnet flag based on env value
 # export MAINNET=true to set build tag mainnet
 ifeq ($(MAINNET),true)
@@ -150,6 +154,16 @@ devdeps-install:
 test-integration: $(BINS)
 	cp akashctl akashd ./_build
 	go test -mod=readonly -p 4 -tags "integration $(BUILD_MAINNET)" -v ./integration/...
+
+test-e2e-integration: $(BINS)
+	# ASSUMES:
+	# 1. cluster created - `kind create cluster --config=_run/kube/kind-config.yaml`
+	# 2. cluster setup - `make -s -C _run/kube kind-ingress-setup`
+	cp akashctl akashd ./_build
+	$(KIND_VARS) go test -mod=readonly -p 4 -tags "e2e integration $(BUILD_MAINNET)" -v ./integration/... -run TestE2EApp
+
+test-query-app: $(BINS)
+	 $(KIND_VARS) go test -mod=readonly -p 4 -tags "e2e integration $(BUILD_MAINNET)" -v ./integration/... -run TestQueryApp
 
 test-k8s-integration:
 	# ASSUMES:
--- a/_run/common-kind.mk
+++ b/_run/common-kind.mk
@@ -1,5 +1,8 @@
+# KIND_NAME NOTE: 'kind' string literal is default for the GH actions 
+# KinD, it's fine to use other names locally, however in GH container name 
+# is configured by engineerd/setup-kind. `kind-control-plane` is the docker
+# image's name in GH Actions.
 KIND_NAME      ?= $(shell basename $$PWD)
-# 'kind' matches the configuration of the GH actions KinD. 
 K8S_CONTEXT    ?= $(shell kubectl config current-context)
 KIND_HTTP_PORT  ?= $(shell docker inspect \
 										--type container "$(KIND_NAME)-control-plane" \
@@ -19,7 +22,7 @@ KIND_PORT_BINDINGS ?= $(shell docker inspect "$(KIND_NAME)-control-plane" \
 KIND_CONFIG       ?= kind-config.yaml
 
 PROVIDER_HOSTNAME ?= localhost
-PROVIDER_HOST     ?= $(PROVIDER_HOSTNAME):$(APP_HTTP_PORT)
+PROVIDER_HOST     ?= $(PROVIDER_HOSTNAME):$(KIND_HTTP_PORT)
 PROVIDER_ENDPOINT ?= http://$(PROVIDER_HOST)
 
 # TODO: referencing master/latest on a k8s repository is prone to breaking at some point.
--- a/_run/common-kind.mk
+++ b/_run/common-kind.mk
@@ -1,25 +1,52 @@
 KIND_NAME      ?= $(shell basename $$PWD)
+# 'kind' matches the configuration of the GH actions KinD. 
 K8S_CONTEXT    ?= $(shell kubectl config current-context)
-KIND_HTTP_PORT  = $(shell docker inspect \
+KIND_HTTP_PORT  ?= $(shell docker inspect \
 										--type container "$(KIND_NAME)-control-plane" \
-										--format '{{index .HostConfig.PortBindings "80/tcp" 0 "HostPort"}}')
+										--format '{{index .NetworkSettings.Ports "80/tcp" 0 "HostPort"}}')
+
+KIND_HTTP_IP  ?= $(shell docker inspect \
+										--type container "$(KIND_NAME)-control-plane" \
+										--format '{{index .NetworkSettings.Ports "80/tcp" 0 "HostIp"}}')
+
+KIND_K8S_IP ?= $(shell docker inspect \
+										--type container "$(KIND_NAME)-control-plane" \
+										--format '{{index .NetworkSettings.Ports "6443/tcp" 0 "HostIp"}}')
+
+KIND_PORT_BINDINGS ?= $(shell docker inspect "$(KIND_NAME)-control-plane" \
+										--format '{{index .NetworkSettings.Ports "80/tcp" 0 "HostPort"}}')
 
 KIND_CONFIG       ?= kind-config.yaml
 
 PROVIDER_HOSTNAME ?= localhost
-PROVIDER_HOST     ?= $(PROVIDER_HOSTNAME):$(KIND_HTTP_PORT)
+PROVIDER_HOST     ?= $(PROVIDER_HOSTNAME):$(APP_HTTP_PORT)
 PROVIDER_ENDPOINT ?= http://$(PROVIDER_HOST)
 
+# TODO: referencing master/latest on a k8s repository is prone to breaking at some point.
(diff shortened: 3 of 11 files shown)


Makefile
```diff
@@ -151,6 +151,13 @@ test-integration: $(BINS)
 	cp akashctl akashd ./_build
 	go test -mod=readonly -p 4 -tags "integration $(BUILD_MAINNET)" -v ./integration/...
 
+test-e2e-integration: $(BINS)
+	cp akashctl akashd ./_build
+	# ASSUMES:
+	# 1. cluster created - `kind create cluster --config=_run/kube/kind-config-80.yaml`
```
boz, line 157: we shouldn't be using port 80.

_run/kube/deployment.yaml
```diff
@@ -19,16 +19,14 @@ profiles:
       memory: 16Mi
       storage: 128Mi
   placement:
-    westcoast:
-      attributes:
-        region: us-west
+    global:
```
boz, line 22: was this from trying to fix something?

integration/e2e_test.go
```diff
@@ -0,0 +1,214 @@
...
+	host := "localhost"
+	ns := "lease" // this is an odd variable which by default is 'lease'
+	log := log.NewTMLogger(os.Stdout)
+	k8sClient, err := kube.NewClient(log, host, ns)
+	require.NoError(t, err)
+
+	// TODO: get the app's selected port from k8s Deployment info
+	k8sDeployments, err := k8sClient.Deployments(ctx)
+	require.NoError(t, err)
+	assert.NotEqual(t, k8sDeployments, 0)
```
boz, line 167: why wouldn't this fail all the time?

integration/e2e_test.go
```diff
@@ -0,0 +1,214 @@
...
+	k8sGroup := k8sDeployments[0].ManifestGroup()
+	for _, dep := range k8sDeployments {
+		t.Logf("deployment: %#v", dep)
+		for _, svc := range k8sGroup.Services {
+			t.Logf("service: %#v", svc)
+		}
+	}
```
boz, line 174: not testing anything?

fwiw the client is [already tested](https://github.com/ovrclk/akash/blob/faf26a8704ecf39730597254b9633d3da6542ca7/provider/cluster/kube/k8s_integration_test.go#L55-L62)

integration/e2e_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+// Assert provider launches app in kind cluster
+func queryApp(t *testing.T) {
+	connectStr := fmt.Sprintf("http://127.0.0.1/")
+	req, err := http.NewRequest("GET", connectStr, nil)
+	req.Host = "hello.localhost" // NOTE: cannot be inserted as a req.Header element, that is overwritten by this req.Host field.
+	req.Header.Add("Cache-Control", "no-cache")
+	req.Header.Add("Connection", "keep-alive")
+	req.Header.Add("Accept", "text/html,application/xhtml+xml,*/*")
+	req.Header.Add("User-Agent", "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.105 Safari/537.36")
```
boz, line 185: :scissors:

integration/gobash.go
```diff
@@ -0,0 +1,116 @@
+// +build integration
```
boz, line 1: can you put this and `process.go` in a separate package?  `integration/exec` or something.

integration/test_helpers.go
```diff
@@ -83,6 +97,12 @@ func NewFixtures(t *testing.T) *Fixtures {
 	tmpDir, err := ioutil.TempDir("", "akash_integration_"+t.Name()+"_")
 	require.NoError(t, err)
 
+	// Prevent akashd errors on exit due to data saving behavior.
+	tmpStat, err := os.Lstat(tmpDir)
+	require.NoError(t, err)
+	err = os.MkdirAll(fmt.Sprintf("%s/.akashd/data/cs.wal", tmpDir), tmpStat.Mode())
```
boz, line 103: shouldn't `cs.wal` be a file?

integration/test_helpers.go
```diff
@@ -156,6 +177,7 @@ func InitFixtures(t *testing.T) (f *Fixtures) {
 	// start an account with tokens
 	f.AddGenesisAccount(f.KeyAddress(keyFoo), startCoins())
 	f.AddGenesisAccount(f.KeyAddress(keyBar), startCoins())
+	f.AddGenesisAccount(f.KeyAddress(keyBaz), startCoins())
```
boz, line 180: i read the last argument as :hankey: coins

_run/kube/deploy.yaml
```diff
@@ -0,0 +1,670 @@
+
```
boz, line 1: :scissors:

integration/test_helpers.go
Ropes: I saw errors trying to write beneath it as a directory, thus mkdir call.

_run/kube/deployment.yaml
Ropes: Enables the deployment to match against the attributes of the Provider being run in the integration test.

.github/workflows/k8s-integration.yaml
```diff
@@ -10,13 +10,30 @@ on:
...
         run: make test-k8s-integration
+      - name: Testing
+        run: |
+          kubectl cluster-info
+          kubectl get pods -n kube-system
+          echo "current-context:" $(kubectl config current-context)
+          echo "environment-kubeconfig:" ${KUBECONFIG}
+          kubectl get pods -A
+          kubectl get ingress -A 
+          kubectl get svc -A
```
boz, line 39: If we're going to be printing environment info, can you make it clean & consistent?  For instance, them into one `run:` step, group similar things together, and use consistent techniques.

also this does four lines at once:
```
kubectl get pods,ingress,svc -A
```

Makefile
```diff
@@ -151,6 +154,21 @@ test-integration: $(BINS)
 	cp akashctl akashd ./_build
 	go test -mod=readonly -p 4 -tags "integration $(BUILD_MAINNET)" -v ./integration/...
 
+test-e2e-integration: export KIND_K8S_IP=$(KIND_API_IP)
+test-e2e-integration: export KIND_APP_PORT=$(KIND_HTTP_PORT)
```
boz, line 158: There are two things about this that I didn't know were possible.  I still don't get how the `export X=Y` dependency thing works.

Choose a method and use it consistently.  Either this magic or the more standard way of doing things:

```
KIND_VARS = FOO=x BAR=y
# ...

test-foo:
    $(KIND_VARS) go test foo
test-bar
    $(KIND_VARS) go test bar
```

_run/common-kind.mk
```diff
@@ -1,19 +1,41 @@
-KIND_NAME      ?= $(shell basename $$PWD)
+KIND_NAME      = kind
```
boz, line 1: can you set `KIND_NAME=kind` in the environment in the workflow yaml?

_run/common-kind.mk
```diff
@@ -1,19 +1,41 @@
...
+# 'kind' matches the configuration of the GH actions KinD. 
 K8S_CONTEXT    ?= $(shell kubectl config current-context)
-KIND_HTTP_PORT  = $(shell docker inspect \
+APP_HTTP_PORT  ?= $(shell docker inspect \
```
boz, line 4: why rename this one to `APP_` when the others are `KIND_`?  It is used elsewhere.

_run/kube/deployment-global.yaml
```diff
@@ -0,0 +1,32 @@
+---
+version: "1.5"
```
boz, line 2: I'd prefer to not have sdl files all over the place.  Either fix the provider attributes of the test so that this isn't needed or put this inside `integration/_testdata` or something.

integration/cosmostests/LICENSE
```diff
@@ -0,0 +1,204 @@
+Cosmos-SDK
+License: Apache2.0
```
boz, line 2: really think this is necessary?

integration/test_helpers.go
```diff
@@ -42,7 +49,15 @@ const (
 	feeStartValue        = 1000000
 	deploymentFilePath   = "./../x/deployment/testdata/deployment.yaml"
 	deploymentV2FilePath = "./../x/deployment/testdata/deployment-v2.yaml"
+	deploymentOvrclkApp  = "./../_run/kube/deployment-global.yaml"
 	providerFilePath     = "./../x/provider/testdata/provider.yaml"
+	providerTemplate     = `host: %s
+attributes:
+  - key: region
+    value: sfo
```
boz, line 57: this is what can be tweaked to match the original deployment yaml.

integration/e2e_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+	fooAcc := f.QueryAccount(fooAddr)
+	startTokens := sdk.TokensFromConsensusPower(denomStartValue)
+	require.Equal(t, startTokens, fooAcc.GetCoins().AmountOf(denom))
+
+	// address for provider to listen on
+	_, port, err := server.FreeTCPAddr()
+	if err != nil {
+		f.T.Fatal(err)
+	}
```
boz, line 54: `require.NoError(t,err)`

integration/e2e_test.go
```diff
@@ -0,0 +1,188 @@
...
+		DisableKeepAlives: false,
+	}
+	client := &http.Client{
+		Transport: tr,
+	}
+
+	// retry mechanism
+	var resp *http.Response
+	for i := 0; i < 5; i++ {
+		time.Sleep(10 * time.Second) // reduce absurdly long wait period
```
boz, line 170: yeah, wait 1 second, try 50 times.

integration/test_helpers.go
```diff
@@ -236,13 +259,63 @@ func (f *Fixtures) CollectGenTxs(flags ...string) {
...
+		f.Flags(),
+	)
+
+	_, stderr := tests.ExecuteT(f.T, addFlags(cmd, flags), "")
+	assert.Empty(f.T, stderr, "sendmanifest stderr")
+	tests.WaitForNextNBlocksTM(2, f.Port)
+}
+
+// K8sClient provides access to the k8s API
+func (f *Fixtures) K8sClient(host string) (cluster.Client, error) {
```
boz, line 313: unused?

.github/workflows/k8s-integration.yaml
Ropes: I'll combine it into one or two steps.

Makefile
Ropes: Yeah, I agree this is weird wizardry, the reason I switched it back to this is that the envvar is scoped to only the `test-e2e-integration` target. Cuts down on variables being defined globally, but only used in one place. 

Second reason is that when I prefixed the envvars inline to the command; `ENVV=foo go test...` the GH action's stage output never seemed to return testing output until the full duration of test, which added to the PITA feedback loop... So it doesn't matter when things are working, but increases debugging agony a bit.

I like your KIND_VARS prefix grouping better though, it makes adding new targets possible with less boilerplate.

_run/common-kind.mk
Ropes: There isn't a strait `name` param in the config :shrug: Kinda odd but seems to be intentionally a CLI flag.

_run/common-kind.mk
Ropes: good catch.

_run/kube/deployment-global.yaml
Ropes: Fixed matching requirements, removed.

integration/cosmostests/LICENSE
Ropes: When in doubt, it's typically safer to preserve and not potentially breach copyright licenses. 

I doubt the Cosmos team will take issue, but learning from the k8s community, not preserving Licenses correctly is a legal quagmire.

_run/common-kind.mk
boz: I mean [set an environment variable](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#env).  It is using `?=`; the right-hand side here is a default if `$KIND_NAME` isn't defined.

Ropes: Done!

boz: I want to keep the previous behavior - it defaults to the name of the current directory.

Ropes: Ahh, I see. Reset it.

The reason I changed it was there was a problem with the cluster being named `kube-control-plane` and messing up downstream queries. However Looks like the variable is getting applied where applicable so local testing shows it's working with $$PWD!

boz: You can override the make variable's default value with an environment variable.  This is what I've been talking about the whole time.

I don't know why the cluster name would be `kube-control-plane`, hopefully that was just a fluke.

boz: looks like it's still a problem @Ropes 

```Error: No such container: kube-control-plane
Error: No such container: kube-control-plane
# ASSUMES:
# 1. cluster created - `kind create cluster --config=_run/kube/kind-config.yaml`
# 2. cluster setup - `make -s -C _run/kube kind-ingress-setup`
cp akashctl akashd ./_build
KIND_APP_PORT="" KIND_K8S_IP="" go test -mod=readonly -p 4 -tags "e2e integration " -v ./integration/... -run TestE2EApp
```

Ropes: The naming that `engineerd/setup-kind` uses is rather peculiar. eg the actual k8s cluster name is `kind-kind` by default, and docker container name is `kind-control-plane`(locally `kube-control-plane`).
```
 Kube Environment0s
kube-system     service/metrics-server                       ClusterIP   10.96.40.80    <none>        443/TCP                      67s
Run kubectl config view
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: DATA+OMITTED
    server: https://127.0.0.1:32768
  name: kind-kind
contexts:
- context:
    cluster: kind-kind
    user: kind-kind
  name: kind-kind
current-context: kind-kind
```

Tests are now passing again by setting the job's env KIND_NAME to tell the Makefiles where to look.

_run/common-kind.mk
```diff
@@ -1,19 +1,41 @@
 KIND_NAME      ?= $(shell basename $$PWD)
+# NOTE: 'kind' matches the default configuration of the GH actions KinD, it's fine to use other names locally, however in GH container name is configured by engineerd/setup-kind. `kind-control-plane` is the docker image's name in GH Actions.
```
boz, line 2: nit for next time:

* put doc above the thing it's documenting
* wrap at 80 columns
* it's not really clear what this is talking about since `'kind'` isn't here anywhere.

_run/common-kind.mk
```diff
@@ -1,19 +1,41 @@
...
+										--format '{{index .NetworkSettings.Ports "6443/tcp" 0 "HostIp"}}')
+
+KIND_PORT_BINDINGS ?= $(shell docker inspect "$(KIND_NAME)-control-plane" \
+										--format '{{index .NetworkSettings.Ports "80/tcp" 0 "HostPort"}}')
 
 KIND_CONFIG       ?= kind-config.yaml
 
 PROVIDER_HOSTNAME ?= localhost
-PROVIDER_HOST     ?= $(PROVIDER_HOSTNAME):$(KIND_HTTP_PORT)
+PROVIDER_HOST     ?= $(PROVIDER_HOSTNAME):$(APP_HTTP_PORT)
```
boz, line 22: This `APP_HTTP_PORT` doesn't exist anymore.

Makefile
```diff
@@ -5,6 +5,10 @@ APP_DIR := ./app
 GO := GO111MODULE=on go
 GOBIN := $(shell go env GOPATH)/bin
 
+KIND_API_IP ?= $(shell make -sC _run/kube kind-k8s-ip)
+KIND_HTTP_PORT ?= $(shell make -sC _run/kube app-http-port)
+KIND_VARS ?= KIND_APP_PORT="$(KIND_HTTP_PORT)" KIND_K8S_IP="$(KIND_API_IP)"
```
boz, line 10: why create new var names for the same thing?  can you use this?

```
KIND_VARS ?= KIND_HTTP_PORT="$(KIND_HTTP_PORT)" KIND_API_IP="$(KIND_API_IP)"
```

Ropes: Originally I saw a Makefile error of something like "cannot assign variable name recursively", however possibly after adding the quote interpolation it ceased. So good catch, cleaning up the name spam, and making consistently `KIND_APP....`

_run/common-kind.mk
Ropes: Edited, hopefully clearer now.

PR #781, opened 2020-08-10
Update Akash to Cosmos-SDK 0.40 - Proto Migration

### LINKED ISSUE #755: Protobuf migration

akhilkumarpilli:
Fixes #755

---
This change is [Reviewable](https://reviewable.io/reviews/ovrclk/akash/781)


Makefile
```diff
@@ -225,23 +225,126 @@ update-swagger-docs:
...
+  PROTOC_ZIP ?= protoc-3.11.2-osx-x86_64.zip
+endif
+
+proto-tools: proto-tools-stamp buf
+
+proto-tools-stamp:
+	echo "Installing protoc compiler..."
+	(cd /tmp; \
+	curl -OL "https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VERSION}/${PROTOC_ZIP}"; \
+	unzip -o ${PROTOC_ZIP} -d $(PREFIX) bin/protoc; \
```
Ropes, line 326: Another option is to install in `$GOPATH/bin` then the script doesn't need root permissions.

jackzampolin:
Ok, taking a first crack at running the chain and the demo:

* When creating a `gentx` I'm unable to set the `client-home` so when using `--keyring-backend` its looking for the keys in `~/.akashd` not `~/.akashctl`. 
* What happened to `akashctl config`? This breaks all the `_run` examples
* Larger question: do we want to do what `gaia` has done and only release one binary `akash` with all of the functionality?

anilCSE:
> * When creating a `gentx` I'm unable to set the `client-home` so when using `--keyring-backend` its looking for the keys in `~/.akashd` not `~/.akashctl`.

Might be some issue with `config`. If we decide to build single binary, this should be straightforward. 

> * What happened to `akashctl config`? This breaks all the `_run` examples

Config is removed now (upstream). @akhilkumarpilli is there a way to fix this?

> * Larger question: do we want to do what `gaia` has done and only release one binary `akash` with all of the functionality?

I suggest that, let's have single binary.

akhilkumarpilli:
>     * When creating a `gentx` I'm unable to set the `client-home` so when using `--keyring-backend` its looking for the keys in `~/.akashd` not `~/.akashctl`.
>
`gentx command ` is modified and we don't have any `client-home` flag now in upstream. If we decide to build single binary, this will not be issue.
>     * What happened to `akashctl config`? This breaks all the `_run` examples
> 
`config command` is removed completely and we need to pass flags for every command.

jackzampolin:
Let's go for the single binary `akash` and rebuild the config command. Thanks @anilCSE and @akhilkumarpilli.

jackzampolin:
Ok @anilCSE @akhilkumarpilli I've merged in the single binary PR and fixed a number of the tests. The simulation tests are all failing on proto migration related issues. Do you mind starting there tomorrow morning?

_run/common.mk
```diff
@@ -60,25 +51,24 @@ node-init-genesis: init-dirs
...
 .PHONY: node-init-genesis-accounts
 node-init-genesis-accounts: $(patsubst %,node-init-genesis-account-%,$(KEY_NAMES))
-	$(AKASHD) validate-genesis
+	# $(AKASHD) validate-genesis
 
 .PHONY: node-init-genesis-account-%
 node-init-genesis-account-%:
 	$(AKASHD) add-genesis-account \
-		"$(shell $(AKASHCTL_NONODE) keys show "$(@:node-init-genesis-account-%=%)" -a)" \
+		"$(shell $(AKASHCTL_NONODE) keys show --keyring-backend "test" "$(@:node-init-genesis-account-%=%)" -a)" \
```
boz, line 59: `$(KEY_OPTS)`?

_run/kube/Makefile
```diff
@@ -9,6 +9,7 @@ GATEWAY_ENDPOINT ?= http://$(GATEWAY_HOST)
 provider-run:
 	AKASH_DEPLOYMENT_INGRESS_STATIC_HOSTS="false" \
 		$(AKASHCTL) $(KEY_OPTS) provider run \
+			--chain-id "$(CHAIN_NAME)" \
```
boz, line 12: I think this is in `$(KEY_OPTS)`.

app/app.go
```diff
@@ -75,129 +108,143 @@ type AkashApp struct {
...
-	app.keeper.supply = supply.NewKeeper(
-		app.cdc,
-		app.keys[supply.StoreKey],
+	// app.keeper.supply = supply.NewKeeper(
+	// 	app.cdc,
+	// 	app.keys[supply.StoreKey],
+	// 	app.keeper.acct,
+	// 	app.keeper.bank,
+	// 	macPerms(),
+	// )
```
boz, line 169: TODO or no-longer needed?

app/app_test.go
```diff
@@ -1,48 +1,48 @@
...
+// 	app1 := app.NewApp(log.NewTMLogger(log.NewSyncWriter(os.Stdout)), db, nil, 0, map[int64]bool{}, app.DefaultHome)
 
-	for acc := range macPerms() {
-		require.True(t, app.keeper.bank.BlacklistedAddr(app.keeper.supply.GetModuleAddress(acc)))
-	}
-}
+// 	for acc := range app.MacPerms() {
+// 		require.True(t, app1.keeper.bank.BlockedAddr(app1.keeper.acct.GetModuleAddress(acc)))
+// 	}
+// }
```
boz, line 48: todo?

provider/bidengine/service.go
```diff
@@ -170,16 +171,23 @@ loop:
...
+	order *mtypes.Order
+	bid   *mtypes.Bid
 }
 
 func queryExistingOrders(_ context.Context, session session.Session) ([]existingOrder, error) {
-	orders, err := session.Client().Query().Orders(mquery.OrderFilters{})
+	params := &mtypes.QueryOrdersRequest{
+		Filters: mtypes.OrderFilters{},
+		Pagination: &sdkquery.PageRequest{
+			Limit: 10000,
```
boz, line 182: todo #726

provider/manifest/service.go
```diff
@@ -269,11 +266,17 @@ func fetchExistingLeases(_ context.Context, session session.Session) ([]event.Le
 
 	items := make([]event.LeaseWon, 0, len(leases))
 	for _, lease := range leases {
-		dgroup, err := session.Client().Query().Group(lease.GroupID())
+		res, err := session.Client().Query().Group(
+			context.Background(),
+			&dtypes.QueryGroupRequest{
+				ID: lease.LeaseID.BidID().GroupID(),
```
boz, line 272: We should have `DeploymentID()`, `GroupID()`, `OrderID()` methods on `LeaseID` too.

x/deployment/client/cli/flags.go
```diff
@@ -71,41 +76,59 @@ func GroupIDFromFlags(flags *pflag.FlagSet) (types.GroupID, error) {
...
+// 	dfilters, err := DepFiltersFromFlags(flags)
+// 	if err != nil {
+// 		return query.GroupFilters{}, err
+// 	}
+// 	gfilters := query.GroupFilters{
+// 		Owner:        dfilters.Owner,
+// 		StateFlagVal: dfilters.StateFlagVal,
+// 	}
+// 	return gfilters, nil
+// }
```
boz, line 134: todo?

x/deployment/client/cli/query.go
```diff
@@ -21,67 +20,105 @@ func GetQueryCmd(key string, cdc *codec.Codec) *cobra.Command {
...
+			dfilters, state, err := DepFiltersFromFlags(cmd.Flags())
 			if err != nil {
 				return err
 			}
 
-			obj, err := query.NewClient(ctx, key).Deployments(dfilters)
+			// checking state flag
+			stateVal, ok := types.Deployment_State_value[state]
+
+			if (!ok && (state != "")) || state == "invalid" {
```
boz, line 54: use `stateVal` for the second comparison.

x/deployment/client/cli/query.go
```diff
@@ -21,67 +20,105 @@ func GetQueryCmd(key string, cdc *codec.Codec) *cobra.Command {
...
+			clientCtx := client.GetClientContextFromCmd(cmd)
+			clientCtx, err := client.ReadQueryCommandFlags(clientCtx, cmd.Flags())
+			if err != nil {
+				return err
+			}
+
+			queryClient := types.NewQueryClient(clientCtx)
 
-			id, err := DeploymentIDFromFlags(cmd.Flags(), ctx.FromAddress.String())
+			id, err := DeploymentIDFromFlags(cmd.Flags(), "")
```
boz, line 100: is the `--from` address picked up automatically another way?

x/deployment/keeper/grpc_query.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkquery "github.com/cosmos/cosmos-sdk/types/query"
+	"github.com/ovrclk/akash/x/deployment/types"
+)
+
+// Querier is used as Keeper will have duplicate methods if used directly, and gRPC names take precedence over keeper
+type Querier struct {
+	Keeper
+}
```
boz, line 18: can we keep this private?

boz (commented):
Monster change.  Great job.  Some comments inline.  I'll have to take a second pass too.

Is this ready for merging?  Please do call out as many "todo" items that we can push out until after merging this - there's no need to get everything in one fell swoop.

x/deployment/client/cli/query.go
akhilkumarpilli: We defined state with 0 value as InvalidState in types. So, when user enters `invalid` as state filter value, then it will considered as a State and will return all values. To solve this, I am checking entered value not to be `invalid`.

x/deployment/client/cli/query.go
akhilkumarpilli: `--from` flag is not included in queries

x/deployment/client/cli/query.go
boz: yeah, can you check `stateVal == types.DeploymentState_Invalid` (or whatever the constant's name is) instead of the string comparison?  It's easier to keep track of that way.

akhilkumarpilli: No,`stateval` will be equal to `types.DeploymentState_Invalid` when state filter is not applied. So we cannot use stateval to compare here.

boz: if `state == "invalid"`, won't `stateVal == types.DeploymentState_Invalid`?

akhilkumarpilli: Even if `state==" "`, `stateVal == types.DeploymentState_Invalid` because by default, stateVal will be 0

boz: That would be handled by `!ok && state != ""`. Also, people shouldn't be typing `"invalid"` ever - they should leave it as blank to get all results.

akhilkumarpilli: If people type invalid then ok will be true. That's the reason I am handling it to just avoid such cases. If not needed, I can remove it but you eill get all results if state filter value entered is `invalid`.

boz: Right.  I was asking for the second part to use the constant.  You keep saying that it won't work for whatever reason.

Forget it.  Maybe I'm missing something.  I'll change it myself if not.

_run/common.mk
```diff
@@ -60,25 +51,24 @@ node-init-genesis: init-dirs
 
 .PHONY: node-init-genesis-accounts
 node-init-genesis-accounts: $(patsubst %,node-init-genesis-account-%,$(KEY_NAMES))
-	$(AKASHD) validate-genesis
+	# $(AKASHD) validate-genesis
```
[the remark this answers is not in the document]
jackzampolin: I was having issues with the genesis encoding. This is still TODO

_run/common.mk
jackzampolin: Messy on my part 🤦

provider/bidengine/service.go
jackzampolin: cc @akhilkumarpilli @anilCSE

x/deployment/client/cli/query.go
jackzampolin: Yes, from gets picked up in the `init()` function and saved on `context`.

_run/kube/Makefile
jackzampolin: Its not which is why I had to add this. We used to pass it in via `akashctl config`, now needs to be added manually. Would be nice if the SDK picked that up off of the genesis file...

boz: maybe it's been changed since then but it looks like it's included [now](https://github.com/ovrclk/akash/pull/781/files/2a91040524a235ddbd992b34ede7c712f8806db4#diff-a4a9755559e6a7113337c16c483d84d8R18)

agreed that not having the config command makes this really cumbersome.

x/deployment/client/cli/query.go
boz: Sorry, I'm an idiot.  You're right.

_run/kube/Makefile
jackzampolin: I have trouble thinking about the Makefile import cycle sometimes 🤦 thank you for finding that

PR #782, opened 2020-08-12
feat: SDL v2

troian:


jackzampolin:
cc @anilCSE @akhilkumarpilli RE: any feedback on proto types for the new resources

sdl/cpu.go
```diff
@@ -0,0 +1,48 @@
...
+
+type v2ResourceCPU struct {
+	Units      cpuQuantity     `yaml:"units"`
+	Attributes v2CPUAttributes `yaml:"attributes,omitempty"`
+}
+
+func (sdl *v2CPUAttributes) UnmarshalYAML(node *yaml.Node) error {
+	attr := make(v2CPUAttributes)
+
+	for i := 0; i < len(node.Content); i += 2 {
```
boz, line 21: `i+1 < len(node.Content)`

validation/resources.go
```diff
@@ -45,6 +46,32 @@ func validateResourceLists(config ValConfig, rlists []types.ResourceGroup) error
...
+	u.cpu.Add(rhs.cpu)
+	u.memory.Add(rhs.memory)
+	u.storage.Add(rhs.storage)
+}
+
+func (u *resourceLimits) mul(count uint32) {
+	u.cpu.MulUint64(uint64(count))
+	u.memory.MulUint64(uint64(count))
+	u.storage.MulUint64(uint64(count))
+}
```
boz, line 73: :fire:

manifest/types.go
```diff
@@ -42,14 +42,15 @@ type Service struct {
 	Command []string
 	Args    []string
 	Env     []string
-	Unit    types.Unit
+	// rename Units to Resources???
```
boz, line 45: :+1:

sdl/cpu.go
boz: :point_up:

sdl/sdl.go
```diff
@@ -14,13 +16,60 @@ import (
...
+	data    SDL            `yaml:"-"`
+}
+
+func (s *sdl) UnmarshalYAML(node *yaml.Node) error {
+	var result sdl
+
+	for idx := range node.Content {
+		if node.Content[idx].Value == "version" {
+			var err error
+			if result.Version, err = semver.ParseTolerant(node.Content[idx+1].Value); err != nil {
```
boz, line 42: `node.Content[idx+1]` might not exist.

sdl/sdl.go
```diff
@@ -52,7 +96,7 @@ func Read(buf []byte) (SDL, error) {
 		vgroups = append(vgroups, *dgroup)
 	}
 
-	if err := validation.ValidateDeploymentGroups(vgroups); err != nil {
+	if err = validation.ValidateDeploymentGroups(vgroups); err != nil {
```
boz, line 99: `:=`

sdl/sdl.go
```diff
@@ -95,5 +139,22 @@ func ManifestVersion(manifest manifest.Manifest) ([]byte, error) {
...
+func (s *sdl) DeploymentGroups() ([]*dtypes.GroupSpec, error) {
+	if s.data == nil {
+		return []*dtypes.GroupSpec{}, errUninitializedConfig
+	}
+
+	return s.data.DeploymentGroups()
+}
+
+func (s *sdl) Manifest() (manifest.Manifest, error) {
+	if s.data == nil {
```
boz, line 155: might want to do `len(s.data) == 0` in these so there's no issue of the various kinds of `nil`

sdl/cpu.go
troian: missed that comment. shouldn't be necessary, or i missed thing 😄

types/resource.go
```diff
@@ -1,27 +1,309 @@
...
+	}
+
+	return
+}
+
+// Sub tbd
+func (u ResourceUnits) Sub(rhs ResourceUnits) (res ResourceUnits, err error) {
+	// check for in case some smartypants wants to overflow
+	defer func() {
+		if r := recover(); r != nil {
```
boz, line 114: let's just use big ints and return error if there's an overflow.  this catches too much.

types/resource.go
```diff
@@ -1,27 +1,309 @@
...
+			err = errors.Errorf("%v", r)
+		}
+
+		if err != nil {
+			res = ResourceUnits{}
+		}
+	}()
+
+	if (u.CPU == nil && rhs.CPU != nil) ||
```
boz, line 123: can we treat `nil` as `0`?  we do that for the lhs in `Add()`.

sdl/sdl.go
troian: sure thing

types/resource.go
troian: thats a yet another topic to discuss.
in this implementation if lhs is missing it will be copied from rhs

manifest/types.go
troian: seems good to rename it in SDL too then

sdl/cpu.go
boz: defensive programming.  it's not clear why the yaml would always be a map.

types/resource.go
boz: in `Add()` it is, but here you're returning an error.

proto/akash/resourcevalue.proto
```diff
@@ -0,0 +1,15 @@
...
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/ovrclk/akash/types";
+
+// Unit stores cpu, memory and storage metrics
+message ResourceValue {
+  option (gogoproto.equal) = true;
+  bytes val = 1 [
+    (gogoproto.nullable) = false,
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int"
```
boz, line 13: why wrap this in `ResourceValue` instead of just using a `types.Int` for the field values?

types/attribute.go
```diff
@@ -0,0 +1,140 @@
...
+
+import (
+	"reflect"
+
+	"gopkg.in/yaml.v3"
+)
+
+type Attributes []Attribute
+
+type AttributeValue struct {
```
boz, line 11: I thought we were going with string values for now?

provider/cluster/inventory_test.go
```diff
@@ -51,13 +57,17 @@ func TestInventory_reservationAllocateable(t *testing.T) {
 	}{
 		{mkres(false, mkrg(100, 1*unit.G, 2)), true},
 		{mkres(false, mkrg(100, 4*unit.G, 1)), true},
-		{mkres(false, mkrg(250, 1*unit.G, 1)), true},
+		{mkres(false, mkrg(250, 1*unit.G, 1)), false},
```
boz, line 60: is this an open/closed boundary change?

provider/cluster/inventory_test.go
```diff
@@ -51,13 +57,17 @@ func TestInventory_reservationAllocateable(t *testing.T) {
...
 		{mkres(false, mkrg(1000, 1*unit.G, 1)), false},
 		{mkres(false, mkrg(100, 9*unit.G, 1)), false},
 	}
 
-	for _, test := range tests {
-		assert.Equal(t, test.ok, reservationAllocateable(inventory, reservations, test.res))
+	for idx, test := range tests {
+		res := reservationAllocateable(inventory, reservations, test.res)
+		if idx == 0 && res {
```
boz, line 67: just do the first one as a separate test.

sdl/_testdata/simple.yaml
```diff
@@ -1,22 +1,22 @@
...
-      memory: "128Mi"
-      storage: "1Gi"
-
+      resources:
+        cpu:
+          units: "100m"
+        memory:
+          amount: "128Mi"
+        storage:
+          amount: "1Gi"
```
boz, line 19: `amount` -> `size`?

sdl/v1.go
```diff
@@ -226,11 +200,11 @@ func (sdl *v1) Manifest() (manifest.Manifest, error) {
...
-				Unit: types.Unit{
-					CPU:     uint32(compute.CPU),
-					Memory:  uint64(compute.Memory),
-					Storage: uint64(compute.Storage),
-				},
+				// Unit: types.Unit{
+				// 	CPU:     uint32(compute.CPU),
+				// 	Memory:  uint64(compute.Memory),
+				// 	Storage: uint64(compute.Storage),
+				// },
```
boz, line 207: what to do with all of the commented out sections in here?

fwiw, we don't need to maintain backwards compatibility

types/resource.go
```diff
@@ -1,20 +1,282 @@
...
+var _ Unit = (*CPU)(nil)
+var _ Unit = (*Memory)(nil)
+var _ Unit = (*Storage)(nil)
+
+// AddUnit it rather searches for existing entry of the same type and sums values
+// if type not found it appends
+func (m ResourceUnits) Add(rhs ResourceUnits) (res ResourceUnits, err error) {
+	res = m
+
+	defer func() {
```
boz, line 43: let's not use `defer` for this - just return like normal, without using `(res Foo, err error)` style variable declarations.

x/deployment/types/types.go
```diff
@@ -12,10 +12,10 @@ import (
 // DefaultOrderBiddingDuration is the default time limit for an Order being active.
 // After the duration, the Order is automatically closed.
 // ( 24(hr) * 3600(seconds per hour) ) / 7s-Block
-const DefaultOrderBiddingDuration int64 = int64(12342)
```
boz, line 15: thank you

x/market/client/cli/cli_test.go
```diff
@@ -196,7 +196,8 @@ func (s *IntegrationTestSuite) Test2CreateBid() {
 	resp, err = cli.QueryOrdersExec(val.ClientCtx.WithOutputFormat("json"))
 	s.Require().NoError(err)
 
-	var result *types.QueryOrdersResponse = &types.QueryOrdersResponse{}
+	fmt.Println(string(resp.Bytes()))
+	var result = &types.QueryOrdersResponse{}
```
boz, line 200: thank you.   one step farther: `result := &types.QueryOrdersResponse{}`

boz (commented):
Looking great!

pkg/apis/akash.network/v1/types.go
```diff
@@ -255,36 +259,50 @@ func manifestServiceExposeFromAkash(amse manifest.ServiceExpose) ManifestService
...
 
-func resourceUnitFromAkash(aru types.Unit) ResourceUnit {
-	return ResourceUnit{
-		CPU:     aru.CPU,
-		Memory:  strconv.FormatUint(aru.Memory, 10),
-		Storage: strconv.FormatUint(aru.Storage, 10),
+func resourceUnitsFromAkash(aru types.ResourceUnits) ResourceUnits {
+	res := ResourceUnits{}
+	if aru.CPU != nil {
+		res.CPU = uint32(aru.CPU.Units.Value())
```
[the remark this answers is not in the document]
troian: good point, depends on what maximum allowed boundaries for CPU on validation step here https://github.com/ovrclk/akash/blob/f76e2c6a2a5ef4a5664682fe85012aba2a091852/validation/config.go#L9. if thats up to max value handled by `uint` (on 64 bit machine) than have to change. Needs CRD update then too

sdl/_testdata/simple.yaml
troian: previously it was `Size` member in Storage and Memory, but protobuf generates own `Size()`. can have field named `Amount` or `Sz` but tags as `size`?

boz: Can we keep it `size` in the yaml?

boz:
`Quantity` works for field name, less worried about that right now tho.  `amount` seems a bit off tho.  Will sleep on it.

sdl/_testdata/simple.yaml
troian: yes

provider/cluster/inventory.go
```diff
@@ -320,40 +317,32 @@ func (is *inventoryService) runCheck(ctx context.Context) <-chan runner.Result {
 	})
 }
 
-func (is *inventoryService) getStatus(
-	inventory []Node, reservations []*reservation) InventoryStatus {
-
-	status := InventoryStatus{}
+func (is *inventoryService) getStatus(inventory []Node, reservations []*reservation) (status InventoryStatus) {
```
boz, line 320: :nail_care: let's only use the `(x Foo)` when it's absolutely necessary (catching a `panic`).

types/resource.go
```diff
@@ -1,20 +1,271 @@
...
+	return res, nil
+}
+
+// Sub tbd
+func (m ResourceUnits) Sub(rhs ResourceUnits) (res ResourceUnits, err error) {
+	defer func() {
+		if err != nil {
+			res = ResourceUnits{}
+		}
+	}()
```
boz, line 76: :scissors: `defer`

validation/pricing.go
```diff
@@ -31,13 +30,15 @@ func validateGroupPricing(config ValConfig, gspec dtypes.GroupSpec) error {
 			price = price.Add(rprice)
 		}
 
-		mem = mem.Add(
-			sdk.NewIntFromUint64(resource.Unit.Memory).
-				Mul(sdk.NewIntFromUint64(uint64(resource.Count))))
+		memCount := sdk.NewInt(0)
+		if u := resource.Resources.Memory; u != nil {
```
boz, line 34: do we ever ensure that there's memory declared?

validation/resources.go
```diff
@@ -114,21 +137,35 @@ func validateResourceGroup(config ValConfig, rg types.Resource) error {
...
 }
 
-func validateResourceUnit(config ValConfig, unit types.Unit) error {
-	if unit.CPU > uint32(config.MaxUnitCPU) || unit.CPU < uint32(config.MinUnitCPU) {
-		return errors.Errorf("error: invalide unit cpu (%v > %v > %v fails)",
-			config.MaxUnitCPU, unit.CPU, config.MinUnitCPU)
+func validateResourceUnit(config ValConfig, units types.ResourceUnits) (resourceLimits, error) {
+	limits := newLimits()
+
+	if u := units.CPU; u != nil {
```
boz, line 146: do we ever ensure that `CPU` is declared?

validation/resources.go
```diff
@@ -114,21 +137,35 @@ func validateResourceGroup(config ValConfig, rg types.Resource) error {
...
+			return resourceLimits{}, errors.Errorf("error: invalid unit memory (%v > %v > %v fails)",
+				config.MaxUnitMemory, u.Quantity.Value(), config.MinUnitMemory)
+		}
+		limits.memory = limits.memory.Add(u.Quantity.Val)
 	}
-	if unit.Storage > uint64(config.MaxUnitStorage) || unit.Storage < uint64(config.MinUnitStorage) {
-		return errors.Errorf("error: invalid unit storage (%v > %v > %v fails)",
-			config.MaxUnitStorage, unit.Storage, config.MinUnitStorage)
+
+	if u := units.Storage; u != nil {
```
boz, line 162: same with storage - needs to be not nil.

boz (commented):
huge

validation/resources.go
troian: mean instantiation of big.Int value?

boz: Like did the user say the amount.  Basically, I think that `nil` should be an error

troian: i think we decided to keep them like this on initial, but i could be wrong.
where `CPU` and `Memory` should be `required`, storage though is more wide. for example deployment does not need storage (apart from docker image which in case of ML can occupy gigabytes) just default it to 0?

boz: This is validation, if cpu, memory, or storage are 0 or nil it is an error.

boz: I think I read your comment wrong because this line section allows nil on all of them.  Yes, I think defaulting storage to zero is okay for now.

troian: all resource values if they are != nil, validated against `ValConfig` values, if `Storage` is not nil and defaulted to 0, then check fails as default `MinUnitStorage` is 1024.

boz: Yes but if they are nil then it is an invalid resource

troian: yah, thats confusing, remember initially we though to parse resources into slice instead of using pointers and on validation step we woudln't deal with pointers

boz: I don't understand why we're talking past each other.  Validation should return an error when there is no cpu, memory.  If that is not the behavior then the behavior needs to be changed. 

Validation is used in multiple places.  It is the definition of a correctly-formed group/resource/etc.  It needs to be strict and it needs to be correct.

PR #784, opened 2020-08-20
Provider Pod Spec Security Defaults

Ropes:
Following Kubernetes best practices for deploying Pods to limit a
tennants ability to escape the container and escalate privileges. Just a
small baseline to reduce attack vectors, much more work to do.

* PodsSpec:
    * RunAsNonRoot setting to disable images with user set to root
    * Disable mounting Service Account tokens to Containers
* Container Security Context
    * Privileged set to false
    * AllowPrivilegeEscalation disabled


provider/cluster/kube/builder.go
```diff
@@ -99,6 +99,9 @@ func (b *deploymentBuilder) labels() map[string]string {
 
 func (b *deploymentBuilder) create() (*appsv1.Deployment, error) { // nolint:golint,unparam
 	replicas := int32(b.service.Count)
+	falseVal := false
```
troian, line 102: `runAsRoot := true`

provider/cluster/kube/builder.go
```diff
@@ -136,6 +143,7 @@ func (b *deploymentBuilder) update(obj *appsv1.Deployment) (*appsv1.Deployment,
 func (b *deploymentBuilder) container() corev1.Container {
 	qcpu := resource.NewScaledQuantity(int64(b.service.Unit.CPU), resource.Milli)
 	qmem := resource.NewQuantity(int64(b.service.Unit.Memory), resource.DecimalSI)
+	falseVal := false
```
troian, line 146: `privileged := false` 
`privilegeEscalation := false`

provider/cluster/kube/builder.go
Ropes: Agreed, good to be explicit.

PR #785, opened 2020-08-25
Integrate IBC

akhilkumarpilli:


jackzampolin:
One thing we need to add to this branch before merge is a `Dockerfile` that exposes the interface needed for relayer testing. [See documentation here](https://github.com/iqlusioninc/relayer/blob/master/docs/testing.md). I'll be working on the relayer in the coming weeks, and this would ensure that Akash is compatable.

PR #786, opened 2020-08-25
Migrate to single binary: akash

jackzampolin:
cc @boz @akhilkumarpilli @anilCSE


cmd/akash/cmd/genaccounts.go
```diff
@@ -158,14 +158,14 @@ contain valid denominations. Accounts may optionally be supplied with vesting pa
...
-				return errors.Errorf("failed to marshal bank genesis state: %w", err)
+				return errors.Errorf("failed to marshal bank genesis state: %v", err)
 			}
 
 			appState[banktypes.ModuleName] = bankGenStateBz
 
 			appStateJSON, err := json.Marshal(appState)
 			if err != nil {
-				return errors.Errorf("failed to marshal application genesis state: %w", err)
+				return errors.Errorf("failed to marshal application genesis state: %v", err)
```
boz, line 168: `%w` is for [wrapping errors](https://blog.golang.org/go1.13-errors)

events/publish.go
```diff
@@ -38,7 +38,7 @@ func Publish(ctx context.Context, tmbus tmclient.EventsClient, name string, bus
 	if err != nil {
 		return err
 	}
-	defer tmbus.UnsubscribeAll(ctx, blkname)
+	defer tmbus.UnsubscribeAll(ctx, txname)
```
boz, line 41: `txname` is unsubscribed from above; this should be `blkname`.

events/publish.go
```diff
@@ -80,7 +80,9 @@ loop:
 func processEvents(bus pubsub.Bus, events []abci.Event) {
 	for _, ev := range events {
 		if mev, ok := processEvent(ev); ok {
-			bus.Publish(mev)
+			if err := bus.Publish(mev); err != nil {
```
boz, line 83: panic or at least log it.  this shouldn't really happen.

provider/cmd/run.go
```diff
@@ -130,7 +136,9 @@ func doRunCmd(ctx context.Context, cmd *cobra.Command, _ []string) error {
 
 	service, err := provider.NewService(ctx, session, bus, cclient)
 	if err != nil {
-		group.Wait()
+		if err = group.Wait(); err != nil {
+			return err
+		}
```
boz, line 141: `return group.Wait()`

pubsub/bus.go
```diff
@@ -140,7 +140,9 @@ loop:
 
 			// Publish to children.
 			for sub := range b.subscriptions {
-				sub.Publish(ev)
+				if err := sub.Publish(ev); err != nil {
```
boz, line 143: same here.

x/deployment/keeper/key.go
```diff
@@ -18,16 +18,22 @@ var (
 func deploymentKey(id types.DeploymentID) []byte {
 	buf := bytes.NewBuffer(deploymentPrefix)
 	buf.Write(id.Owner.Bytes())
-	binary.Write(buf, binary.BigEndian, id.DSeq)
+	if err := binary.Write(buf, binary.BigEndian, id.DSeq); err != nil {
```
boz, line 21: panic on all of these.

x/deployment/types/event.go
```diff
@@ -151,7 +151,9 @@ func encodeHex(src []byte) []byte {
 
 func decodeHex(src []byte) []byte {
 	dst := make([]byte, hex.DecodedLen(len(src)))
-	hex.Decode(dst, src)
```
boz, line 154: have this function return an error

x/market/handler/endblock.go
```diff
@@ -52,7 +52,7 @@ func transferFundsForActiveLeases(ctx sdk.Context, keepers Keepers) error {
 	return nil
 }
 
-var errNoBids error = errors.New("no bids to pick winner from")
+var ErrNoBids error = errors.New("no bids to pick winner from")
```
boz, line 55: the `error` isn't needed here.  I think that this wasn't exported because it's not used anywhere, would prefer to keep it that way.

x/market/keeper/key.go
```diff
@@ -19,18 +19,30 @@ var (
 func orderKey(id types.OrderID) []byte {
 	buf := bytes.NewBuffer(orderPrefix)
 	buf.Write(id.Owner.Bytes())
-	binary.Write(buf, binary.BigEndian, id.DSeq)
-	binary.Write(buf, binary.BigEndian, id.GSeq)
-	binary.Write(buf, binary.BigEndian, id.OSeq)
+	if err := binary.Write(buf, binary.BigEndian, id.DSeq); err != nil {
```
boz, line 22: again all of these key generation things should blow up if they fail.  even if we return an error, it's not clear how to handle it gracefully.

x/market/module.go
```diff
@@ -155,7 +155,9 @@ func (am AppModule) BeginBlock(_ sdk.Context, _ abci.RequestBeginBlock) {}
 // EndBlock returns the end blocker for the market module. It returns no validator
 // updates.
 func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
-	handler.OnEndBlock(ctx, am.keepers)
+	if err := handler.OnEndBlock(ctx, am.keepers); err != nil {
+		return nil
```
boz, line 159: panic?

jackzampolin:
@boz RE: moving back to single binary. While the SDK isn't technically opinionated on this subject (we could still do 2 binaries) `gaia` has moved to a single binary and some of the CLI functionality wasn't working as intended. Also I've heard an internal desire for a single binary in the past.

cmd/akash/cmd/genaccounts.go
jackzampolin: Thanks, aware of this. `make test-full` was throwing errors/warnings telling me that doing `%w` calls in `Errorf` is not recommended which is why I changed them. If you would like me to change them back I'm happy to.

x/market/handler/endblock.go
jackzampolin: This was causing a failing test, and i exported it to make the `error.Is` call pass properly.

x/market/keeper/key.go
jackzampolin: `panic`ing now

_docs/kustomize/akashd/deployment.yaml
```diff
@@ -1,19 +1,19 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: akashd
+  name: akash
```
Ropes, line 4: Is this the pattern that `gaia` is going with for its "Node" service type? My inclination is for a little bit more explicit information on the Deployment name itself. `akash-node`? We'll need to make it clear between the node and `provider` service. 

Also need to apply the name to labels and selectors.

_docs/kustomize/akashd/kustomization.yaml
```diff
@@ -3,19 +3,19 @@ resources:
   - service.yaml
   - ingress.yaml
 commonLabels:
-  app: akashd
-  akash.network/component: akashd
+  app: akash
+  akash.network/component: akash
```
Ropes, line 7: Related to my previous explicit naming comment.

cmd/akash/cmd/genaccounts.go
jackzampolin: Marking resolved per our conversation at standup.

_docs/kustomize/akashd/deployment.yaml
jackzampolin: thank you for catching that! I've reverted that change.

Makefile
```diff
@@ -121,7 +111,7 @@ SUBLINTERS = deadcode \
 
 # TODO: ^ gochecknoglobals
 
-LINT = $(GOBIN)/golangci-lint run ./... --disable-all --enable 
+LINT = $(GOBIN)/golangci-lint run ./... --disable-all --deadline=5m --enable
```
troian, line 114: `LINT` variable seems used when running specific linter for example `$(LINT) structcheck` and looks like 5m deadline is too big for that. 1m or even 30s?

_run/common-commands.mk
```diff
@@ -1,10 +1,15 @@
 KEY_NAME          ?= main
-KEY_ADDRESS       ?= $(shell $(AKASHCTL_NONODE) keys show "$(KEY_NAME)" -a)
+KEY_OPTS 		  := --keyring-backend=test
+CHAIN_NAME 		  := local
+CHAIN_OPTS   	  := --chain-id=$(CHAIN_NAME)
+
+KEY_ADDRESS       ?= $(shell $(AKASHCTL_NONODE) keys show $(KEY_OPTS) "$(KEY_NAME)" -a)
```
troian, line 6: alignment for var values

events/cmd/root.go
```diff
@@ -27,7 +27,9 @@ func EventCmd() *cobra.Command {
 	}
 
 	cmd.Flags().String(flags.FlagNode, "tcp://localhost:26657", "The node address")
-	_ = viper.BindPFlag(flags.FlagNode, cmd.Flags().Lookup(flags.FlagNode))
+	if err := viper.BindPFlag(flags.FlagNode, cmd.Flags().Lookup(flags.FlagNode)); err != nil {
+		return nil
```
troian, line 31: does caller check for nil result from EventCmd?

events/publish.go
```diff
@@ -80,7 +86,9 @@ loop:
...
+				panic(err)
```
troian, line 90: should we handle err to gracefully shutdown?

provider/cmd/root.go
```diff
@@ -14,7 +14,9 @@ func RootCmd() *cobra.Command {
 	}
 
 	cmd.PersistentFlags().String(flags.FlagNode, "http://localhost:26657", "The node address")
-	_ = viper.BindPFlag(flags.FlagNode, cmd.PersistentFlags().Lookup(flags.FlagNode))
+	if err := viper.BindPFlag(flags.FlagNode, cmd.PersistentFlags().Lookup(flags.FlagNode)); err != nil {
+		return nil
```
troian, line 18: same as EventsCmd

provider/manifest/service.go
```diff
@@ -90,7 +89,7 @@ type service struct {
 	managers  map[string]*manager
 	managerch chan *manager
 
-	deploymentch chan runner.Result
+	// deploymentch chan runner.Result
```
troian, line 92: comment why it should stay commented?

x/deployment/keeper/key.go
```diff
@@ -18,16 +18,22 @@ var (
...
 }
 
 // groupKey provides prefixed key for a Group's marshalled data.
 func groupKey(id types.GroupID) []byte {
 	buf := bytes.NewBuffer(groupPrefix)
 	buf.Write(id.Owner.Bytes())
-	binary.Write(buf, binary.BigEndian, id.DSeq)
-	binary.Write(buf, binary.BigEndian, id.GSeq)
+	if err := binary.Write(buf, binary.BigEndian, id.DSeq); err != nil {
+		panic(err)
```
troian, line 32: generally we would want to avoid using panics as they immediately shutdown app.
maybe this way?
```go
func groupKey(id types.GroupID) ([]byte, error)
```

events/publish.go
```diff
@@ -50,7 +54,9 @@ func Publish(ctx context.Context, tmbus tmclient.EventsClient, name string, bus
 		return publishEvents(ctx, blkch, bus)
 	})
 
-	return g.Wait()
+	err = g.Wait()
+
+	return err
```
troian, line 59: don't need to `return err` if err declared with return type
can just do `return`

x/deployment/keeper/key.go
jackzampolin: In the keepers we have a pattern of panic-ing on unexpected behavior. Safety over liveness as a design decision.

provider/manifest/service.go
jackzampolin: Its not used anywhere so I've commented it out.

events/cmd/root.go
jackzampolin: A `nil` will result in the app not starting at all. Errors here only happen when adding conflicting flags. IMO this is sufficent.

Makefile
jackzampolin: cc @akhilkumarpilli for rationale here.

provider/manifest/service.go
troian: worth to remove then, so won't confuse in further

events/publish.go
jackzampolin:
```go
bus.Close()
return
```

jackzampolin:
Thanks for the review @troian, I'm going to go ahead and merge this PR.

PR #798, opened 2020-09-03
mainnet updates

boz:


.github/workflows/release.yaml
```diff
@@ -2,8 +2,15 @@ name: release
 
 on:
   push:
-    branches: [ master ]
-    tags:     [  v*    ]
+    branches:
+      - master
+      - mainnet/main
+    tags:
+      - v*
```
troian, line 9: shall we validate it is a proper semver? like this `^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`

boz: absolutely not lol.  at most `^v\.\d+\.\d+\.\d+-.*$` but I don't even think the time to tinker with it is worth it.

troian: was going crazy 🤣

PR #803, opened 2020-09-05
Setup grpc-gateway and register grpc routes

akhilkumarpilli:
**Changes:**   
   - Update SDK version to `v0.34.4-0.20200928165114-e17dd4fd5008`
   - Add gRPC gateway
   - Add swagger generation for gRPC gateway routes
   - Change *Filters.State from `int` to `string`
   - Add proto-breaking check

Fixes: #810


x/deployment/client/grpc_rest_test.go
```diff
@@ -0,0 +1,271 @@
...
+	)
+	s.Require().NoError(err)
+
+	s.Require().NoError(s.network.WaitForNextBlock())
+
+	// get deployment
+	resp, err := cli.QueryDeploymentsExec(val.ClientCtx.WithOutputFormat("json"))
+	s.Require().NoError(err)
+
+	var out *types.QueryDeploymentsResponse = &types.QueryDeploymentsResponse{}
```
boz, line 63: mind using the `out := &types.QueryDeploymentsResponse{}` form for all of these from now on?

x/deployment/keeper/grpc_query_test.go
```diff
@@ -149,15 +149,15 @@ func TestGRPCQueryDeployments(t *testing.T) {
 				req = &types.QueryDeploymentsRequest{
 					Filters: types.DeploymentFilters{
 						DSeq:  37,
-						State: types.Deployment_State(4),
+						State: "closed",
```
boz, line 152: why did you change the type to a string?

boz (changes requested):
Looks great.  Is there any way we can not check in the swagger-generated files?

x/deployment/keeper/grpc_query_test.go
akhilkumarpilli: In cli, state filter is taken as `string` and converted to `int`. But we can't do this conversion in grpc rest as there is no way to do it. To solve this, we changed filter type to `string` and converting state value to int in grpc querier instead of in cli.

boz: thanks.  you mean in grpc-gateway?  can you show me where?  Ideally the api would accept proto generated types only.

akhilkumarpilli: Yes, api would accept proto generated types. That's the reason I changed `State` in `DeploymentFilters` to string [here](https://github.com/ovrclk/akash/pull/803/files#diff-4fbe7df455aab18e9713e6819c2adf53R149),

And I am converting state filter to int [here](https://github.com/ovrclk/akash/pull/803/files#diff-80d646c61a94fd1837b53588833bb71bR28-R34) instead of doing it in cli [previously](https://github.com/ovrclk/akash/pull/803/files#diff-dcdc703455bc99749c6ba962ddb6d7edL51-L56)

boz: what is grpc rest?

akhilkumarpilli: grpc rest is the rest client of grpc which uses grpc gateway

anilCSE: @boz gRPC was using the original deployment filter types (i.e., `int` values for `status` filters) but that doesn't look like a better UX as gRPC users get to supply this `int` value for filters. For CLI queries, we are taking `string` as the request input and converting it back to `int`. Considering that as a better UX, we thought of introducing the same in gRPC queries as well. 
We had two solutions,
1) Introduce an interceptor/middleware for gRPC to handle this conversion
2) Change gRPC query logic directly to handle it

As CLI is also using gRPC's `queryClient` for queries, adding an interceptor just for gRPC or gRPC gateway requests didn't sound like a plan. We discussed internally and thought changing gRPC filters to consider `string` as input for filters and converting them back to original types as needed. With this, all the clients (CLI, gRPC, gRPC gateway) can use `string` for filters and all the casting logic takes place in gRPC queryClient.

We can replace this hard coded string value with some constant like `DEPLOYMENT_STATUS_CLOSED` or something, but this logic lgtm

client/grpc-gateway/config.json
```diff
@@ -0,0 +1,21 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "title": "AKASH - GRPC Gateway",
+    "version": "1.0.0"
+  },
+  "apis": [
```
anilCSE, line 7: Shall we include all `cosmos` APIs in the same swagger? I suggest we should.

Makefile
```diff
@@ -349,8 +352,36 @@ $(PROTOC):
...
+proto-tools: cache-setup $(BUF) $(PROTOC) grpc-gateway protoc-swagger
 
 tools-clean:
 	rm -rf $(CACHE)
+
+proto-swagger-gen:
+	./script/protoc-swagger-gen.sh
+
+update-swagger-docs:
+	statik -src=client/grpc-gateway -dest=client/grpc-gateway -f -m
```
anilCSE, line 381: lets use `client/docs` for docs.
```suggestion
	statik -src=client/ -dest=client/docs -f -m
```
I am still working on a PR upstream to fix swagger API registration.

Makefile
```diff
@@ -250,13 +250,16 @@ CACHE          := $(CACHE_BASE)
 CACHE_BIN      := $(CACHE)/bin
 CACHE_INCLUDE  := $(CACHE)/include
 BUF_VERSION    ?= 0.20.5
-PROTOC_VERSION ?= 3.11.2
+PROTOC_VERSION ?= 3.13.0
+PROTOC_GRPC_GATEWAY_VERSION ?= 1.14.7
```
anilCSE, line 254: +1

app/export.go
```diff
@@ -21,23 +21,33 @@ import (
...
 
 	// as if they could withdraw from the start of the next block
 	ctx := app.NewContext(true, tmproto.Header{Height: app.LastBlockHeight()})
 
+	// We export at last height + 1, because that's the height at which
+	// Tendermint will start InitChain.
+	height := app.LastBlockHeight() + 1
+
 	if forZeroHeight {
+		height = 0
```
anilCSE, line 34: This looks bad but yeah that's how it is

client/grpc-gateway/config.json
```diff
@@ -0,0 +1,21 @@
...
+      "url": "./akash/market/v1beta1/query.swagger.json"
+    },
+    {
+      "url": "./akash/provider/v1beta1/query.swagger.json"
+    },
+    {
+      "url": "./client/grpc-gateway/swagger-sdk.json"
+    }
+  ]
+}
```
anilCSE, line 21:
```suggestion
}

```

go.mod
```diff
@@ -6,23 +6,24 @@ require (
...
 	github.com/go-kit/kit v0.10.0
 	github.com/gogo/protobuf v1.3.1
+	github.com/golang/protobuf v1.4.2
 	github.com/golang/snappy v0.0.2-0.20190904063534-ff6b7dc882cf // indirect
 	github.com/google/uuid v1.1.2-0.20190416172445-c2e93f3ae59f // indirect
 	github.com/gorilla/context v1.1.1
 	github.com/gorilla/mux v1.8.0
 	github.com/gorilla/websocket v1.4.2
-	github.com/grpc-ecosystem/grpc-gateway v1.14.7
+	github.com/grpc-ecosystem/grpc-gateway v1.14.8
```
anilCSE, line 19: Is this intentional? We are using `1.14.7` right?

proto/akash/base/v1beta1/resource.proto
```diff
@@ -1,5 +1,5 @@
 syntax = "proto3";
-package akash;
+package akash.base.v1beta1;
```
anilCSE, line 2: :tada:

proto/akash/deployment/v1beta1/query.proto
```diff
@@ -1,55 +1,62 @@
...
-    DeploymentFilters Filters = 1 [(gogoproto.nullable) = false];
+    DeploymentFilters filters = 1 [(gogoproto.nullable) = false];
 
     cosmos.base.query.v1beta1.PageRequest pagination = 2;
 }
 
 // QueryDeploymentsResponse is response type for the Query/Deployments RPC method
 message QueryDeploymentsResponse{
-    repeated DeploymentResponse Deployments = 1 [(gogoproto.nullable) = false,(gogoproto.castrepeated)="DeploymentResponses"];
+    repeated DeploymentResponse deployments = 1 [(gogoproto.nullable) = false,(gogoproto.castrepeated)="DeploymentResponses"];
```
anilCSE, line 39:
```suggestion
    repeated DeploymentResponse deployments = 1 [(gogoproto.nullable) = false, (gogoproto.castrepeated)="DeploymentResponses"];
```

anilCSE: Let's add a formatting tool for proto files. `clang-format` is the best out there, that can be separate PR

script/protoc-swagger-gen.sh
```diff
@@ -0,0 +1,25 @@
...
+    -I "third_party/proto" \
+    "$query_file" \
+    --swagger_out=logtostderr=true,stderrthreshold=1000,fqn_for_swagger_name=true,simple_operation_ids=true:.
+  fi
+done
+
+# combine swagger files
+# uses nodejs package `swagger-combine`.
+# all the individual swagger files need to be configured in `config.json` for merging
+swagger-combine ./client/grpc-gateway/config.json -o ./client/grpc-gateway/swagger.json --continueOnConflictingPaths true --includeDefinitions true
```
anilCSE, line 22:
```suggestion
swagger-combine ./client/docs/config.json -o ./client/docs/swagger.json --continueOnConflictingPaths true --includeDefinitions true
```

script/protocgen.sh
```diff
@@ -12,6 +12,14 @@ for dir in $proto_dirs; do
   --gocosmos_out=plugins=interfacetype+grpc,\
 Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:. \
   $(find "${dir}" -maxdepth 1 -name '*.proto')
+
+  # command to generate gRPC gateway (*.pb.gw.go in respective modules) files
+  .cache/bin/protoc \
```
anilCSE, line 17: Why are we using it from `.cache`?
```suggestion
  protoc \
```

x/deployment/client/cli/flags.go
```diff
@@ -80,34 +80,33 @@ func AddDeploymentFilterFlags(flags *pflag.FlagSet) {
 }
 
 // DepFiltersFromFlags returns DeploymentFilters with given flags and error if occurred
-func DepFiltersFromFlags(flags *pflag.FlagSet) (types.DeploymentFilters, string, error) {
+func DepFiltersFromFlags(flags *pflag.FlagSet) (types.DeploymentFilters, error) {
```
anilCSE, line 83: +1

x/provider/module.go
```diff
@@ -78,7 +79,8 @@ func (AppModuleBasic) RegisterRESTRoutes(clientCtx client.Context, rtr *mux.Rout
 }
 
 // RegisterGRPCRoutes registers the gRPC Gateway routes for the provider module.
-func (AppModuleBasic) RegisterGRPCRoutes(_ client.Context, _ *runtime.ServeMux) {
+func (AppModuleBasic) RegisterGRPCRoutes(clientCtx client.Context, mux *runtime.ServeMux) {
+	types.RegisterQueryHandlerClient(context.Background(), mux, types.NewQueryClient(clientCtx))
```
anilCSE, line 83: :tada:

go.mod
akhilkumarpilli: No, using `v1.14.8` in upstream and it got automatically updated based on upstream version.

Makefile
```diff
@@ -349,8 +352,36 @@ $(PROTOC):
 	unzip -oq ${PROTOC_ZIP} -d $(CACHE) 'include/*'; \
 	rm -f ${PROTOC_ZIP})
 
+grpc-gateway:
+	@echo "Installing protoc-gen-grpc-gateway..."
+	@rm -f $@
+	@curl -o "${GOBIN}/protoc-gen-grpc-gateway" -L \
```
boz, line 358: Let's install stuff into `$(CACHE_BIN)` now.

Makefile
```diff
@@ -349,8 +352,36 @@ $(PROTOC):
...
+	"https://github.com/grpc-ecosystem/grpc-gateway/releases/download/v${PROTOC_GRPC_GATEWAY_VERSION}/${PROTOC_GRPC_GATEWAY_BIN}"
+	chmod +x "${GOBIN}/protoc-gen-grpc-gateway"
+
+protoc-swagger:
+ifeq (, $(shell which protoc-gen-swagger))
+	@echo "Installing protoc-gen-swagger..."
+	@go install github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger
```
boz, line 365: Add this to `tools.go` and install along with everything else from there in `devdeps-install`.

script/protoc-swagger-gen.sh
boz: :nail_care: use multiple lines to make this more readable please :pray:

script/protocgen.sh
boz: we're installing downloaded binaries into `.cache/bin` now.

boz (commented):
Looks great, thanks @akhilkumarpilli .  Again if there's a way to not commit the intermediate json files, that'd be really great.  Same with the third-party .proto files.

Also, let's not use the `@` prefix in the `Makefile` anymore, unless it's for an `@echo "some message"`.

Makefile
akhilkumarpilli: Yeah I tried it, but `protoc-gen-grpc-gateway` will be ran by `protoc` internally and we cannot set path for it. It's generally looking from installed binaries. That's the reason, we need to install `protoc-gen-grpc-gateway` globallly

boz: I see.  Okay, thanks for trying.

Can you try modifying path in the `Makefile`?  If it doesn't work no problem.

```
export PATH = $(PATH):$(CACHE_BIN)
```

Makefile
```diff
@@ -289,6 +295,7 @@ TM_LIBS				= third_party/proto/tendermint/libs/bits
 
 GOGO_PROTO_TYPES    = third_party/proto/gogoproto
 COSMOS_PROTO_TYPES  = third_party/proto/cosmos_proto
+GOOGLE_PROTO_TYPES  = third_party/proto/google/api
```
troian, line 298: Please rebase this PR to latest master as we no longer use third_party dir and HTTP to download proto files, instead use modvendor to sync them into vendor dir

akhilkumarpilli: Proto files are not generating in vendor directory using modvendor. I tried by rebasing to master and proto files are not generating. Even tested in master branch directly and facing same issue there too.

boz: does #818 help?

akhilkumarpilli: I didn't notice this commit. Just now tested in master and its working. Will try same by rebasing this PR to master. Thanks @boz

akhilkumarpilli: @troian `github.com/googleapis/googleapis/google/api` is a non go repository. So we cannot use `go mod vendor` and sync proto files here. Is there any other way we can solve this?

troian: @akhilkumarpilli this way should be ok https://github.com/ovrclk/akash/pull/819

boz:
bump

troian:
@akhilkumarpilli plz rebase it to master

boz (commented):
This looks like it has a lot in common with #828 and/or #831.  Which one would you like to merge first?

akhilkumarpilli:
#828 Need to be merged first. But we still have one issue with buf for `proto-lint` and `proto-check-breaking`. This is due to migration from third_party to vendor. We are working on it.

.github/workflows/proto.yaml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@master
+      - name: lint
+        run: make proto-lint
+  breakage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
```
troian, line 18: `actions/checkout@v2` plz

a rude way to fetch master will be in use of [this approach](https://github.com/actions/checkout#Fetch-all-history-for-all-tags-and-branches)
another option to check would be `- run: git fetch --prune --unshallow`. [example](https://github.com/ovrclk/akash/blob/ac8ca601b9d87dff1507e103a6a028ed08064906/.github/workflows/release.yaml#L18)

this should fix [this](https://github.com/ovrclk/akash/pull/803/checks?check_run_id=1224481285)

akhilkumarpilli: @troian , This approach is also throwing error. Working only with absolute git https url.

troian:
merged as #858

PR #813, opened 2020-09-12
goreleaser: enable ledger

boz:


.goreleaser.yaml
```diff
@@ -9,7 +9,7 @@ builds:
       - goos: darwin
         goarch: 386
     flags:
-      - -tags={{if eq .Env.MAINNET "true"}}mainnet{{else}}testnet{{end}}
+      - -tags=netgo,ledger,{{if eq .Env.MAINNET "true"}}mainnet{{else}}testnet{{end}}
```
Ropes, line 12: To stay consistent with our other templates lines, put in `"`?

troian (approved):

.goreleaser.yaml
boz: This was here first and... I kind of hate the `\"` :joy:

PR #815, opened 2020-09-14
tests: enable e2e

troian:


Ropes (approved):
Self :hand: , but integration tests finally passing again!

Makefile
```diff
@@ -158,14 +158,12 @@ test-integration: $(BINS)
...
 	go test -mod=readonly -p 4 -tags "integration $(BUILD_MAINNET)" -v ./integration/...
 
-test-e2e-integration: $(BINS)
+test-e2e-integration:
 	# ASSUMES:
-	# 1. cluster created - `kind create cluster --config=_run/kube/kind-config.yaml`
-	# 2. cluster setup - `make -s -C _run/kube kind-ingress-setup`
-	cp akashctl akashd ./_build
-	$(KIND_VARS) go test -mod=readonly -p 4 -tags "e2e integration $(BUILD_MAINNET)" -v ./integration/... -run TestE2EApp
+	# 1. cluster created - `make -s -C _run/kube kind-cluster-create` or for CI:`kind create cluster --config=_run/kube/kind-config.yaml` + bootstrapping metrics-server
```
boz, line 163: :nail_care: long line.  there's only one assumption, no need for numbered list.

provider/cmd/run.go
```diff
@@ -79,7 +81,15 @@ func runCmd() *cobra.Command {
 // doRunCmd initializes all of the Provider functionality, hangs, and awaits shutdown signals.
 func doRunCmd(ctx context.Context, cmd *cobra.Command, _ []string) error {
 	cctx := sdkclient.GetClientContextFromCmd(cmd)
-	cctx, err := sdkclient.ReadTxCommandFlags(cctx, cmd.Flags())
+
+	flagSet := cmd.Flags()
+	from, _ := flagSet.GetString(flags.FlagFrom)
```
boz, line 86: just use `cmd.Flags().GetString(flags.FlagFrom)`

provider/cmd/test_helpers.go
```diff
@@ -0,0 +1,33 @@
+package cmd
```
boz, line 1: put this in a file that ends in `_test.go` so that it's only included when testing.  or, ideally, in a separate test util package.

boz (commented):
looks great

provider/cmd/test_helpers.go
Ropes: That is a much better pattern. :rocket:

PR #818, opened 2020-09-16
fix: modvendor not copying cosmos-sdk .proto files

### LINKED ISSUE #817: modvendor not copying cosmos-sdk .proto files
### Issue type
Bug

### Akash version
This bug is on the build of the latest master checkout

### Akash token balance
N/A

### What happened
.proto files are not copied into place with a fresh GOPATH checkout

### How to reproduce it
In the makefile, the fix is to remove the single quotes in the protovendor target where it concatenates the -include directories:

protovendor: modsensure $(MODVENDOR)
        @echo "vendoring *.proto files..."
        $(MODVENDOR) -copy="**/*.proto" -include='\
github.com/cosmos/cosmos-sdk/proto,\
github.com/tendermint/tendermint/proto,\
github.com/gogo/protobuf,\
github.com/regen-network/cosmos-proto/cosmos.proto'

What appears to have been happening is the tool was pulling the correct files only if there were already generated .pb.go files present.  In a fresh cosmos-sdk checkout, the cosmos-sdk/proto directory contains only .proto files.  The modvendor source code was splitting the string but because of the initial single quote which got passed to the tool the cosmos-sdk proto directory was not being copied.

### Environment
Ubuntu Linux

troian:
fixes #817

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -262,11 +262,11 @@ proto-check-breaking: $(BUF)
 .PHONY: protovendor
 protovendor: modsensure $(MODVENDOR)
 	@echo "vendoring *.proto files..."
-	$(MODVENDOR) -copy="**/*.proto" -include='\
+	$(MODVENDOR) -copy="**/*.proto" -include=\
 github.com/cosmos/cosmos-sdk/proto,\
 github.com/tendermint/tendermint/proto,\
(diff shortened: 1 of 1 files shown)


boz (approved):
:rocket:

PR #825, opened 2020-09-21
ci: fix crosscompile issue backport from master

troian:
ledger uses github.com/zondax/hid which includes some C files
in order to get ledger detected akash has to deal with CGO
and crosscompilation

fix is using Goreleaser packed in
https://github.com/gythialy/golang-cross Docker image
with bunch of toolchains for crosscompilation

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -11,7 +11,7 @@ on:
 env:
   # bleh
   MAINNET: ${{ github.ref == 'refs/heads/mainnet/main' || (startsWith(github.ref,'refs/tags/v') && ( contains(github.ref,'.0.') || contains(github.ref,'.2.') || contains(github.ref,'.4.') || contains(github.ref,'.6.') || contains(github.ref,'.8.') )) }}
-
+  GOLANG_CROSS_VERSION: v1.15.2
 jobs:
   goreleaser:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


boz (approved):
:rocket: :pray:

.github/workflows/release.yaml
```diff
@@ -11,29 +11,27 @@ on:
 env:
   # bleh
   MAINNET: ${{ github.ref == 'refs/heads/mainnet/main' || (startsWith(github.ref,'refs/tags/v') && ( contains(github.ref,'.0.') || contains(github.ref,'.2.') || contains(github.ref,'.4.') || contains(github.ref,'.6.') || contains(github.ref,'.8.') )) }}
```
boz, line 13: You know, on `mainnet/main` I hastily flipped the `BUILD_MAINNET` logic in the `Makefile`.  It now defaults to mainnet unless `MAINNET=false`.  We should probably keep this in sync with that at some point.  Not necessary now, just thought of it.

boz (approved):
:ship:

.github/workflows/release.yaml
troian: oh, i forgot that. this should turn into make target, otherwise won't peek MAINNET
```yaml
        run: |-
          docker run \
          --priviledged \
          -e MAINNET=${MAINNET} \
          -e DOCKER_USERNAME=${{secrets.DOCKERHUB_USERNAME}} \
          -e DOCKER_PASSWORD=${{secrets.DOCKERHUB_PASSWORD}} \
          -e GITHUB_TOKEN=${{ secrets.GORELEASER_ACCESS_TOKEN }} \
          -v /var/run/docker.sock:/var/run/docker.sock \
          -v `pwd`:/go/src/github.com/ovrclk/akash \
          goreng/golang-cross:${GOLANG_CROSS_VERSION} \
          release --rm-dist
```

boz: wait, actually, are we using the makefile with golang-cross?

what is important is that the build tag `mainnet` is set when compiling for the binary release.

boz (approved):
:rocket:

boz (approved):
:pray: :pray: :pray: :pray:

PR #828, opened 2020-09-30
refactor: change address from bytes to bech32 string

akhilkumarpilli:


integration/e2e_test.go
```diff
@@ -78,7 +78,7 @@ func (s *IntegrationTestSuite) TearDownSuite() {
 	_, err = deploycli.TxCloseDeploymentExec(
 		val.ClientCtx,
 		keyTenant.GetAddress(),
-		fmt.Sprintf("--owner=%s", createdDep.Groups[0].GroupID.Owner.String()),
+		fmt.Sprintf("--owner=%s", createdDep.Groups[0].GroupID.Owner),
```
boz, line 81: :+1:

sdkutil/object_id.go
```diff
@@ -104,21 +107,25 @@ func ReflectBlockID(id interface{}) BlockID {
 				}
 			}
 		case "Provider":
-			if field.Type() == reflect.TypeOf(sdk.AccAddress{}) {
-				x, ok := field.Interface().(sdk.AccAddress)
-				if ok && len(x) == sdk.AddrLen {
-					b.Provider = &x
+			if field.Type() == reflect.TypeOf(string("")) {
```
boz, line 110: I think we'll have to take a look to see if this code makes sense anymore.

x/deployment/types/msgs.go
```diff
@@ -37,7 +37,12 @@ func (msg MsgCreateDeployment) GetSignBytes() []byte {
 
 // GetSigners defines whose signature is required
 func (msg MsgCreateDeployment) GetSigners() []sdk.AccAddress {
-	return []sdk.AccAddress{msg.ID.Owner}
+	owner, err := sdk.AccAddressFromBech32(msg.ID.Owner)
```
boz, line 40: Can you validate all addresses in `ValidateBasic()`?

x/deployment/types/types.go
```diff
@@ -17,59 +17,11 @@ const DefaultOrderBiddingDuration = int64(12342)
...
-// 	"active": DeploymentActive,
-// 	"closed": DeploymentClosed,
-// }
-
-// Deployment stores deploymentID, state and version details
-// type Deployment struct {
-// 	DeploymentID `json:"id"`
-// 	State        DeploymentState `json:"state"`
-// 	Version      []byte          `json:"version"`
-// }
```
boz, line 43: :pray:

x/market/client/cli/cli_test.go
```diff
@@ -222,12 +222,12 @@ func (s *IntegrationTestSuite) Test2CreateBid() {
 	resp, err = cli.QueryBidsExec(val.ClientCtx.WithOutputFormat("json"))
 	s.Require().NoError(err)
 
-	var bidRes *types.QueryBidsResponse = &types.QueryBidsResponse{}
+	bidRes := &types.QueryBidsResponse{}
```
boz, line 225: :pray:

boz (commented):
Thanks @akhilkumarpilli .  Some comments inline.  Can you rebase with master so we can queue this up to merge?

sdkutil/object_id.go
boz: @akhilkumarpilli I have replaced these with a simpler method in master.

x/deployment/types/msgs.go
troian: @boz @akhilkumarpilli, as we doing `msg.ID.Validate()` (which checks errors from sdk.AccAddressFromBech32) in all `ValidateBasic` can remove panic then

troian:
@akhilkumarpilli there are a bunch of conflicts unfortunately 😿

troian (approved):

boz (approved):
:+1:

x/deployment/keeper/key.go
```diff
@@ -17,7 +17,7 @@ var (
 
 func deploymentKey(id types.DeploymentID) []byte {
 	buf := bytes.NewBuffer(deploymentPrefix)
-	buf.Write(id.Owner.Bytes())
+	buf.Write([]byte(id.Owner))
```
boz, line 20: This really increases the size of our indexes.

We should look at using `AccAddress` instead.  It's not necessary for this PR tho.

boz (approved):
Looks good @akhilkumarpilli .  I imagine that the proto check failure is expected?

I do wonder - why did we change all of the fields in IDs to strings tho?

boz:
I'd like to understand the motivation for this before merging it.  There are many benefits to having typed values that we lose when everything is turned into a string.

boz (approved):
Spoke with @akhilkumarpilli  - this is needed so that json requests to grpc-gateway can use bech strings instead of bytes.

Can look into another way of doing it later.

PR #831, opened 2020-10-01
feat: register swagger api

akhilkumarpilli:
**Changes**:
* Add swagger-ui files in `client/docs/swagger-ui`
* Register swagger api in `app.go`


boz:
@akhilkumarpilli are there issues with rebasing with master for this?  any other blockers?

client/docs/config.json
```diff
@@ -1,21 +1,25 @@
...
-      "url": "./akash/deployment/v1beta1/query.swagger.json"
+      "url": "./tmp-swagger-gen/akash/deployment/v1beta1/query.swagger.json"
     },
     {
-      "url": "./akash/market/v1beta1/query.swagger.json"
+      "url": "./tmp-swagger-gen/akash/market/v1beta1/query.swagger.json"
     },
     {
-      "url": "./akash/provider/v1beta1/query.swagger.json"
+      "url": "./tmp-swagger-gen/akash/provider/v1beta1/query.swagger.json"
```
boz, line 16: what is the `tmp-` prefix about?  can we not check all of these artifacts in?  if we're packing it with statik I'd prefer to only check in the packed version.

akhilkumarpilli: This is just a temporary folder in which we generate swagger docs of each module. We will combine all those swagger json files into one doc using this config.json. Once after combining all, we will delete this temp folder. 

[This](https://github.com/ovrclk/akash/pull/831/files#diff-11f7032d70c63ebaa71cf65c25ec2a66f4fba95d9a4472caaee7829dba417a82) script creates temp folder, combines all docs and delete temp folder.

boz: I see, thank you

boz (approved):
It'd be nice to use `statik` to pack the files into a single go file so that we don't have to commit all of the source files, but we can try that later.

This was a tough one to get in, thanks for finishing it up.

akhilkumarpilli:
Yeah, we are using `statik` to combine all files into single `statik.go` and we are serving this file for swagger docs route. We are committing just swagger-ui files here as we need them to update statik.go if there will be any modifications.

akhilkumarpilli:
We are using swagger-ui files which are downloaded from official swagger tools.

boz:
Right, but can you have the script tag reference a cdn instead of a local copy?

akhilkumarpilli:
@boz , removed swagger-ui files and used cdn urls instead

boz (approved):
:rocket:

troian (approved):

PR #833, opened 2020-10-01
feat (provider): Kubernetes Network Policies applied to tenant Namespaces

Ropes:
This change adds Network Policies for Hardened Multi-Tenancy to Tenant Namespaces. Specifically, this change affects Ingress into Network Policies.

## Ingress Rules
* Deny all Ingress into Tenant namespace. This is the base for providing controlled access to the Namespace.
* Allow Ingress traffic from containers withing the Tenant's Namespace.(Inter container traffic)
* Allow Ingress traffic from Ingress Controller(`ingress-nginx` for now, but should be made to reference generic ingress controllers).
* Allow Ingress traffic from `metrics-server` which the Provider uses to determine usage of the cluster, so `metrics-server` needs access for health-checks.

### Egress Rules
* Deny All Egress from Tenant namespace by default.
* Allow Egress Traffic to other pods within the Tenant Namespace.
* Allow Egress Traffic to all addresses 0.0.0.0/0 EXCEPT: 10.0.0.0/8(Internal Subnet of Cluster).
  * TODO: Testing indicates this assertion does work as expected. Connections from the tenant namespace are able to open connections to the Kubernetes API, which is what this rule attempts to block.
    * The previous two rules are in effect, and connections to Internet are blocked until this rule is applied, however it does not exclude preventing connections to 10.0.0.0/8 as expected.
* Allow Egress Traffic `UDP Port:53` to the `coredns` deployments hosted in the `kube-system` namespace.

Basic Unit tests for the NetworkPolicy builder struct.
~E2E Test updated to use the Calico plugin.~(Calico does not play well with the CI workflow environment).


Ropes:
There seems to be a problem running the Calico CNI in Kind on GH Workflows?

boz:
@Ropes what's the status on this?  It looks like `TestLidNsSanity` is failing, not anything specific to calico configuration.

Ropes:
Updated docs and code tidying, ready for review :pray:

provider/cluster/kube/builder.go
```diff
@@ -243,6 +338,231 @@ func (b *serviceBuilder) ports() []corev1.ServicePort {
...
+				PolicyTypes: []netv1.PolicyType{
+					netv1.PolicyTypeEgress,
+				},
+			},
+		},
+	}, nil
+}
+
+// Update a single NetworkPolicy with correct labels.
+func (b *netPolBuilder) update(obj *netv1.NetworkPolicy) (*netv1.NetworkPolicy, error) { // nolint:golint,unparam
```
boz, line 561: is the nolint still needed?  looks like all params are used.

boz (approved):
:+1:

provider/cluster/kube/builder.go
Ropes: It's required because the methods always return nil to implement the interface.

PR #835, opened 2020-10-05
Fix buildfile issues around Docker, switch to Golang base image

hydrogen18:
Changes

1. Fix a typo, I think you meant `mod tidy` not `mod tid` here
2. Switch to Golang as base image to fix issue around dynamic libraries like `librt`. Tried alpine but Go binaries on alpine have other issues. My locally built container can run the `akash` binary after this change.
3. Pass the `-f` command to all the invocations of the golang-cross container. Not 100% sure on this one but `release-dry-run` was failing for me.

### DIFF
--- a/make/releasing.mk
+++ b/make/releasing.mk
@@ -41,7 +41,7 @@ release-dry-run: modvendor
 		-v `pwd`:/go/src/github.com/ovrclk/akash \
 		-w /go/src/github.com/ovrclk/akash \
 		troian/golang-cross:${GOLANG_CROSS_VERSION} \
-		--rm-dist --skip-validate --skip-publish
+		-f .goreleaser-docker.yaml --rm-dist --skip-validate --skip-publish
 
 .PHONY: release
 release: modvendor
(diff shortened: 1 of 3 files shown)


make/releasing.mk
```diff
@@ -41,7 +41,7 @@ release-dry-run: modvendor
 		-v `pwd`:/go/src/github.com/ovrclk/akash \
 		-w /go/src/github.com/ovrclk/akash \
 		troian/golang-cross:${GOLANG_CROSS_VERSION} \
-		--rm-dist --skip-validate --skip-publish
+		-f .goreleaser-docker.yaml --rm-dist --skip-validate --skip-publish
```
troian, line 44: `release-dry-run` and `release` targets should use default goreleaser config `.goreleaser.yaml`

make/releasing.mk
```diff
@@ -60,4 +60,4 @@ release: modvendor
 		-v `pwd`:/go/src/github.com/ovrclk/akash \
 		-w /go/src/github.com/ovrclk/akash \
 		troian/golang-cross:${GOLANG_CROSS_VERSION} \
-		release --rm-dist
+		-f .goreleaser-docker.yaml release --rm-dist
```
troian, line 63: ☝️

troian (approved):

PR #837, opened 2020-10-06
fix: build docker image on top of debian:buster

troian:


hydrogen18:
I've confirmed that the Dockerfile is now only 193MB after building locally and I was able to launch a docker container and run `/bin/akash`

Ropes (approved):
Tested locally, image size down to 193MB! :raised_hands:  :pray:

PR #841, opened 2020-10-08
refactor: include external proto from cosmos/third_party

troian:
fix .clang-format config
install clang-format from default homebrew keg


troian:
@akhilkumarpilli clang-format and proto-lint are working now but proto files needs some tuning to pass linter

boz (approved):
:+1:

PR #843, opened 2020-10-08
Use docker to run clang-format and buf tools

akhilkumarpilli:
**Changes**:
* Remove `buf` and `clang-format` install scripts
* Use docker for `proto-format`, `proto-lint`  and `proto-check-breaking`


make/proto.mk
```diff
@@ -11,11 +11,14 @@ ifeq ($(UNAME_OS),Darwin)
...
-proto-lint: $(BUF)
-	$(BUF) check lint --error-format=json
+proto-lint: 
+	$(DOCKER_BUF) check lint --error-format=json
+.PHONY: proto-lint
 
-proto-check-breaking: $(BUF)
-	$(BUF) check breaking --against-input '.git#branch=master'
+proto-check-breaking:
+	$(DOCKER_BUF) check breaking --against-input $(HTTPS_GIT)#branch=master
```
troian, line 19: i think HTTPS is unnecessary. we can compare it against local copy

akhilkumarpilli: Actually, github workflow is failing when using just `.git`. Please check [this](https://github.com/ovrclk/akash/pull/803/checks?check_run_id=1224481285) once for reference

boz (approved):
:100:

make/proto.mk
troian: looks like it needs actions/checkout to pull `master` too

akhilkumarpilli: I already included `actions/checkout@master` in workflow. Will try to check with `.git` once again. But this PR can be merged as issue is with `proto.yml` workflow which is not yet included in master.

troian: see if [this works](https://github.com/ovrclk/akash/pull/803#discussion_r501853316)

troian (approved):

PR #844, opened 2020-10-08
refactor: set *all* to be executed as default target

troian:


make/lint.mk
```diff
@@ -17,11 +17,14 @@ SUBLINTERS = deadcode \
...
 test-sublinter-misspell:
 	$(LINT) misspell --no-config
 
+.PHONY: test-sublinter-ineffassign
 test-sublinter-ineffassign:
 	$(LINT) ineffassign --no-config
 
+.PHONY: test-sublinter-%
 test-sublinter-%:
 	$(LINT) "$(@:test-sublinter-%=%)"
```
boz, line 30: I saw something recently that made it look like we can use `$*` instead of `$(@:test-sublinter-%=%)`

boz (approved):
:rocket:

make/lint.mk
troian: oh, thats awesome, tnx!

Issue #847, opened 2020-10-08
cmd: e2e testing for provider commands


troian:
recent migration to proto seems to break `provider status` command as `provider.Status` types and few other are not registered within deprecated amino codec.

hydrogen18:
Fixed in PR #848

Issue #845, opened 2020-10-08
"provider status" command failing with "Unregistered interface error'"


hydrogen18:
If I run `make provider-status` in the examples it invokes `make provider status` and I get a message about "Unregistered interface error'. I added some debugging code and was able to determine this comes from this code

```
 if err = cctx.PrintOutputLegacy(result); err != nil {
   return err
  }
```
It looks like the result is retrieved correctly in this command but the PrintOutputLegacy command doesn't know how to display it? 

I added some debugging code with `fmt.Printf` and it looks like it is valid

```
result of gclient.Status: &{0xc0000a2060 0xc00102e7e0 0xc00102e7e4}
cluster: &{0 {[] [{units:<val:"100" >  quantity:<val:"16777216" >  quantity:<val:"134217728" > }] [{units:<val:"7806" > attributes:<key:"arch" value:"amd64" >  quantity:<val:"14789169152" >  quantity:<val:"269167448064" > }] <nil>}}
bidengine: &{0}
manifest: &{1}
```

So the fix for this should just be to find a better way of printing this.

hydrogen18:
Need to investigate replacing this with `PrintOutput` if possible.

hydrogen18:
Example output after converting to JSON

```
{
  "cluster": {
    "leases": 0,
    "inventory": {
      "active": null,
      "pending": null,
      "available": [
        {
          "cpu": {
            "units": {
              "val": "7808"
            },
            "attributes": [
              {
                "key": "arch",
                "value": "amd64"
              }
            ]
          },
          "memory": {
            "size": {
              "val": "15585153024"
            }
          },
          "storage": {
            "size": {
              "val": "269167448064"
            }
          }
        }
      ],
      "error": null
    }
  },
  "bidengine": {
    "orders": 0
  },
  "manifest": {
    "deployments": 0
  }
}


```

boz:
This is fixed, right?

PR #848, opened 2020-10-09
Switch 'provider status' command to use JSON to dump data

hydrogen18:


provider/cmd/status.go
```diff
@@ -50,9 +53,21 @@ func doStatus(cmd *cobra.Command) error {
...
+	if err != nil {
+		return err
+	}
+	_, err = buf.WriteRune('\n')
+	if err != nil {
 		return err
 	}
 
-	return nil
+	_, err = fmt.Print(buf.String())
```
boz, line 71: you can use `os.Stdout.Write(buf)` to avoid converting to a string.

boz (commented):
Looks great!

Can you make a utility function for this in `provider/cmd`?  This way it can be used by other commands.

provider/cmd/status.go
hydrogen18: Good point, change made

boz (approved):
:rocket: :ship:

It'd be good to grep around and replace other uses of `PrintOutputLegacy` next (if they exist)

hydrogen18:
It looks like it is also used in these places

```
./provider/cmd/leaseStatus.go:60:	if err = cctx.PrintOutputLegacy(result); err != nil {
./provider/cmd/serviceLogs.go:103:			if err = cctx.PrintOutputLegacy(msg); err != nil {
./provider/cmd/serviceStatus.go:65:	if err = cctx.PrintOutputLegacy(result); err != nil {
./events/cmd/root.go:66:				if err := cctx.PrintOutputLegacy(ev); err != nil {
```

I'll see about refactoring them.

boz:
Ok awesome @hydrogen18 .  I think that since there's one in `cmd/akash` that the utility should probably be moved to `cmd/common`

troian (approved):

hydrogen18:
fixes #847

hydrogen18:
fixes #845

Issue #838, opened 2020-10-06
Fix configuration inconsistency in app &  Kubernetes depoyment in '_run/single'


hydrogen18:
We did some investigation and found a few things

1. Once the akashd pod is running the container is up but you can't get to it, you just get a 502 Bad gateway response
2. The akash application is listening on `tcp://localhost:26657` not on the wildcard address
3. Tendermint is creating a default `config.toml` file since we aren't providing one in the container
4. Even specifying the command line flag for the TCP port for RPC doesn't seem to override it, this should be fixed since command line switches should always win.

We need to make our configuration consistent & document how we're doing it as well as recommended methods.

hydrogen18:
Here's the result of my investigation on this so far

1. I played with setting environmental variables like `AKASH_RPC_LADDR` on my local machine and they don't do anything that I can see with. I think we may have to call `viper.SetEnvKeyReplacer` to get it to replace the dot character with underscores when checking the environmental variables for configuration
2. There appears to be a bug in cosmos SDK when loading the configuration for tendermint, explained below

In `cmd /akash/cmd/root.go` we install a `PersistenPreRunE` function into cobra which gets executed and calls `server.InterceptConfigsPreRunHandler` which is expected from the cosmos SDK. That function is defined in this file 

https://github.com/cosmos/cosmos-sdk/blob/9856327d95eba4c3ae2d15d26a76ad3ac17b20b8/server/util.go#L57

That function calls a private function called `interceptConfigs` which has the following section 

```
	rootDir := rootViper.GetString(flags.FlagHome)
	configPath := filepath.Join(rootDir, "config")
	configFile := filepath.Join(configPath, "config.toml")

	conf := tmcfg.DefaultConfig()

	if _, err := os.Stat(configFile); os.IsNotExist(err) {
		tmcfg.EnsureRoot(rootDir)

		if err = conf.ValidateBasic(); err != nil {
			return nil, fmt.Errorf("error in config file: %v", err)
		}

		conf.RPC.PprofListenAddress = "localhost:6060"
		conf.P2P.RecvRate = 5120000
		conf.P2P.SendRate = 5120000
		conf.Consensus.TimeoutCommit = 5 * time.Second
		tmcfg.WriteConfigFile(configFile, conf)
	} else {
		rootViper.SetConfigType("toml")
		rootViper.SetConfigName("config")
		rootViper.AddConfigPath(configPath)
		if err := rootViper.ReadInConfig(); err != nil {
[...]

PR #862, opened 2020-10-14
Fix kustomize in the 'single' example #838

hydrogen18:
This is a bunch of changes I've made to get the example `single` working. I have it working with my custom hacked up docker image, once Cosmos SDK merges my fix it should be possible to have a working example again.

This is a brief summary of the changes

1. Fix things referencing `akashctl` & `AKASH_CTL` since those are obsolete
2. Update the `run.sh` scripts that are part of kustomize to set the correct flags
3. Add the ping command to the Makefile for this example
4. Allow uploading a custom image via Makefile commands
5. Remove some un-needed verbs in the Makefile
6. Update the Ingress declarations to be v1

### DIFF
--- a/_docs/kustomize/akash-provider/deployment.yaml
+++ b/_docs/kustomize/akash-provider/deployment.yaml
@@ -13,7 +13,7 @@ spec:
     spec:
       containers:
         - name: akash-provider
-          image: ovrclk/akashctl:latest
+          image: ovrclk/akash:latest
           imagePullPolicy: IfNotPresent
           command: [ "/bin/sh", "/boot/run.sh" ]
           env:
@@ -26,7 +26,7 @@ spec:
             #
             # key.txt:      exported wallet key
             # key-pass.txt: password used to encrypt exported key
-            - name: AKASHCTL_BOOT_KEYS
+            - name: AKASH_BOOT_KEYS
               value: /boot-keys
 
             ##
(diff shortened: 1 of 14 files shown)


_docs/kustomize/akashd/run.sh
```diff
@@ -14,4 +14,4 @@ cp "$AKASH_BOOT_DATA/genesis.json" "$AKASH_HOME/config/"
 cp "$AKASH_BOOT_KEYS/node_key.json" "$AKASH_HOME/config/"
 cp "$AKASH_BOOT_KEYS/priv_validator_key.json" "$AKASH_HOME/config/"
 
-/akash start
+/akash --home=$AKASH_HOME start
```
troian, line 17: trailing whitespace

_docs/kustomize/akashd/service.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: akash
 spec:
   selector:
-    app: akash
+    app: akashd
```
troian, line 7: should it be `akash` ?

_docs/kustomize/akashd/ingress.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
```
troian, line 1: 👍

_docs/kustomize/akashd/service.yaml
hydrogen18: I think we can name it whatever we like here. The other `service.yaml` for `akash-provider` is named `akash-provider` so I am trying to be consistent with names.

_docs/kustomize/akash-provider/run.sh
```diff
@@ -1,24 +1,31 @@
 #!/bin/sh
 
-set -e
+set -xe
```
Ropes, line 3: Do we need to keep `-x` here? Usually makes CI more verbose than necessary.

_docs/kustomize/akash-provider/deployment.yaml
```diff
@@ -26,50 +26,50 @@ spec:
             #
             # key.txt:      exported wallet key
             # key-pass.txt: password used to encrypt exported key
-            - name: AKASHCTL_BOOT_KEYS
+            - name: AKASH_BOOT_KEYS
```
Ropes, line 29: hooray for naming sanity!

_docs/kustomize/akashd/service.yaml
Ropes: I believe this is correct, it's the Selector, so it needs to match the `akashd` deployments metadata label, which is `app: akashd`. So it was wrong before :grimacing:

_run/single/README.md
```diff
@@ -76,6 +76,15 @@ make clean-all kind-cluster-clean
 make init kustomize-init
 ```
 
+The next step is to configure the docker image to be used. If you want to use a specific
```
Ropes, line 79: :clap:

Ropes (approved):
Looks great! Only one minor question on bash flags.

_docs/kustomize/akashd/service.yaml
troian: oh, make sense

hydrogen18:
I pushed up a change to the `kube` example since I couldn't communicate with the deployment after it was created.

_docs/kustomize/akash-provider/run.sh
hydrogen18: Yes I can remove this.

PR #864, opened 2020-10-14
fix: `tx` commands missing --keyring-backend

Ropes:
Fixes problems related to using keys not hosted in the `os` keyring
backend. This comes up with managing keys for testnets in the
`tx` subcommands. Enables management of remote resources using
local keys.

By adding the flag, `akash` can be correctly directed to reference a
`test` keyring in directories where sets of keys are created and not
stored in the OS keyring.

### DIFF
--- a/cmd/akash/cmd/root.go
+++ b/cmd/akash/cmd/root.go
@@ -217,6 +217,8 @@ func txCmd() *cobra.Command {
 	// add modules' tx commands
 	app.ModuleBasics().AddTxCommands(cmd)
 	cmd.PersistentFlags().String(flags.FlagChainID, "", "The network chain ID")
+	cmd.PersistentFlags().String(flags.FlagKeyringBackend, flags.DefaultKeyringBackend,
+		"Select keyring's backend (os|file|kwallet|pass|test)")
 
 	return cmd
 }


cmd/akash/cmd/root.go
```diff
@@ -190,6 +190,8 @@ func queryCmd() *cobra.Command {
 
 	app.ModuleBasics().AddQueryCommands(cmd)
 	cmd.PersistentFlags().String(flags.FlagChainID, "", "The network chain ID")
+	cmd.PersistentFlags().String(flags.FlagKeyringBackend, flags.DefaultKeyringBackend,
+		"Select keyring's backend (os|file|kwallet|pass|test)")
```
boz, line 194: Why do we need either of these for query commands?

cmd/akash/cmd/root.go
```diff
@@ -217,6 +219,8 @@ func txCmd() *cobra.Command {
 	// add modules' tx commands
 	app.ModuleBasics().AddTxCommands(cmd)
 	cmd.PersistentFlags().String(flags.FlagChainID, "", "The network chain ID")
+	cmd.PersistentFlags().String(flags.FlagKeyringBackend, flags.DefaultKeyringBackend,
+		"Select keyring's backend (os|file|kwallet|pass|test)")
```
boz, line 223: This isn't added by sdk?

boz (commented):
comments inline

cmd/akash/cmd/root.go
Ropes: Good catch, I must have misattributed an error to the flag, after rerunning some of the query commands I made, it's not the keyring. Removed, and updated commit message.

cmd/akash/cmd/root.go
Ropes: Not currently, or some cobra nonsense is removing it from `tx`? Or do you mean in upstream changes?
```
akash tx -h
Transactions subcommands

Usage:
  akash tx [command]

Available Commands:
                      
                      
                      
  auth                Auth transaction subcommands
  bank                Bank transaction subcommands
  broadcast           Broadcast transactions generated offline
  crisis              Crisis transactions subcommands
  decode              Decode an binary encoded transaction string.
  deployment          Deployment transaction subcommands
  distribution        Distribution transactions subcommands
  encode              Encode transactions generated offline
  evidence            Evidence transaction subcommands
  gov                 Governance transactions subcommands
  market              Transaction subcommands
  multisign           Generate multisig signatures for transactions generated offline
  provider            Deployment transaction subcommands
  send                Send funds from one account to another. Note, the'--from' flag is
ignored as it is implied from [from_key_or_address].
  sign                Sign transactions generated offline
  sign-batch          Sign transaction batch files
  slashing            Slashing transaction subcommands
  staking             Staking transaction subcommands
  validate-signatures Validate transactions signatures
  vesting             Vesting transaction subcommands

Flags:
      --chain-id string   The network chain ID
  -h, --help              help for tx

Global Flags:
      --home string   directory for config and data (default "/home/ropes/.akash")
      --trace         print out full stack trace on errors

Additional help topics:
  akash tx upgrade       Upgrade transaction subcommands

Use "akash tx [command] --help" for more information about a command.
```

boz: not all tx commands take `--keyring-backend` - `broadcast` for instance.

these are added in individual commands like [this](https://github.com/ovrclk/akash/blob/26eb69c72b1a0a2ec2df8fc06c31b6dc361ca01a/x/deployment/client/cli/tx.go#L120)

Ropes: Alright, I'll circle back and run through the subcommands, add it only to the ones which hopefully need it.

Ropes:
After pulling master and vendoring cosmos, `tx send` has the --keyring-backend flag, it got fixed upstream :tada:

PR #865, opened 2020-10-15
Update cosmos to v0.40.0-rc0 fork

hydrogen18:
This has the following changes

1. Move us to using the tagged fork of Cosmos SDK hosted in my repo
2. Swap the 'single'  example from using a pruning strategy of `syncable` to `nothing`. It appears `syncable` has been removed
3. Update the makefile so the test-integration command works. I had to export the env. vars. and set `go test` to run just the intended test.


Makefile
```diff
@@ -4,8 +4,8 @@ APP_DIR               := ./app
 GO                    := GO111MODULE=on go
 GOBIN                 := $(shell go env GOPATH)/bin
 
-KIND_APP_IP           ?= $(shell make -sC _run/kube kind-k8s-ip)
-KIND_APP_PORT         ?= $(shell make -sC _run/kube app-http-port)
+export KIND_APP_IP           ?= $(shell make -sC _run/kube kind-k8s-ip)
+export KIND_APP_PORT         ?= $(shell make -sC _run/kube app-http-port)
```
troian, line 8: align assignment with bottom `KIND_VARS`

troian (changes requested):

hydrogen18:
So I think I am good to just get rid of the 'test-integration' target? After my "fix" it isn't doing anything different than the other targets.

Ropes:
> So I think I am good to just get rid of the 'test-integration' target? After my "fix" it isn't doing anything different than the other targets.

:+1: to remove the `make test-integration` target. It's not used by CI right now.

troian (approved):

PR #877, opened 2020-10-21
feat: circulating supply queries

akhilkumarpilli:


boz (commented):
This looks good, but I think it's misleading to say that spendable coins is the circulating supply.  Circulating supply should also include any bonded tokens that are not vesting.  It'd be nice if we could return a struct with more info like illustrated below:

```go
type  Supply struct {
  Vesting struct {
    Bonded   sdk.Coin
    Unbonded sdk.Coin
  }
  Vesting struct {
    Bonded   sdk.Coin
    Unbonded sdk.Coin
  }
}
```

anilCSE:
> This looks good, but I think it's misleading to say that spendable coins is the circulating supply. Circulating supply should also include any bonded tokens that are not vesting. It'd be nice if we could return a struct with more info like illustrated below:
> 
> ```go
> type  Supply struct {
>   Vesting struct {
>     Bonded   sdk.Coin
>     Unbonded sdk.Coin
>   }
>   Vesting struct {
>     Bonded   sdk.Coin
>     Unbonded sdk.Coin
>   }
> }
> ```

correct, we should consider bonded tokens too. @boz do you mean `Vesting` & `Available`/`VestingFree` tokens?
```go
type  Supply struct {
  Vesting struct {
    Bonded   sdk.Coin
    Unbonded sdk.Coin
  }
  Available struct {
    Bonded   sdk.Coin
    Unbonded sdk.Coin
  }
}
```

boz:
@anilCSE yes, I did.  Thanks.

anilCSE:
@boz we might need to include rewards as well right?

boz:
You mean the pool (un withdrawn)?  Yeah good point.

boz:
Tho we can already get that from other queries.

anilCSE:
> You mean the pool (un withdrawn)? Yeah good point.

Yes

anilCSE:
> Tho we can already get that from other queries.

You are right, total supply includes this and we can just use that for calculating circulating supply

akhilkumarpilli:
@boz , for calculating `Available.Bonded` tokens, we need to use staking keeper and do some extra calculations. And this value is not useful for calculating `circulating supply`. What's your point on this, do we need to calculate `Available.Bonded`?

boz:
@akhilkumarpilli we are defining "circulating supply" as the number of tokens that are not locked up by a vesting schedule.  Anil called this `Available` above.

Of the `Available` tokens, it'd be nice to know the quantity that is bonded and the quantity that is unbonded.

Does that make sense?

x/supply/query/querier.go
```diff
@@ -0,0 +1,68 @@
...
+		default:
+			return nil, sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "unknown query for endpoint: %s", path[0])
+		}
+	}
+}
+
+func queryCirculatingSupply(ctx sdk.Context, cdc *codec.Codec, accKeeper types.AccountKeeper,
+	supKeeper types.SupplyKeeper) (res []byte, err error) {
+	var supplyData Supply
+	var totalOriginal, totalVesting sdk.Coins
```
anilCSE, line 32:
```suggestion
	var totalOriginalVesting, totalCurrentVesting sdk.Coins
```

x/supply/query/querier.go
```diff
@@ -0,0 +1,68 @@
...
+
+	initialSupply := sdk.NewCoins(sdk.NewCoin("uakt", sdk.NewInt(100000000000000)))
```
anilCSE, line 34: May be we should read this from supplyKeeper itself. Can we have something like, `getSupply` at `height:1`

boz: Agree, we shouldn't hard-code this.

But I think that we shouldn't be concerned about initial supply - iterating through accounts should give us the info we need.

I do think that we'll have to look into the specific account types rather than using just using `exported.Account`, though.

x/supply/client/cli/query.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/cosmos/cosmos-sdk/client/context"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/codec"
+	"github.com/spf13/cobra"
+)
+
+// GetCirculatingSupply returns circulation supply
+func GetCirculatingSupply(cdc *codec.Codec) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "circulating",
```
anilCSE, line 16: Since we are including all the supply information, may be we should rename this to `all` or something better. @boz wdyt?

x/supply/query/types.go
```diff
@@ -0,0 +1,24 @@
...
+	Bonded   sdk.Coins `json:"bonded"`
+	Unbonded sdk.Coins `json:"unbonded"`
+}
+
+// Supply represents total coins vested, available and circulating supply
+type Supply struct {
+	Vesting     CoinDetails `json:"vesting"`
+	Available   CoinDetails `json:"available"`
+	Circulating sdk.Coins   `json:"circulating"`
+	Total	sdk.Coins `json:"total"`
```
anilCSE, line 18: Added `Total` key so that `Clients` can use single API/cmd to get all the supply information at once

x/supply/types/key.go
```diff
@@ -0,0 +1,5 @@
+package types
+
+const (
+	ModuleName = "supplyqueries"
```
anilCSE, line 4: `supplyqueries` is odd. Can we just use `supply` directly?

anilCSE, line 4: `supplyqueries` is odd. Can we just use `supply` directly?

```suggestion
	ModuleName = "supply"
```

x/supply/client/cli/query.go
anilCSE, line 16: Since we are including all the supply information, may be we should change this to `all` or something better. @boz wdyt?

```suggestion
		Use:   "all",
```

x/supply/query/types.go
```diff
@@ -0,0 +1,24 @@
...
+	Bonded   sdk.Coins `json:"bonded"`
+	Unbonded sdk.Coins `json:"unbonded"`
+}
+
+// Supply represents total coins vested, available and circulating supply
+type Supply struct {
+	Vesting     CoinDetails `json:"vesting"`
+	Available   CoinDetails `json:"available"`
+	Circulating sdk.Coins   `json:"circulating"`
+	Total       sdk.Coins   `json:"total"`
```
anilCSE, line 18: Added Total key so that Clients can use single API/cmd to get all the supply information at once

x/supply/types/key.go
anilCSE: Looks like we can't use `supply`, how about `akashsupply`?

x/supply/query/querier.go
anilCSE: Not relavant anymore

boz (approved):
Looks great!

x/supply/client/cli/query.go
anilCSE: May be `summary` or `stats`

boz: :+1: summary

boz:
you know, maybe we can release this as a self-contained repo and module - `github.com/ovrclk/cosmos-circulating-supply`.  thoughts @anilCSE ?

anilCSE:
> you know, maybe we can release this as a self-contained repo and module - `github.com/ovrclk/cosmos-circulating-supply`. thoughts @anilCSE ?

+1 for a separate, self-contained repo. Good thought @boz :tada:

troian:
@akhilkumarpilli new repo with PR is [here](https://github.com/ovrclk/cosmos-circulating-supply/pull/1). feel to edit it

anilCSE:
> @akhilkumarpilli new repo with PR is [here](https://github.com/ovrclk/cosmos-circulating-supply/pull/2) and integration is [here](https://github.com/ovrclk/akash/pull/886). feel to edit it

@troian can we rename the repo? It's not just `circulating-supply` now, may be `supply-summary` would be a better name? Also @akhilkumarpilli let's rename the command as well as discussed here: https://github.com/ovrclk/akash/pull/877#discussion_r511935594

troian:
@anilCSE sounds good to me

PR #879, opened 2020-10-21
feat(provider): add options for computing bid price fixes #868

### LINKED ISSUE #868: Bid Pricing Configuration
# Overview

Bid pricing currently consists of a multiplier on the amount of memory requested.  This needs to be extended to the other resources.

- CPU (units)
- ~Storage (class, iops)~
- ~Endpoint (shared, bandwidth caps)~

hydrogen18:
I've added two new strategies for bid pricing. The default remains unchanged.

1. A strategy which allows scaling with memory / cpu / storage. At least one of the scales must be nonzero.
2. A strategy that calls any executable (presumably a shell script, but anything should work). The script reads in a JSON description of the cpu, memory, & storage that is being bid on from stdin. It returns a single number as JSON via stdout. I probably need to document this, but where?

Other changes

1. Add lots of tests around `provider/order.go`
2. Rename the service pointer passed into `newOrder` to `svc`. I find `e` to be a very strange argument name
3. Rename struct member `order` to `orderId` in `provider/order.go`. It's the ID of the order, not the whole thing. It's confusing to have a struct named `order` with a member named `order`
4. Save the reservation to the local variable before bid pricing runs. If bid pricing fails, we need to call unreserve. 
5. Only check `o.shouldBid` if `o.bid` is nil. Otherwise, continue along. This can be set in `newOrder`
6. In pubsub, ignore any error where the subscriber is shutting down when we try and send a message to it.


provider/bidengine/order.go
```diff
@@ -150,44 +152,52 @@ loop:
 			res := result.Value().(dtypes.Group)
 			group = &res
 
-			if !o.shouldBid(group) {
+			if o.bid == nil && !o.shouldBid(group) {
 				break
```
boz, line 156: Good catch.   This part is tricky.  If `o.bid` is populated that means we restarted the process and recovered the bid from a previous execution.  We should validate that it can still be reserved, and close the bid if it can't be.

Also, if we can no longer fulfill the group (`o.shouldBid()` is false), we should cancel the original bid anyways.

So I think that the right thing to do is:

1. continue reserving space if `shouldBid()` is true.
1. exit the main loop if `shouldBid()` is false.
1. when cleaning up, cancel the bid if we exited before knowing if we won the lease.
1. when cleaning up, unconditionally unreserve space if we did not win the lease.  I think that `Unreserve` method signature can be changed to only require the order ID. 

Thoughts?

provider/bidengine/order_test.go
```diff
@@ -0,0 +1,591 @@
...
+}
+
+func (_ fauxQueryClient) Provider(ctx context.Context, in *ptypes.QueryProviderRequest, opts ...grpc.CallOption) (*ptypes.QueryProviderResponse, error) {
+	return nil, errNotImplemented
+}
+
+// Deployments queries deployments
+func (_ fauxQueryClient) Deployments(ctx context.Context, in *dtypes.QueryDeploymentsRequest, opts ...grpc.CallOption) (*dtypes.QueryDeploymentsResponse, error) {
+	return nil, errNotImplemented
+}
```
boz, line 44: If we don't have mocks for this, let's generate one.  See the pattern [here](https://github.com/ovrclk/akash/blob/494572ac32a7a48d801a27397e6d2047d80419ec/make/codegen.mk#L5-L12)

provider/bidengine/order_test.go
```diff
@@ -0,0 +1,591 @@
...
+	lastCall.resourceGroup = rg
+	lastCall.orderId = orderID
+
+	select {
+	case fc.reserveCallNotify <- 0:
+
+	default:
+	}
+
+	return fauxReservation{orderId: orderID, resources: rg}, nil
```
boz, line 192: I feel like we should just make [this](https://github.com/ovrclk/akash/blob/494572ac32a7a48d801a27397e6d2047d80419ec/provider/cluster/reservation.go#L14) public.

provider/bidengine/order_test.go
```diff
@@ -0,0 +1,591 @@
...
+		require.Nil(t, err)
+		require.NotNil(t, pricing)
+	}
+
+	var scaffold orderTestScaffold
+
+	deploymentId := dtypes.DeploymentID{
+		Owner: "",
+		DSeq:  1,
+	}
```
boz, line 226: `testutil.DeploymentID(t)`

provider/bidengine/order_test.go
```diff
@@ -0,0 +1,591 @@
...
+
+	scaffold.groupId = dtypes.MakeGroupID(deploymentId, 2)
+	scaffold.orderId = mtypes.MakeOrderID(scaffold.groupId, rand.Uint32())
+
+	// TODO - silence logs depending on env. var.
+	myLog := log.NewTMLogger(os.Stdout)
```
boz, line 232: `testutil.Logger(t)`

provider/bidengine/order_test.go
```diff
@@ -0,0 +1,591 @@
...
+	require.GreaterOrEqual(t, priceAmount, int64(1))
+	require.Less(t, priceAmount, int64(100))
+
+	leaseId := mtypes.LeaseID{
+		Owner:    "",
+		DSeq:     scaffold.groupId.DSeq,
+		GSeq:     scaffold.groupId.GSeq,
+		OSeq:     4,
+		Provider: scaffold.testAddr,
+	}
```
boz, line 351: do we have the order id here?  `MakeLeaseID` if not (please add if it's missing).

provider/bidengine/order_test.go
```diff
@@ -0,0 +1,591 @@
...
+	ev := mtypes.EventLeaseCreated{
+		Context: sdkutil.BaseModuleEvent{},
+		ID:      leaseId,
+		Price:   sdk.NewInt64Coin("akt", 1),
```
boz, line 355: `testutil.AkashCoin(t)` or `testutil.Coin(t)`.

provider/bidengine/order_test.go
```diff
@@ -0,0 +1,591 @@
...
+	}
+	require.Equal(t, order.orderId.GroupID(), ev.ID.GroupID())
+
+	err := scaffold.testBus.Publish(ev)
+	require.Nil(t, err)
```
boz, line 360: `require.NoError(t,err)` for errors.

provider/bidengine/pricing.go
```diff
@@ -1,15 +1,123 @@
...
+		storageTotal.Sign() < 0 || !storageTotal.IsInt64() {
+		return sdk.Coin{}, ErrBidQuantityInvalid
+	}
+
+	cpuCost := sdk.NewCoin(denom, sdk.NewIntFromBigInt(cpuTotal))
+	memoryCost := sdk.NewCoin(denom, sdk.NewIntFromBigInt(memoryTotal))
+	storageCost := sdk.NewCoin(denom, sdk.NewIntFromBigInt(storageTotal))
+
+	// Check for less than or equal to zero
+	cost := cpuCost.Add(memoryCost).Add(storageCost)
```
boz, line 104: do the overflow checks after this.

provider/bidengine/pricing.go
```diff
@@ -66,3 +174,105 @@ func calculatePriceRange(gspec *dtypes.GroupSpec) (sdk.Coin, sdk.Coin) {
...
+	}
+
+	result := shellScriptPricing{
+		path:         path,
+		processLimit: make(chan int, processLimit),
+	}
+
+	// Use the channel as a semaphore to limit the number of processes created for computing bid processes
+	// Most platforms put a limit on the number of processes a user can open. Even if the limit is high
+	// it isn't a good idea to open thuosands of processes.
```
boz, line 198: I like the defensive programming, but I just don't know if it's necessary at this point.

provider/cmd/run.go
```diff
@@ -79,15 +86,101 @@ func RunCmd() *cobra.Command {
...
 
+const BidPricingStrategyScale = "scale"
+const BidPricingStrategyRandomRange = "randomRange"
+const BidPricingStrategyShellScript = "shellScript"
+
+var AllowedBidPricingStrategies = [...]string{
+	BidPricingStrategyScale,
+	BidPricingStrategyRandomRange,
+	BidPricingStrategyShellScript,
+}
```
boz, line 126: is this used?

provider/cmd/run.go
```diff
@@ -79,15 +86,101 @@ func RunCmd() *cobra.Command {
...
+	if err := viper.BindPFlag(FlagBidPriceScriptProcessLimit, cmd.Flags().Lookup(FlagBidPriceScriptProcessLimit)); err != nil {
+		return nil
+	}
+
 	return cmd
 }
 
+const BidPricingStrategyScale = "scale"
+const BidPricingStrategyRandomRange = "randomRange"
+const BidPricingStrategyShellScript = "shellScript"
```
boz, line 120: nit: single const; it looks like these don't need to be exported.

boz (commented):
Really great work.  I left some comments inline.

provider/bidengine/pricing.go
hydrogen18: Doesn't the coin object use a big integer internally?

boz: ah, true.  too late for me.

provider/bidengine/pricing.go
```diff
@@ -1,15 +1,123 @@
...
+	}
+
+	cpuCost := sdk.NewCoin(denom, sdk.NewIntFromBigInt(cpuTotal))
+	memoryCost := sdk.NewCoin(denom, sdk.NewIntFromBigInt(memoryTotal))
+	storageCost := sdk.NewCoin(denom, sdk.NewIntFromBigInt(storageTotal))
+
+	// Check for less than or equal to zero
+	cost := cpuCost.Add(memoryCost).Add(storageCost)
+
+	if cost.Amount.Uint64() == 0 {
```
boz, line 106: Is there an `IsZero()` method?

provider/bidengine/pricing.go
hydrogen18: Is there any limit to the number of orders that a provider can receive in a given time period? Without this check it feels like an attack vector against providers.

provider/bidengine/order_test.go
hydrogen18: That's possible but what would the exported thing be named? We can't use `Reservation` since the interface is already called that. We'd have to use some name like `ReservationImpl` , `BasicReservation` or `ConcreteReservation`

provider/cmd/run.go
hydrogen18: I'm making a code change that will show this if a correct value isn't given. I'll change it to non-exported since it should never needed anywhere but here.

provider/bidengine/order.go
hydrogen18: To close the bid I'll need to send a `mtypes.MsgCloseBid` I am presuming?

provider/bidengine/pricing.go
boz: It's limited by number of orders that can make it into a block - I'd be surprised to see 500 in a block (~6 seconds).

I'm kinda neutral on it.  It feels like it should be more baked but it's okay to leave for now if you feel strongly about it.  It is a good thing to protect after all.

In a follow-up issue it'd be nice to

1. make it cancel-able (exits early if `ctx.Done()` is readable.
1. set a limit on execution time (there's a `cmd.RunContext()` I believe).

provider/bidengine/order_test.go
hydrogen18: I am using a mock for this now

provider/bidengine/pricing.go
hydrogen18: This is done now.

provider/bidengine/order.go
hydrogen18: I'm sending a `*mtypes.MsgCloseBid` now as well as the other changes you recommended.

provider/bidengine/pricing.go
```diff
@@ -1,15 +1,127 @@
...
 	"github.com/ovrclk/akash/validation"
 	dtypes "github.com/ovrclk/akash/x/deployment/types"
 )
 
-func calculatePrice(gspec *dtypes.GroupSpec) (sdk.Coin, error) {
+type BidPricingStrategy interface {
+	calculatePrice(gspec *dtypes.GroupSpec, lc lifecycle.Lifecycle) (sdk.Coin, error)
+}
+
+const denom = "uakt"
```
boz, line 24: do we have this defined somewhere else?

provider/bidengine/pricing.go
```diff
@@ -60,9 +172,175 @@ func calculatePriceRange(gspec *dtypes.GroupSpec) (sdk.Coin, sdk.Coin) {
...
+	path         string
+	processLimit chan int
+	runtimeLimit uint
+}
+
+var errPathEmpty = errors.New("script path cannot be the empty string")
+var errProcessLimitZero = errors.New("process limit must be greater than zero")
+var errProcessRuntimeLimitZero = errors.New("process runtime limit must be greater than zero")
+
+func MakeShellScriptPricing(path string, processLimit uint, runtimeLimit uint) (BidPricingStrategy, error) {
```
boz, line 196: just use `time.Duration` for `runtimeLimit` from the get-go.

provider/bidengine/pricing.go
```diff
@@ -60,9 +172,175 @@ func calculatePriceRange(gspec *dtypes.GroupSpec) (sdk.Coin, sdk.Coin) {
...
+	Memory  uint64 `json:"memory"`
+	CPU     uint64 `json:"cpu"`
+	Storage uint64 `json:"storage"`
+	Count   uint32 `json:"count"`
+}
+
+var errCalculatePriceCancelled = errors.New("calculate price cancelled")
+var errCalculatePriceTimeout = errors.New("calculate price timed out")
+
+func (ssp shellScriptPricing) calculatePrice(gspec *dtypes.GroupSpec, lc lifecycle.Lifecycle) (sdk.Coin, error) {
```
boz, line 233: this should take a `context.Context`.  Then you can layer on execution timeout with [`context.WithTimeout`](https://golang.org/pkg/context/#WithTimeout) and halt the program with [`CommandContext()`](https://golang.org/pkg/os/exec/#CommandContext).

provider/bidengine/pricing.go
```diff
@@ -60,9 +172,175 @@ func calculatePriceRange(gspec *dtypes.GroupSpec) (sdk.Coin, sdk.Coin) {
...
+	case <-time.After(time.Duration(ssp.runtimeLimit) * time.Millisecond):
+		timeout = true
+	case err := <-cmdDone:
+		if err != nil {
+			return sdk.Coin{}, err
+		}
+
+	}
+
+	if halt || timeout {
```
boz, line 295: use [`CommandContext()`](https://golang.org/pkg/os/exec/#CommandContext) so you don't have to halt it yourself.

boz (commented):
Looks great.  Some comments inline.

I still think that if `shouldBid()` is false that we should be doing `break loop` instead of `break`.

provider/bidengine/pricing.go
boz: you can use `time.ParseDuration` to parse the flag/env so that the input can be things like `"500ms"`.

provider/cmd/run.go
```diff
@@ -38,7 +39,14 @@ const (
...
 	// FlagGatewayListenAddress determines listening address for Manifests
-	FlagGatewayListenAddress = "gateway-listen-address"
+	FlagGatewayListenAddress       = "gateway-listen-address"
+	FlagBidPricingStrategy         = "bid-price-strategy"
+	FlagBidPriceCPUScale           = "bid-price-cpu-scale"
+	FlagBidPriceMemoryScale        = "bid-price-memory-scale"
+	FlagBidPriceStorageScale       = "bid-price-storage-scale"
+	FlagBidPriceScriptPath         = "bid-price-script-path"
+	FlagBidPriceScriptProcessLimit = "bid-price-script-process-limit"
+	FlagBidPriceScriptTimeout      = "bid-price-script-process-timeout"
```
boz, line 49: Can you add all of these to the kustomize [deployment environment](https://github.com/ovrclk/akash/blob/494572ac32a7a48d801a27397e6d2047d80419ec/_docs/kustomize/akash-provider/deployment.yaml#L212-L224) and [config](https://github.com/ovrclk/akash/blob/master/_docs/kustomize/akash-provider/kustomization.yaml#L50-L51)?

provider/cmd/run.go
```diff
@@ -79,15 +87,113 @@ func RunCmd() *cobra.Command {
...
 	return cmd
 }
 
+const (
+	bidPricingStrategyScale       = "scale"
+	bidPricingStrategyRandomRange = "randomRange"
+	bidPricingStrategyShellScript = "shellScript"
+)
+
+var allowedBidPricingStrategies = [...]string{
```
Ropes, line 129: Making them types eg: `type bidPricingStrategy string`  then declare them as `var scale bidPricingStrategy = "scale"`; it would make the comparison between types, and not raw string data. 

Then it becomes trivial to use a type switch for selection, and easy default clause for the errNoSuchBidPricing...

Unless there's some issue with the viper flags that this pattern escapes with :grimacing:

Ropes (commented):
Looks great! 

Left some comments for clarification.

provider/bidengine/order_test.go
```diff
@@ -0,0 +1,434 @@
...
+	default:
+	}
+	// Should never have broadcast
+	require.Nil(t, broadcast)
+}
+
+type testBidPricingStrategy int64
+
+func (tbps testBidPricingStrategy) calculatePrice(gspec *dtypes.GroupSpec, _ lifecycle.Lifecycle) (sdk.Coin, error) {
+	return sdk.NewInt64Coin("akt", int64(tbps)), nil
```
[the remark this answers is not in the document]
hydrogen18: I think I am just testing that the result get passed back here, when the bid is published as a transaction

provider/cmd/run.go
hydrogen18: Are you suggesting I'd cast from a string into `bidPricingStrategy` ?

provider/bidengine/order.go
```diff
@@ -154,40 +156,47 @@ loop:
...
-			reservation = result.Value().(cluster.Reservation)
+			pricech = runner.Do(func() runner.Result {
+				// Calculate price & bid
+				return runner.NewResult(o.pricingStrategy.calculatePrice(&group.GroupSpec, o.lc))
 
-			price, err := calculatePrice(&group.GroupSpec)
-			if err != nil {
-				o.log.Error("error calculating price", "err", err)
+			})
+		case result := <-pricech:
```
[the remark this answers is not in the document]
hydrogen18: I'm not sure I understand what you're describing here.

provider/bidengine/pricing.go
hydrogen18: If I understand what you're describing, I'd have to create a context in `provider/bidengine/order.go` and then pass it into this function.  Then when the `provider/bidengine/order.go` hits the case of `<-o.lc.ShutdownRequest()`  I'd have cancel the context _if it exists at that time_. 

I'm not sure how I understand this is simpler.

provider/bidengine/pricing.go
hydrogen18: What is a cosmos params file?

provider/bidengine/order_test.go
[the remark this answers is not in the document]
Ropes: Ah, I'd still say lets keep it simple and avoid using `akt` since we don't really use that denom internally.

`testutil.CoinDenom` keeps us consistent for tests too!

provider/bidengine/pricing.go
hydrogen18: > Looks great. Some comments inline.
> 
> I still think that if `shouldBid()` is false that we should be doing `break loop` instead of `break`.

If we do `break loop` when `shouldBid()` is false then that means no reservation is made and the goroutine associated with the `order` object terminates immediately after the first query comes back.

This is straightforward to implement. But what I don't understand is what the purpose of launching the goroutine in the first place? It's going to terminate basically immediately, the only side effect being the query that loads the group.

provider/cmd/run.go
Ropes: That's what I was effectively saying, but realized since we're just representing a state, we don't really need the string aspect to describe if we use a stringification. For example
```
//go:generate stringer -linecomment -output=autogen_stringer.go -type=bidPricingStrategy

type bidPricingStrategy uint8

const (
	scaling bidPricingStrategy = iota + 1
	randomRange
	shellsScript
)
```

Then by running `go generate ...` we'll be able an access a string of these types with the generated code:
```go
func _() {
	// An "invalid array index" compiler error signifies that the constant values have changed.
	// Re-run the stringer command to generate them again.
	var x [1]struct{}
	_ = x[scaling-1]
	_ = x[randomRange-2]
	_ = x[shellsScript-3]
}

const _bidPricingStrategy_name = "scalingrandomRangeshellsScript"

var _bidPricingStrategy_index = [...]uint8{0, 7, 18, 30}

func (i bidPricingStrategy) String() string {
	i -= 1
	if i >= bidPricingStrategy(len(_bidPricingStrategy_index)-1) {
		return "bidPricingStrategy(" + strconv.FormatInt(int64(i+1), 10) + ")"
	}
	return _bidPricingStrategy_name[_bidPricingStrategy_index[i]:_bidPricingStrategy_index[i+1]]
}
```

It very nice!  Then all the comparisons are between static types, no string comparison necessary, but we get efficient String() formatting for debugging! Not necessary to change, but we've used it in the past.

provider/bidengine/pricing.go
boz: Just make a context at the beginning of run and always cancel it before returning from run. 

Context is the common lifetime object to pass around.  Using it plus runcontext makes it so that you don't need to do all this grouting handling in here.

It's also just not a good idea to leak the lifecycle object as you've done.  There is a reader interface you can use but please just pass around context instead.

provider/bidengine/pricing.go
boz: It handles an order.  Goroutines are very cheap.

_docs/kustomize/akash-provider/deployment.yaml
```diff
@@ -223,6 +223,54 @@ spec:
                   key: mem-price-max
                   optional: true
```
boz, line 224: Is the old stuff still relevant?  `mem-price-min`, etc...

provider/bidengine/order.go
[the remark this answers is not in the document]
boz: It should ensure that it cleans up anything running for pricing before exiting `run()`.  To do this you set `pricech` to `nil` after you've read from it and later wait for it to be readable before returning if it's not nil.  I linked to the pattern for the other channels.

provider/bidengine/order.go
```diff
@@ -29,52 +31,55 @@ type order struct {
...
+		svc.drainch <- order
 	}()
 
 	return order, nil
 }
 
 func (o *order) run() {
 	defer o.lc.ShutdownCompleted()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
boz, line 75: you'll want to `cancel()` right after leaving the main event loop to start draining everything, not here.

provider/cmd/run.go
boz: When this becomes more involved we can do a more typical registry pattern.  for now I think it's just fine.

provider/bidengine/order.go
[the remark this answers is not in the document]
hydrogen18: Oh ok, it seems like we could add a utility function that does this with a list of channels.

provider/bidengine/pricing.go
hydrogen18: This change has been made

provider/bidengine/pricing.go
hydrogen18: This change has been made, the `order` has its own context and each pricing operation that calls a script makes a child context with a deadline to handle timeouts.

_docs/kustomize/akash-provider/deployment.yaml
hydrogen18: The original pricing algorithm is still present in the code. I'm assuming its used in some of the e2e tests.

provider/bidengine/order.go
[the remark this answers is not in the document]
boz: agreed, especially because we mostly use `chan runner.Result`.  maybe add a utility in the `runner` package soon?

boz (approved):
:rocket:

boz:
@troian please take a look when you can.

PR #883, opened 2020-10-22
switch to akash fork of cosmos SDK fixes #838

### LINKED ISSUE #838: Fix configuration inconsistency in app &  Kubernetes depoyment in '_run/single'

hydrogen18:


boz (approved):
Can you check to see if your upstream PR has been merged first?

hydrogen18:
It has been merged but is not in a tag

Issue #778, opened 2020-08-05
provider sdl validation


boz:
Validate SDL upon retrieval.  Some validations:

1. if [`DEPLOYMENT_INGRESS_STATIC_HOSTS`](https://github.com/ovrclk/akash/blob/faf26a8704ecf39730597254b9633d3da6542ca7/provider/cluster/kube/settings.go#L18-L20) is `false` then services [must have a host](https://github.com/ovrclk/akash/blob/faf26a8704ecf39730597254b9633d3da6542ca7/manifest/types.go#L67).
1. Don't allow clobbering of existing ingress hosts
1. Add a configurable set of disallowed user-supplied hosts.

boz:
> 1. If a bid is cancelled due to hostname collision, stop bidding on orders from that addr in the network for some time period. Effective, but kind of dumb IMHO.

Yes we could do that or ignore the deployment and/or group.

> 2. Include a 32 bit hash of each hostname in the deployment (on the blockchain) and check that against the existing hosts before bidding. This increases the amount of data on the blockchain and is not perfect, for example hash collisions could cause a false positive
> 
I like this idea.  I think we should consider it, but let's look at the various name service protocols in the cosmos space that we might be able to leverage first.

===

I think that in the case of a hostname conflict, we should just reject the manifest - return an error during send-manifest - and leave it up to the user to update the deployment or close the lease.

We definitely need the other two things too tho:

* static/configurable hostname blacklist
* validation

hydrogen18:
When you say " various name service protocols in the cosmos space" what are you referring to?

hydrogen18:
Next step on this is to check in the handler for send-manifest if there is a hostname conflict and reject the manifest there with appropriate error code

PR #894, opened 2020-11-05
feat(provider): Add support for exposing services on TCP & UDP fixes #867

### LINKED ISSUE #867: Endpoint Resource
# Overview

Deployments that are open globally are assumed to be virtual-hosted behind a single IP, along with other tenants.  This form of multi-tenancy works okay for some classes of web applications but it falls short of supporting things like:

- UDP protocols
- Arbitrary TCP protocols
- High-performance networking

In lieu of implementing non-shared IP addresses, we can allow UDP, TCP protocols to be accessed through random ports on the provider incoming endpoints.

The user's intent to use a random port can be declared as part of an "Endpoint" resource.  The platform can then expose the user's workloads over a random port, and advertise the IP,port in the lease status API.

# Components

- [ ] SDL - declare endpoint resources
- [ ] Blockchain: affects matching, etc...
- [ ] Platform: implement `NodePort` ingresses when applicable
- [ ] Platform: expose ports via the lease status gateway API.

hydrogen18:
So far I've done the following on this 

1. Added an endpoint resource in the order, which the presence of indicates the desire for an external port. The actual
message doesn't describe anything because it is getting delegated to NodePort which is random.
2. Updated the manifest to have Global indicate a port should be exposed
3. Updated the provisioning process to provision Port 80/TCP as kubernetes ingress. Everything else is done via NodePort
3a. When Kubernetes service objects are created they get suffixed with "-global" if they are getting declared as NodePort
4. Updates to the newer version of the k8s go client
5. Moved from beta Kubernetes Ingress to V1 Ingress
6. Updated the inventory service to track the number of external ports in use
7. Added a command line switch to "provider run" to indicate the number of external ports
8. Added a command line switch to "provider run" to indicate the extrenal hostname of the cluster
9. Added to the lease status the port information
10. Updated the deployment monitor to take into account that not all deployments on Akash are going to have an Ingress now

In the process of adding tests I generated mocks for the Kubernetes client, so this PR is quite large. Most of it just 
those generated mocks however.

I am still working on making the examples work, I will need to add the flags to the kustomize service definition as
environmental variables.

Additional validation of the manifest is going to happen under #778.One item we need to make sure of is that the number of "global" ports declared in the manifest matches the number specified in the order.
However I am not sure where to add that at this time.


hydrogen18:
This is for issue #867

provider/cluster/kube/client.go
```diff
@@ -443,7 +516,8 @@ func (c *client) deploymentsForLease(ctx context.Context, lid mtypes.LeaseID) ([
 		c.log.Error(err.Error())
 		return nil, errors.Wrap(err, ErrInternalError.Error())
 	}
-	if deployments == nil {
+
+	if deployments == nil || 0 == len(deployments.Items) {
```
boz, line 520: :nail_care: `len(nil) == 0`; you don't need both & it's better to always use the one you added.

I think that the differentiation was intentional - zero deployments is the empty set, not an error; but I don't remember why it was that way.

provider/cluster/monitor.go
```diff
@@ -148,26 +148,45 @@ func (m *deploymentMonitor) doCheck() (bool, error) {
...
+		for _, exposed := range spec.Expose {
+			if exposed.Global {
+				expectedPortCount++
+				// Get the list of all forwarded ports for this deployment
+				forwardedPorts := status.ForwardedPorts[spec.Name]
+				for _, forwarded := range forwardedPorts {
+					// Search for a match on port & protocol
+					if exposed.Port == forwarded.Port && exposed.Proto == forwarded.Proto {
+						foundPortCount++
+						if uint32(forwarded.Available) < spec.Count {
```
boz, line 174: Again I'm not sure that `spec.Count` makes sense here?  Maybe I'm misunderstanding how it's accounted for in k8s.

provider/cluster/monitor.go
```diff
@@ -148,26 +148,45 @@ func (m *deploymentMonitor) doCheck() (bool, error) {
...
+							badsvc++
```
boz, line 175: will this cause a double `badsvc++` below?

provider/cluster/monitor.go
```diff
@@ -148,26 +148,45 @@ func (m *deploymentMonitor) doCheck() (bool, error) {
...
 					"target", spec.Count,
 				)
 			}
 		}
 
-		if !found {
+		expectedPortCount := 0
+		foundPortCount := 0
+		for _, exposed := range spec.Expose {
+			if exposed.Global {
```
boz, line 166: maybe put all of this in a function

provider/cluster/service.go
```diff
@@ -40,7 +40,7 @@ type Service interface {
 }
 
 // NewService returns new Service instance
-func NewService(ctx context.Context, session session.Session, bus pubsub.Bus, client Client) (Service, error) {
+func NewService(ctx context.Context, session session.Session, bus pubsub.Bus, client Client, clusterExternalPortQuantity uint) (Service, error) {
 	log := session.Log().With("module", "provider-cluster", "cmp", "service")
 
 	config := config{}
```
boz, line 46: Are the CLI flags not affecting this config object?  I think we need to get rid of the `env.Parse` here and just pass this object into `NewService`

provider/service.go
```diff
@@ -41,7 +41,7 @@ type Service interface {
 
 // NewService creates and returns new Service instance
 // Simple wrapper around various services needed for running a provider.
-func NewService(ctx context.Context, session session.Session, bus pubsub.Bus, cclient cluster.Client) (Service, error) {
+func NewService(ctx context.Context, session session.Session, bus pubsub.Bus, cclient cluster.Client, clusterPublicHostname string, clusterExternalPortQuantity uint) (Service, error) {
```
boz, line 44: yeah let's populate a config object to pass in here instead of adding more and more params.

provider/service.go
```diff
@@ -86,15 +86,16 @@ func NewService(ctx context.Context, session session.Session, bus pubsub.Bus, cc
...
+		session:               session,
+		bus:                   bus,
+		cluster:               cluster,
+		cclient:               cclient,
+		bidengine:             bidengine,
+		manifest:              manifest,
+		ctx:                   ctx,
+		cancel:                cancel,
+		lc:                    lifecycle.New(),
+		clusterPublicHostname: clusterPublicHostname,
```
boz, line 98: this should be in `config`.

provider/testutil/provider.go
```diff
@@ -37,6 +37,9 @@ func TestSendManifest(clientCtx client.Context, id mtypes.BidID, sdlPath string,
 	return testutilcli.ExecTestCLICmd(clientCtx, pcmd.SendManifestCmd(), args...)
 }
 
+const TestClusterPublicHostname = "e2e.test"
+const TestClusterNodePortQuantity = 100
```
boz, line 41: :nail_care: 
```go
const (
  ...
)
```

sdl/v2.go
```diff
@@ -178,6 +179,11 @@ func (sdl *v2) Manifest() (manifest.Manifest, error) {
 
 			for _, expose := range svc.Expose {
 				for _, to := range expose.To {
+					if expose.Proto == "" {
+						expose.Proto = manifest.TCP
+					} else {
+						expose.Proto = strings.ToUpper(expose.Proto)
```
boz, line 185: I think we should have a `manifest.ParseProto()` that returns an enum or type derived from `string`.

boz (commented):
Looks _amazing_.  I took a first pass at it, will do another shortly.

provider/cluster/inventory.go
```diff
@@ -343,40 +363,57 @@ func (is *inventoryService) getStatus(inventory []Node, reservations []*reservat
...
 	return ok
 }
 
-func reservationAdjustInventory(prevInventory []Node, reservation *reservation) ([]Node, bool) {
+func reservationCountEndpoints(reservation *reservation) uint {
+	var externalPortCount uint
+
+	resources := reservation.resources.GetResources()
+	for _, resource := range resources {
+		externalPortCount += uint(len(resource.Resources.Endpoints) * int(resource.Count))
```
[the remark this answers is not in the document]
hydrogen18: Yeah thinking about it, it is just going to round robin amongst them. So it should only be a single port.

provider/cluster/service.go
hydrogen18: What object are you referring to here?

boz: `config{}`.

we were using module-specific configuration via environment variables before instead of stringing the settings through from the command line.  it looks like things have changed and many of the settings are being set from the command line now, which is fine, but I think that some settings have not been migrated over.

* [`provider/config.go`](https://github.com/ovrclk/akash/blob/c0becaf676e20fd807a125776b55c3c90968c19a/provider/config.go#L5-L7)
* [`provider/cluster/config.go`](https://github.com/ovrclk/akash/blob/c0becaf676e20fd807a125776b55c3c90968c19a/provider/cluster/config.go#L6-L7)
* [`provider/cluster/kube/settings.go`](https://github.com/ovrclk/akash/blob/c0becaf676e20fd807a125776b55c3c90968c19a/provider/cluster/kube/settings.go#L14-L25)

provider/cluster/inventory_test.go
```diff
@@ -27,6 +29,10 @@ func TestInventory_reservationAllocateable(t *testing.T) {
 				Memory: &types.Memory{
 					Quantity: types.NewResourceValue(memory),
 				},
+				Storage: &types.Storage{
+					Quantity: types.NewResourceValue(storage),
+				},
+				Endpoints: endpoints,
```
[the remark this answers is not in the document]
hydrogen18: That was my original idea but it isn't surface to the inventory management. I could do that but it'd be a larger refactor.

provider/cluster/service.go
hydrogen18: Alright, I'll have to look some more at this in order to see what is going on.

hydrogen18:
@boz  Thanks for the feedback, I think I've covered everything you brought up.

Tests are passing for me locally, but not in CI because it's trying to find some Kubernetes config (which I have locally on my machine). I'm figuring out how to skip / bypass that config loading in test. In all the tests I am using a stubbed Kubernetes client so no actual Kubernetes cluster is needed.

provider/cluster/inventory.go
```diff
@@ -201,7 +208,17 @@ loop:
 						continue
 					}
 
+					allocatedPrev := res.allocated
 					res.allocated = ev.Status == event.ClusterDeploymentDeployed
+					allocatedChanged := res.allocated == allocatedPrev
```
boz, line 213: `!=`?

provider/cluster/inventory.go
```diff
@@ -343,40 +363,57 @@ func (is *inventoryService) getStatus(inventory []Node, reservations []*reservat
 	return status
 }
 
-func reservationAllocateable(inventory []Node, reservations []*reservation, newReservation *reservation) bool {
+func reservationAllocateable(inventory []Node, externalPortsAvailable uint, reservations []*reservation, newReservation *reservation) bool {
 	// 1. for each unallocated reservation, subtract its resources
 	//    from inventory.
 	// 2. subtract resources for new reservation from inventory.
-	// 3. return true iff 1 and 2 succeed.
+	// 3. return true if 1 and 2 succeed.
```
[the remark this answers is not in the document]
boz: :point_up:

provider/cluster/kube/builder.go
```diff
@@ -287,9 +287,10 @@ func (b *deploymentBuilder) container() corev1.Container {
 // service
 type serviceBuilder struct {
 	deploymentBuilder
+	isGlobal bool
```
boz, line 290: I think that this is confusing because "global" doesn't necessarily mean that it should be a `NodePort` service.

provider/cluster/kube/builder.go
```diff
@@ -300,42 +301,95 @@ func newServiceBuilder(log log.Logger, settings settings, lid mtypes.LeaseID, gr
...
+
+func (b *serviceBuilder) ports() ([]corev1.ServicePort, error) {
 	ports := make([]corev1.ServicePort, 0, len(b.service.Expose))
 	for i, expose := range b.service.Expose {
-		ports = append(ports, corev1.ServicePort{
-			Name:       strconv.Itoa(int(expose.Port)),
-			Port:       exposeExternalPort(&b.service.Expose[i]),
-			TargetPort: intstr.FromInt(int(expose.Port)),
-		})
+		if expose.Global == b.isGlobal {
```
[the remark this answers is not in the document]
boz: :point_up:

boz (commented):
Nice changes.  There's a few remaining comments.

provider/cluster/kube/client.go
```diff
@@ -51,10 +50,8 @@ type client struct {
 
 // NewClient returns new Kubernetes Client instance with provided logger, host and ns. Returns error incase of failure
 func NewClient(log log.Logger, host, ns string) (Client, error) {
-	var settings settings
-	if err := env.Parse(&settings); err != nil {
-		return nil, err
-	}
+	settings := newDefaultSettings()
```
boz, line 53: These need to be able to be configured.  It used to happen above in `env.Parse`, it's not anymore.

provider/service.go
```diff
@@ -112,9 +111,10 @@ type service struct {
 	bidengine bidengine.Service
 	manifest  manifest.Service
 
-	ctx    context.Context
-	cancel context.CancelFunc
-	lc     lifecycle.Lifecycle
+	ctx                   context.Context
+	cancel                context.CancelFunc
+	lc                    lifecycle.Lifecycle
+	clusterPublicHostname string
```
boz, line 117: isn't this in the `config`?

provider/cluster/kube/builder.go
hydrogen18: What is the intent of the "Global" flag then?

hydrogen18: Will refactor struct member name to `requiresNodePort`?

PR #899, opened 2020-11-09
feat(goreleaser): edgenet homebrew tap

### LINKED ISSUE #888: split homebrew release in to multiple taps
Homebrew does not support multiple versions of single formulae thus it installs whatever is in recipe (though it still can install recipe and HEAD version from git).
For example when mainnet released with version `v0.2.0` followed by release on master `v0.3.0` actual formulae will contain release `v0.3.0` which might be incompatible (i.e. network reset etc) with mainnet.

One of options to avoid this is to not allow edge releases to deploy to mainnet homebrew formulae.
Edge releases though still may need to be accessed from homebrew for testing purposes. That can be achieved with creating keg-only formulae named something like `akash-edge`.
This approach imposes maintaining of two separate goreleaser configs.

boz:
fixes #888

### DIFF
--- a/.goreleaser.edge.yaml
+++ b/.goreleaser.edge.yaml
@@ -1,0 +1,117 @@
+env:
+  - GO111MODULE=on
+  - CGO_ENABLED=1
+builds:
+  - id: akash-darwin-amd64
+    binary: akash
+    main: ./cmd/akash
+    goarch:
+      - amd64
+    goos:
+      - darwin
+    env:
+      - CC=o64-clang
+      - CXX=o64-clang++
+    flags:
+      - -trimpath
+      - -mod=readonly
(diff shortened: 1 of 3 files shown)


troian (approved):

troian:
just came with idea we might be able to leave with one `.goreleaser.yaml`

troian:
@boz https://github.com/ovrclk/akash/commit/b05ce9edba9a852295f11cc63a09b0f950f33fa9

troian:
@boz force-pushed change of edge tap name to `akash-edge`
https://github.com/ovrclk/akash/pull/900/commits/a5b2b498f2bbb4166b5b590c074ff4bacf7f14ca

boz:
superseded by #900

PR #904, opened 2020-11-11
Bid pricing configuration & endpoint resource Fixes #867 #868

### LINKED ISSUE #867: Endpoint Resource

hydrogen18:
This combines both of my branches into a single one.


proto/akash/deployment/v1beta1/group.proto
```diff
@@ -95,4 +95,5 @@ message Resource {
   uint32                   count = 2 [(gogoproto.jsontag) = "count", (gogoproto.moretags) = "yaml:\"count\""];
   cosmos.base.v1beta1.Coin price = 3
       [(gogoproto.nullable) = false, (gogoproto.jsontag) = "price", (gogoproto.moretags) = "yaml:\"price\""];
+  ;
```
boz, line 98: :scissors:

boz (approved):

manifest/parse.go
```diff
@@ -0,0 +1,27 @@
...
+	"strings"
+)
+
+var ErrUnsupportedServiceProtocol = errors.New("Unsuported service protocol")
+
+func ParseServiceProtocol(input string) (ServiceProtocol, error) {
+	var result ServiceProtocol
+
+	// This is not a case sensitive parse, so make all input
+	// upppcase
```
troian, line 14: uppercase ?

provider/cluster/inventory.go
```diff
@@ -201,7 +209,17 @@ loop:
 						continue
 					}
 
+					allocatedPrev := res.allocated
 					res.allocated = ev.Status == event.ClusterDeploymentDeployed
+					allocatedChanged := res.allocated != allocatedPrev
+					if allocatedChanged {
```
troian, line 215: `if res.allocated != allocatedPrev {`

provider/cluster/inventory.go
```diff
@@ -355,27 +376,46 @@ func reservationAllocateable(inventory []Node, reservations []*reservation, newR
...
+
+	return externalPortCount
+}
+
+func reservationAdjustInventory(prevInventory []ctypes.Node, externalPortsAvailable uint, reservation *reservation) ([]ctypes.Node, uint, bool) {
 	// for each node in the inventory
 	//   subtract resource capacity from node capacity if the former will fit in the latter
 	//   remove resource capacity that fit in node capacity from requested resource capacity
-	// return remaining inventory, true iff all resources were able to fit
+	// return remaining inventory, true if all resources are able to fit
```
troian, line 407: `iff` if an only if

provider/cluster/kube/client.go
```diff
@@ -115,19 +112,17 @@ func openKubeConfig(log log.Logger) (*rest.Config, error) {
 	return rest.InClusterConfig()
 }
 
-func (c *client) shouldExpose(expose *manifest.ServiceExpose) bool {
-	return expose.Global &&
-		(expose.ExternalPort == 80 ||
-			(expose.ExternalPort == 0 && expose.Port == 80))
+func shouldExpose(expose *manifest.ServiceExpose) bool {
+	return expose.Proto == manifest.TCP && expose.Global && 80 == exposeExternalPort(expose)
```
troian, line 116: make an issue or some note that only `TCP` ports can be exposed, some protocols like `QUIC` are based on `UDP` and won't work

provider/cluster/monitor.go
```diff
@@ -148,26 +148,45 @@ func (m *deploymentMonitor) doCheck() (bool, error) {
...
+					"available", service.Available,
 					"target", spec.Count,
 				)
 			}
 		}
 
-		if !found {
+		expectedPortCount := 0
+		foundPortCount := 0
+		for _, exposed := range spec.Expose {
```
troian, line 165: can we somehow refactor block down below? to many nested levels.

provider/cluster/service.go
```diff
@@ -95,7 +89,7 @@ type service struct {
 
 	inventory *inventoryService
 
-	statusch  chan chan<- *Status
+	statusch  chan chan<- *ctypes.Status
```
troian, line 92: maybe for later. change to value, not pointer. pointer creates some pressure on GC

troian (changes requested):
awesome work, just couple minor comments

provider/cluster/inventory.go
hydrogen18: I agree we could eliminate the local variable here. I usually like to do this however because the comparison `!=` is just a comparison. But by assigning it to a local variable with a name, you can document that meaning of the comparison.

provider/cluster/kube/client.go
hydrogen18: I don't know that kubernetes ingress supports QUIC at this time? HTTP 2.0 should work in theory.

Maybe QUIC support should be an issue down the road? I don't think it blocks any of the present use cases.

provider/cluster/monitor.go
hydrogen18: I agree it is pretty deep, but I think we will refactor this whenever we add support for more endpoint features like individual public IP, etc.

provider/cluster/service.go
hydrogen18: I will add an issue to address this in the future.

provider/cluster/inventory.go
troian: I think pattern `curr != prev` is self descriptive though

troian (approved):

pkg/apis/akash.network/v1/types.go
```diff
@@ -255,22 +260,27 @@ type ManifestServiceExpose struct {
 	Hosts []string `json:"hosts,omitempty"`
 }
 
-func (mse ManifestServiceExpose) toAkash() manifest.ServiceExpose {
+func (mse ManifestServiceExpose) toAkash() (manifest.ServiceExpose, error) {
+	proto, err := manifest.ParseServiceProtocol(mse.Proto)
+	if err != nil {
+		fmt.Printf("foobar: %q\n", mse.Proto)
```
boz, line 266: :scissors:

_run/lite/Makefile
```diff
@@ -6,6 +6,8 @@ PROVIDER_HOSTNAME ?= http://localhost:8080
 provider-run:
 	$(AKASHCTL) $(KEY_OPTS) provider run $(CHAIN_OPTS) \
 		--from "$(PROVIDER_KEY_NAME)"
```
boz, line 8: :bug:

integration/test_helpers.go
```diff
@@ -71,11 +72,13 @@ func queryApp(t *testing.T, appURL string, limit int) {
 	var resp *http.Response
 	for i := 0; i < limit; i++ {
 		time.Sleep(1 * time.Second) // reduce absurdly long wait period
+		fmt.Fprintf(os.Stdout, "queryApp: polling app at %q attempt %d\n", req.URL.String(), i)
```
boz, line 75: :scissors:

integration/test_helpers.go
```diff
@@ -71,11 +72,13 @@ func queryApp(t *testing.T, appURL string, limit int) {
...
 		resp, err = client.Do(req)
 		if err != nil {
 			t.Log(err)
 			continue
 		}
+		fmt.Fprintf(os.Stdout, "queryApp: attempt %d response is %d\n", i, resp.StatusCode)
```
boz, line 81: :scissors:

Ropes (approved):

PR #905, opened 2020-11-11
feat: implement audited attributes

### LINKED ISSUE #870: Audited Attributes
# Overview

A provider's attributes are currently not audited in any way at all - a provider can populate their attributes with whatever they want, whether they accurately represent the provider or not.

Akash needs to facilitate the formation of *auditors* who attest to the validity of provider attributes.

In particular, at initial launch, we need to act as an auditor that blesses our datacenter providers.  Without this it will be difficult to have any kind of quality control for the network.

# Requirements

- Accounts can sign some set of attributes for any provider.
- Signed attributes can be revoked as needed
- Deployment Groups can declare attributes that must be signed, and by who

# Components

- blockchain
    - sign, revoke attributes
    - affects order matching
- SDL
    - declare desired signed attributes
- provider: affects applicable orders

troian:
fixes #870

- test environment is based on _run/lite
- some deployment tests are busted due to change of type for Group.Requirements
- audit/handler tests TBD
- add attributes tests

working
- sign/update/delete attributes
- create deployment with attribute signature as requirement. market checks if provider can bid

limitations
- audit key built as 0x01/validator/owner address address. this causes limitation in querying list of all attributes signed by specific validator 

notes:
- proto/breakage test is expected to fail due to change of Group type

### how to test:
all examples below are tested with `_run/lite`
key `validator` is used as signer in all examples below

**prerequisites**
dump address of validator key and add it to `profiles.placement.westcoast.signedBy.anyOf[0]`
<details>
  <summary>example</summary>

  ```yaml
  ---
version: "2.0"

services:
  web:
    image: nginx
    expose:
      - port: 80
        to:
          - global: true

profiles:

  compute:
    web:
      resources:
        cpu:
          units: "100m"
        memory:
          size: "128Mi"
        storage:
          size: "256Mi"

  placement:
    westcoast:
      attributes:
        region: us-west
      signedBy:
        anyOf:
          - akash1v6aycfg99a9x4j8sltgrspmthpfzppjrn65qe9
      pricing:
        web:
          denom: uakt
          amount: 1000

deployment:
  web:
    westcoast:
      profile: web
      count: 2
  ```
  
</details>

##### values from args
```sh
# there can be as many as needed key/val pairs
../../akash --chain-id=local --keyring-backend=test --home ./cache tx audit attr create <provider_owner> key1 val1 key2 val2 --from=validator
```

##### delete all
```sh
../../akash --chain-id=local --keyring-backend=test --home ./cache tx audit delete <provider_owner> --from= validator
```

##### delete only key1
```sh
../../akash --chain-id=local --keyring-backend=test --home ./cache tx audit delete <provider_owner> key1 --from=validator
```

##### all
```
../../akash --chain-id=local --home ./cache query audit list
```


boz (commented):
Looks great!  Some thoughts:

We don't need to worry about expiration right now.  Also we can revoke via deletion I think?

The interface can be simple - to add some audited attributes:

```sh
# single
akash tx attr create $provider region us-west --from $keyname
# multiple
akash tx attr create $provider region us-west host foo.com --from $keyname
# delete
akash tx attr delete $provider region --from $keyname
# etc...
```

The `Create` message can be something like

```
{
  Owner      string
  Validator  string
  Attributes []Attribute
}
```

The table key can just be `(owner-id,provider-id,attribute-name)`.

I get that having an "Auditor" object is nice, but I'm not sure it's needed right now.  If/when it is it will be easy to add if it continues to be indexed by the owner id.

x/audit/types/types.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+// Providers is the collection of Provider
+type Providers []Provider
+
+// String implements the Stringer interface for a Providers object.
+func (obj Providers) String() string {
+	var buf bytes.Buffer
+
+	const sep = "\n\n"
```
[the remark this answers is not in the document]
troian: all other `x/types.go` have it this way. but good point to check why

sdl/_testdata/simple.yaml
```diff
@@ -25,6 +25,13 @@ profiles:
     westcoast:
       attributes:
         region: us-west
+      signedBy:
+        anyOf:
+          - 1
+          - 2
```
boz, line 31: super cool :1st_place_medal:

types/attribute.go
```diff
@@ -40,6 +53,20 @@ func (m Attribute) SubsetOf(rhs Attribute) bool {
 	return false
 }
 
+func (attr Attributes) Validate() error {
```
boz, line 56: Is it possible to have an empty name here?

I think we should also restrict the names to be `[a-zA-Z][a-zA-Z_-]{0,32}` or something, too.

x/audit/client/cli/tx.go
```diff
@@ -0,0 +1,220 @@
...
+	flags.AddTxFlagsToCmd(cmd)
+
+	if err := cmd.MarkFlagRequired(flags.FlagFrom); err != nil {
+		panic(err.Error())
+	}
+}
+
+// readAttributes try read attributes from both cobra arguments and stdin
+// len of the args must be even
+// read from stdin uses trick to check if it's file descriptor is a pipe
```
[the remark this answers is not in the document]
boz: Yeah, this is unnecessary right now I think.  People can use `xargs` to achieve reading from a file.

I think for this first pass let's keep it simple - only use the `argv` pairs.

boz (commented):
awesome!  some comments inline.

types/attribute.go
troian: attributes in sdl is a map so it will fail if key is empty, but checking name is a good point

sdl/v2.go
```diff
@@ -98,13 +100,16 @@ func (sdl *v2) DeploymentGroups() ([]*dtypes.GroupSpec, error) {
 					Name: placementName,
 				}
 
-				for _, v := range infra.Attributes {
-					group.Requirements = append(group.Requirements, v)
-				}
+				group.Requirements.Attributes = infra.Attributes
```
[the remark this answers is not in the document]
troian: i think it's fine, attributes came from sdl, and not used anywhere else

types/attribute.go
troian: `scheiße` 32 is bit too short, `testutil` generates long keys 🤣

Issue #909, opened 2020-11-12
Akash kustomize Provider upgrade


Ropes:
In preparation for final testnet, the kustomize scripts that install Provider and also Node need to be upgraded.

### Why
* CosmosSDK 0.40 upgrade to config flags has broken Provider configuration. Related: https://github.com/ovrclk/akash/issues/908
* Upgrade testing environment to Kubernetes 1.19 to access features and APIs we're standardizing on.
* Improve usability of kustomize directories and scripts.

Issue #908, opened 2020-11-12
Come up with a way for us to call Cosmos AddTxFlagsToCmd without mandatory CLI flags


hydrogen18:
Cosmos exposes a function `AddTxFlagsToCmd` to add some flags to a cobra command object 

https://github.com/cosmos/cosmos-sdk/blob/v0.40.0-rc3/client/flags/flags.go#L87

The problem is they are doing  `cmd.MarkFlagRequired(FlagChainID) `. This means that a user absolutely must provide `--chain-id` on the command line when it runs. There is no way around this that I could find. In Akash we want users to be able to specify all configuration via whatever means they want (configuration file, env. var, etc.). Also most end users aren't going to be changing chain ID very often so requiring this is not a very good idea.

We need to patch Cosmos SDK to allow us to get the same flags added to the command but without the mandatory requirement. Once we have gotten that patch into Cosmos SDK we need to update to a version that contains the patch.

boz:
Maybe we can set a default value on that flag?

boz:
Maybe we can overwrite what `MarkFlagRequired()` does?

```go
// MarkFlagRequired instructs the various shell completion implementations to
// prioritize the named flag when performing completion,
// and causes your command to report an error if invoked without the flag.
func MarkFlagRequired(flags *pflag.FlagSet, name string) error {
	return flags.SetAnnotation(name, BashCompOneRequiredFlag, []string{"true"})
}
```
Or even just unroll the entire `AddTxFlagsToCmd` and skip the `MarkFlagRequired`.

hydrogen18:
Yeah, I think either one of those would work. If this isn't high priority I'd prefer to just propose a patch to cosmos SDK first that allows us to opt out of this.

PR #910, opened 2020-11-12
Kustomize provisioning Provider updates

Ropes:
~**Hold off on reviewing until after CI runs**~

Documentation updates for running a provider with `kustomize` configuration, running testing through the `_run/single` environment. Main problem was debugging our CLI situation which @hydrogen18 discovered is probably related to the methods used to access command flags.

* `akashctl` -> `akash` binary(cleanup of old references)
* `akt` -> `uakt`
* Kubernetes upgraded to 1.19+ (_run/single)
* Corrected `akash` params for configuration commands
* Utilize local directories for kustomization files
* Major change uses the cli flags to `akash provider` to pass
  configuration.
* Running akash node name: `akashd` -> `akash-node`
* Ingress accessible via: akash.localhost

fixes: https://github.com/ovrclk/akash/issues/909


_docs/examples/provider/README.md
```diff
@@ -25,33 +32,45 @@ mkdir akash-demo && cd akash-demo
 
 ### Kubernetes Cluster
 
-* A Kubernetes cluster with `kubectl` connected to it.
-* A route-able IP address for ingress on the cluster.
-* A "star" domain pointing to the cluster route-able IP.
+* Version: `1.19+`
+* A Kubernetes cluster with your local `kubectl` with admin authentication and authorization.
+* A wildcard DNS registry pointing to the cluster route-able IP. eg: `*.akashian.io`
```
hydrogen18, line 37: Maybe we should just go ahead and change this to indicate "a wildcard A Record resolving to the cluster route-able IP"

Ropes: :100: ty

_docs/examples/provider/README.md
```diff
@@ -153,17 +166,24 @@ EOF
...
 ### Configure Kubernetes Akash Namespace
 
 Create the `akash-services` namespace for running Provider.
 
 ```sh
-kubectl apply -f https://raw.githubusercontent.com/ovrclk/akash/master/_docs/kustomize/namespace.yaml
+kubectl apply -f https://raw.githubusercontent.com/ovrclk/akash/master/_docs/kustomize/networking/namespace.yaml
+```
+
+Local option:
```
boz, line 184: I'm not sure this makes sense to have here.

_docs/examples/provider/kustomization.yaml
```diff
@@ -1,15 +1,16 @@
 bases:
-  - github.com/ovrclk/akash/_docs/kustomize/akash-provider?ref=master
-  # - ../../kustomize/akash-provider
+  - ../../kustomize/akash-provider
```
boz, line 2: this requires you to be doing this in-repo, right?

_docs/examples/provider/kustomization.yaml
```diff
@@ -1,15 +1,16 @@
...
 
+images:
+  - name: ovrclk/akash
     ##
-    # akashctl version
+    # akash version
     ##
 
-    newTag: latest
+    newTag: 0.7.21-doubletap.rc21
```
boz, line 13: have we released `0.9.x` yet?

_docs/kustomize/akash-provider/kustomization.yaml
```diff
@@ -19,7 +19,7 @@ configMapGenerator:
     literals:
       - home=/home
       - from=main
-      - node=tcp://akashd:26657
+      - node=http://akashd:26657
```
boz, line 22: no longer `tcp://`?

_docs/provider.yaml
```diff
@@ -2,5 +2,3 @@ host: http://localhost:3001
 attributes:
   - key: region
     value:  us-west
-  - key: tier
```
boz, line 5: it's for illustration

_run/common-kind.mk
```diff
@@ -75,10 +75,12 @@ kind-ingress-setup:
...
+
 .PHONY: kind-cluster-delete
 kind-cluster-delete:
 	kind delete cluster --name "$(KIND_NAME)"
 
 .PHONY: kind-cluster-clean
 kind-cluster-clean:
-	kubectl delete ns -l akash.network
+	kubectl delete ns -f akash-services leases akash.network
+	# TODO: 'akash.network' used?
```
boz, line 86: `-l akash.network` means match all namespaces (here) that have a label named `akash.network`.

_run/single/kind-config.yaml
```diff
@@ -2,6 +2,7 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
+  image: kindest/node:v1.19.1@sha256:98cf5288864662e37115e362b23e4369c8c4a408f99cbc06e58ac30ddc721600
```
boz, line 5: can you use just `v1.19.1`?

boz (commented):
looks awesome

_docs/examples/provider/kustomization.yaml
Ropes: We do! Now the question; are we safe to set it to `latest`?

_docs/provider.yaml
Ropes: That's true, I'll convert it back to be more of an example than an actual declaration.

_docs/examples/provider/kustomization.yaml
Ropes: Yeah, I'll change it to default to the github URL, but leaving the local path is usefully for quickly getting back to a debugging environment.

_run/common-kind.mk
Ropes: Ahhh I see, that works

_run/single/kind-config.yaml
Ropes: Following the [kind docs](https://kind.sigs.k8s.io/docs/user/quick-start/#setting-kubernetes-version). They suggest including the sha sum.

_docs/kustomize/akash-provider/kustomization.yaml
Ropes: New sdk changes... seems like a weird config artifact that might get flipped back.

_docs/examples/provider/README.md
```diff
@@ -25,33 +32,45 @@ mkdir akash-demo && cd akash-demo
...
 ### Akash Network
 
 We'll be connecting to the testnet.
 
 ```sh
-export AKASHCTL_NODE=tcp://rpc.akashtest.net:26657
-export AKASHCTL_CHAIN_ID=testnet-v4
+export AKASH_NODE=tcp://rpc-edgenet.akashdev.net:26657
+export AKASH_CHAIN_ID=edgenet
+export AKASH_KEYRING_BACKEND=test
```
boz, line 68: Are these not read from the environment later on?  I see the flags are used everywhere.

hydrogen18:
@boz the way that cosmos SDK is built it is mandatory to set `--chain-id` on the command line

boz:
@hydrogen18 what about these?  if not, is the idea to just keep it consistent for now?

* [ ] `--keyring-backend`
* [ ] `--home`
* [ ] `--node`

It's really a bummer.  Maybe there are some quickfix solutions #908

hydrogen18:
For now the easiest solution is to just specify it everywhere, including examples so anyone looking at them nows to set them. In the kustomize scripts we can do stuff like `--chain-id=$AKASH_CHAIN_ID` to work around this until we fix it.

Ropes:
Yeah I'm going to second @hydrogen18, the envvar configuration doesn't work, the required flag burns us. For now unfortunately this verbose CLI configuration is the only way.

To make it messier, we can't just set a flag like `--keyring-backend` on every command, because the CLI will throw an error if it's supplied to a command which does not consume it.

boz:
> In the kustomize scripts we can do stuff like --chain-id=$AKASH_CHAIN_ID to work around this until we fix it.

That's what is being done in the guide right now.

> Yeah I'm going to second @hydrogen18, the envvar configuration doesn't work, the required flag burns us. For now unfortunately this verbose CLI configuration is the only way.

Again, it's only for `--chain-id`, right?

> To make it messier, we can't just set a flag like --keyring-backend on every command, because the CLI will throw an error if it's supplied to a command which does not consume it.

It's not even being set on all key commands right now (see the export section).

Anyways, was any progress made on this today?

Ropes:
I need to rebase this, but otherwise are we good to merge?

boz (approved):

_docs/examples/provider/README.md
```diff
@@ -153,17 +166,24 @@ EOF
...
+(cat key-pass.txt; cat key-pass.txt) | akash --home=$AKASH_HOME keys export provider1 2> key.txt
 ```
 
 ### Configure Kubernetes Akash Namespace
 
 Create the `akash-services` namespace for running Provider.
 
 ```sh
-kubectl apply -f https://raw.githubusercontent.com/ovrclk/akash/master/_docs/kustomize/namespace.yaml
+kubectl apply -f https://raw.githubusercontent.com/ovrclk/akash/master/_docs/kustomize/networking/namespace.yaml
```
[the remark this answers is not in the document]
Ropes: I tend to agree with your point, however this `namespace.yaml` might be one of the few exceptions since it probably won't ever change. 

For example files like `deployment.yaml`, or `kustomization.yaml` which will have version specific changes, are more fragile to break referencing master.

hydrogen18 (approved):

PR #920, opened 2020-11-19
Reinstate validation

hydrogen18:
This re-enables validation basically everywhere I could think of. I had to move validation of a deployment to live in the same directory as the deployment files, otherwise we get a dependency loop.


provider/manifest/service.go
```diff
@@ -190,14 +191,12 @@ loop:
...
 		case req := <-s.mreqch:
-
-			// TODO: validate manifest according to rules in github.com/ovrclk/akash/validation
-			// if err := validation.ValidateManifest(req.value.Manifest); err != nil {
-			// 	h.session.Log().Error("manifest validation failed",
-			// 		"err", err, "deployment", req.value.Deployment)
-			// 	req.ch <- err
-			// 	break
-			// }
+			if err := validation.ValidateManifest(req.value.Manifest); err != nil {
```
Ropes, line 194: :raised_hands:

validation/manifest_validation_errors.go
```diff
@@ -0,0 +1,16 @@
+package validation
+
+import (
+	"errors"
+)
+
+var ErrInvalidManifest = errors.New("invalid manifest")
+var ErrGroupContainsNoServices = errors.New("The deployment group contains 0 services")
```
Ropes, line 8: Not entirely sure why, but Go community does prefer all lowercase error messages.

x/deployment/types/order_bid_duration_validation.go
```diff
@@ -0,0 +1 @@
+package types
```
Ropes, line 1: This file necessary?

x/deployment/types/group_pricing_validation.go
```diff
@@ -0,0 +1,78 @@
...
+func validateUnitPricing(rg Resource) error {
+	if !rg.GetPrice().IsValid() {
+		return errors.Errorf("error: invalid price object")
+	}
+
+	if rg.Price.Amount.GT(sdk.NewIntFromUint64(uint64(validationConfig.MaxUnitPrice))) {
+		return errors.Errorf("error: invalid unit price (%v > %v fails)", validationConfig.MaxUnitPrice, rg.Price)
+	}
+
+	if rg.Price.Amount.GT(sdk.NewIntFromUint64(uint64(validationConfig.MaxUnitPrice))) {
```
Ropes, line 64: I think this should be *L*ess*T*han() based on the error message, and that it's the same comparison above?

x/deployment/types/validation_config.go
```diff
@@ -0,0 +1,51 @@
...
+	MaxGroupUnits int
+
+	MaxGroupCPU     int64
+	MaxGroupMemory  int64
+	MaxGroupStorage int64
+
+	MinGroupMemPrice int64
+	MaxGroupMemPrice int64
+}{
+	MaxUnitCPU:     500,
```
Ropes, line 30: I'm bleary eyed, but a little confused on this MaxUnitCPU. Is this supposed to be 500 CPU cores? Otherwise this would translate into 500milliCPUs(kubernetes-ism), and thus not a lot of processing power.

Ropes (commented):
Thanks for crushing this! 

Only a couple questions and suggestions.

provider/bidengine/order.go
```diff
@@ -266,7 +265,7 @@ func (o *order) shouldBid(group *dtypes.Group) bool {
 		return false
 	}
 
-	if err := validation.ValidateDeploymentGroup(group.GroupSpec); err != nil {
+	if err := group.GroupSpec.ValidateBasic(); err != nil {
```
boz, line 268: cool.  I think that this might not be needed anymore since everything has the same values, but what the heck.

provider/cluster/kube/builder.go
```diff
@@ -266,7 +267,7 @@ func (b *deploymentBuilder) container() corev1.Container {
 	// TODO: this prevents over-subscription.  skip for now.
 
 	for _, env := range b.service.Env {
-		parts := strings.Split(env, "=")
+		parts := strings.SplitN(env, "=", 2)
```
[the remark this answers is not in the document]
boz: This is better, otherwise you can't have values with `=`.  Thank you.

provider/cluster/kube/client.go
```diff
@@ -189,11 +185,11 @@ func (c *client) Deploy(ctx context.Context, lid mtypes.LeaseID, group *manifest
...
 
 		for expIdx := range service.Expose {
-			expose := &service.Expose[expIdx]
-			if !shouldExpose(expose) {
+			expose := service.Expose[expIdx]
+			if !util.ShouldExpose(expose) {
 				continue
 			}
-			if err := applyIngress(ctx, c.kc, newIngressBuilder(c.log, c.settings, c.host, lid, group, service, expose)); err != nil {
+			if err := applyIngress(ctx, c.kc, newIngressBuilder(c.log, c.settings, c.host, lid, group, service, &service.Expose[expIdx])); err != nil {
```
boz, line 192: `&expose`?  hmm is there a reason why this takes a pointer?  generally prefer pass-by-value to contain any possible mutation.

provider/cluster/kube/client.go
```diff
@@ -433,22 +429,6 @@ func (c *client) Inventory(ctx context.Context) ([]ctypes.Node, error) {
 		nodes = append(nodes, cluster.NewNode(knode.Name, resources))
 	}
 
-	if os.Getenv("AKASH_PROVIDER_FAKE_CAPACITY") == "true" {
```
boz, line 436: thanks and sorry.

validation/manifest.go
```diff
@@ -10,21 +14,116 @@ import (
 
 // ValidateManifest does validation for manifest
 func ValidateManifest(m manifest.Manifest) error {
+	if m == nil {
```
boz, line 17: `len(m) == 0`?

validation/manifest.go
```diff
@@ -10,21 +14,116 @@ import (
...
+	}
+
 	return nil
 }
 
-// ValidateManifestWithGroupSpecs does validation for manifest with group specifications
+var hostnameRegex = regexp.MustCompile(`^[[:alnum:],-,\.]+$`)
+
+func isValidHostname(hostname string) bool {
+	return hostnameRegex.MatchString(hostname)
```
boz, line 124: maybe also

* max length: 255
* min length: I don't know, 3?
* require host + . + tld?

It'd be cool if there was a hostname validation library that we could use.

validation/manifest.go
```diff
@@ -33,79 +132,111 @@ func ValidateManifestWithGroupSpecs(m *manifest.Manifest, gspecs []*dtypes.Group
...
 			}
 
+			// Update the value stored in the list
 			mlist[idx] = mrec
 
+			// If the deployment group is fulfilled then break out and
+			// move to the next deployment
 			if drec.Count == 0 {
-				continue mainloop
+				continue deploymentGroupLoop
```
boz, line 213: I'm sorry you had to read through this logic :joy: .  Thank you for annotating it.

validation/manifest_validation_errors.go
```diff
@@ -0,0 +1,16 @@
...
+var ErrServiceNameEmpty = errors.New("The service name is the empty string")
+var ErrServiceImageEmpty = errors.New("The service image naem is empty")
+var ErrServiceEnvVarEmptyName = errors.New("An environmental variable has an empty name")
+var ErrServiceCountIsZero = errors.New("The service count is zero")
+var ErrServiceExposePortZero = errors.New("The service port is zero")
+var ErrServiceExposeInvalidHostname = errors.New("The service hostname is not valid")
+
+var ErrManifestCrossValidation = errors.New("manifest cross validation error")
```
boz, line 16: Let's stick to grouping things in `var ()` and `const ()` sections.

x/deployment/types/validation_config.go
boz: good point I don't remember myself.

validation/manifest_validation_errors.go
troian: yeah, i think it relates to parsing them

provider/bidengine/pricing.go
```diff
@@ -171,12 +168,15 @@ func calculatePriceRange(gspec *dtypes.GroupSpec) (sdk.Coin, sdk.Coin) {
 
 	rmax := gspec.Price()
 
+	const minGroupMemPrice = int64(50)
+	const maxGroupMemPrice = int64(1048576)
```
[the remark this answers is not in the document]
hydrogen18: They were part of the limits read from the environment, that I've eliminated. Long term I think we are doing away with this on the network. This only controls the "randomRange" pricing which is mostly a test feature at this point, so I just copied the constants over.

provider/cluster/kube/builder.go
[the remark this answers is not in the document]
hydrogen18: @Ropes My understanding is that `string.SplitN("foo", "=", 2)` should return a slice with a length of 1

validation/manifest.go
```diff
@@ -10,21 +14,116 @@ import (
 
 // ValidateManifest does validation for manifest
 func ValidateManifest(m manifest.Manifest) error {
+	if m == nil {
+		return fmt.Errorf("%w: manifest is nil", ErrInvalidManifest)
```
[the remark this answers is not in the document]
hydrogen18: I was basing this off of this 

https://github.com/Djarvur/go-err113

Is there an advantage to `errors.Wrap` ? My understand is the special `%w` does this.

validation/manifest_validation_errors.go
hydrogen18: Alright I'll change this.

x/deployment/types/group_pricing_validation.go
hydrogen18: I think I just re-enabled this code, I'll double check it.

x/deployment/types/validation_config.go
hydrogen18: I will double check this, I don't know as I just refactored this code.

validation/manifest.go
hydrogen18: That should work, but its a little strange since it isn't immediately obvious that `Manifest` is a slice.

validation/manifest.go
hydrogen18: I did not find a library but I can put some additional limits on this.

x/deployment/types/validation_config.go
Ropes: Okay, I think this is counted in milliCPU, so 1000 = 1CPU core. So MaxUnitCPU might best be 64000?

provider/cluster/kube/client.go
hydrogen18: I'm not actually sure why we were passing around `*ServiceExpose`. I eliminated it in a few places in the codebase because it was causing some warnings.

x/deployment/types/validation_config.go
hydrogen18: OK, I'm changing it to 64000

x/deployment/types/group_pricing_validation.go
boz: looks like it's checking if the price is greater than the max price; I think this is probably okay.

x/deployment/types/validation_config.go
boz: I think it was actually intentional - we only wanted to allow half a cpu share at the time.  Let's cap it to, I don't know, 1000 = 1 cpu share for now - this is for a single instance, 64 is too large IMHO; you'd need a box with 64 unused cpus to fulfill it.

validation/manifest.go
boz: I agree.  okay.  but are we checking for `len(m) == 0` another way?

x/deployment/types/validation_config.go
hydrogen18: Ok I moved it to 1000

x/deployment/types/group_pricing_validation.go
hydrogen18: If you look at the line above that one is checking "price > maxPrice" so I changed this one to check to "price < minPrice"

integration/e2e_test.go
```diff
@@ -44,8 +44,6 @@ func (s *IntegrationTestSuite) SetupSuite() {
 	cfg := testutil.DefaultConfig()
 	cfg.NumValidators = 1
 	cfg.MinGasPrices = ""
-	cfg.BondDenom = "stake"
```
Ropes, line 47: :raised_hands:

Ropes (approved):
lgtm

Only thing that I disagree with is `MaxUnitCPU`, if we cap it at 1 CPU, then that's the cap for a container, and I can guarantee there are customers who have workloads which will consume more than 1 CPU, or would like that potential, but we can adjust it later.

boz (approved):

PR #924, opened 2020-11-23
docs(sdl): document audited attributes

troian:


boz (approved):

_docs/sdl.md
```diff
@@ -118,15 +118,23 @@ Example:
...
 configuration for each [compute profile](#profilescompute) that will be used within the datacenter.
+It also specifies optional list of signatures of which tenants expects audit of datacenter attributes.
 
 Example:
 
 ```yaml
 westcoast:
   attributes:
     region: us-west
+  signedBy: # optional, allOf has higher priority, anyOf is ignored if there is at least one key in allOf
```
boz, line 131: Not sure there should be any precedence rules required - any reason why one requirement should depend on the other?  If both are populated I'd expect it to effectively be an "AND".

PR #927, opened 2020-11-23
test(provider): Fix a race in the bidengine

hydrogen18:
There is a race in one of the tests I added. It passes for me locally and also on @troian 's machine. It is failing in CI sometimes.

My solution for now is to add a channel that is only used during the unit test to avoid the race condition.


Ropes (approved):
Nice use of nil channels, lgtm

PR #932, opened 2020-11-25
refactor(client): create serial broadcaster

### LINKED ISSUE #605: account sequence handling in long-lived processes
The bid engine [submits messages to the chain](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/provider/bidengine/order.go#L187-L191) using an [sdk utility](https://github.com/ovrclk/akash/blob/082cb048d367cadebc16de5efb592622f0de2700/client/client.go#L81-L84) that looks up the account's sequence number when signing the transaction.

There are all kinds of race conditions with this:

* what if the provider is bidding on two leases at once?
* what if two bids are sequential but end up in the same block?
* etc...

We need a way to gracefully handle concurrent transaction submissions.  We have [previously](https://github.com/ovrclk/akash/blob/stable/txutil/client.go#L79-L87) used a simple counter, but this runs into problems when the provider account is submitting transactions through any other means (operator cancelling leases through the CLI, for instance).

edit: solution suggested by @jackzampolin is to simply [retry](https://github.com/avast/retry-go) events.

boz:
* install `mockery` correctly
* remove unused passphrase handling

fixes #605


client/broadcaster/client.go
```diff
@@ -0,0 +1,70 @@
...
+		return err
+	}
+
+	response, err := cctx.BroadcastTxSync(bytes)
+	if err != nil {
+		return err
+	}
+
+	if response.Code != 0 {
+		return ErrBroadcastTx
```
hydrogen18, line 66: I would prefer something like `fmt.Errorf("%w: response code %d", ErrBroadcastTx, response.Code)` here so we know both that we have an error & what the code is. This assumes of course that tendermint error codes are helpful, which I don't know to be true.

client/broadcaster/serial.go
```diff
@@ -0,0 +1,167 @@
...
+		case <-c.lc.ShuttingDown():
+			return
+		case <-ticker.C:
+
+			// query sequence number
+			_, seq, err := c.cctx.AccountRetriever.
+				GetAccountNumberSequence(c.cctx, c.info.GetAddress())
+
+			// send to main loop if no error
+			if err != nil {
```
hydrogen18, line 158: Isn't this statement backwards? Shouldn't it be `if err == nil` based on the comment ?

make/setup-cache.mk
```diff
@@ -76,7 +76,7 @@ kubetypes-deps-install:
 		"$(shell go env GOPATH)/src/k8s.io/code-generator"
 
 devdeps-install: kubetypes-deps-install
-	$(GO) install github.com/vektra/mockery/.../
+	$(GO) get     github.com/vektra/mockery/.../
```
hydrogen18, line 79: I think this would install the wrong version of mockery, we're on 1.x from what I recall

boz: The version is in `go.mod`, I think it works.  I also tho that they're using a `v2` path for v2.   Could be wrong tho.

client/broadcaster/serial.go
boz: 100%, thank you

PR #934, opened 2020-11-25
test(provider): fix race in test

hydrogen18:
I think in the process of fixing one race, I introduced another one. This refactor should fix that. I also made some changes suggested by @troian  to prevent non obvious timeouts in CI.

### DIFF
--- a/testutil/channel_wait.go
+++ b/testutil/channel_wait.go
@@ -1,0 +1,39 @@
+package testutil
+
+import (
+	"reflect"
+	"testing"
+	"time"
+)
+
+func ChannelWaitForValueUpTo(t *testing.T, waitOn interface{}, waitFor time.Duration) interface{} {
+	cases := make([]reflect.SelectCase, 2)
+	cases[0] = reflect.SelectCase{
+		Dir:  reflect.SelectRecv,
+		Chan: reflect.ValueOf(waitOn),
+		Send: reflect.Value{},
+	}
+
+	delayChan := time.After(waitFor)
+
+	cases[1] = reflect.SelectCase{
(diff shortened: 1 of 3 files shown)


testutil/channel_wait.go
```diff
@@ -0,0 +1,35 @@
+package testutil
+
+import (
+	"reflect"
+	"testing"
+	"time"
+)
+
+func ChannelWaitForValue(t *testing.T, waitOn interface{}) interface{} {
```
troian, line 9: can we have here options interface allowing caller to set custom timeout value?

hydrogen18: Sure, I will have two functions in this file. One with a default value set and the other where the caller can set how long to wait.

troian: you can do one function with options pattern like [here](https://github.com/VolantMQ/volantmq/blob/master/connection/writerOptions.go)

hydrogen18: I don't really see an advantage to doing that for what is a single option. It's just going to create a complex way of doing an obvious thing.

boz (approved):

testutil/channel_wait.go
```diff
@@ -0,0 +1,39 @@
...
+
+	delayChan := time.After(waitFor)
+
+	cases[1] = reflect.SelectCase{
+		Dir:  reflect.SelectRecv,
+		Chan: reflect.ValueOf(delayChan),
+		Send: reflect.Value{},
+	}
+
+	idx, v, ok := reflect.Select(cases)
```
Ropes, line 25: TIL :exploding_head:

Ropes (approved):

PR #935, opened 2020-11-25
feat(provider): Use per megabyte for scale of storage & memory pricin…

hydrogen18:
This makes the memory & storage pricing per megabyte instead of per byte. I added some additional tests to cover the logic I implemented to get this working.


boz (approved):
I think we should do gigabytes for storage at least, but it can be done later

boz (approved):

PR #942, opened 2020-11-26
fix(node): Insert endpoints in the deployment groups for each service…

hydrogen18:
For each declared service that becomes a NodePort in kubernetes an `Endpoint` needs to be declared. This informs the bidding price logic and is necessary for manifest & deployment cross validation logic to pass.


Ropes (approved):
nice catch, lgtm

PR #948, opened 2020-11-27
feat(kustomize): Allow use of a python 3 shell script for pricing in …

hydrogen18:
This changes the following

1. Adds python3 to the docker container for akash
2. Adds another env. var. to kustomize
3. If the env. var. is set when the provider is launched the contents are placed into a shell script and the provider is set to use that as the pricing strategy.

I ran this locally with kind and it worked

### DIFF
--- a/_build/Dockerfile.akash
+++ b/_build/Dockerfile.akash
@@ -1,5 +1,7 @@
 FROM debian:buster
 
+RUN apt-get update && apt-get --no-install-recommends -y install python3 && apt-get clean
+
 COPY ./akash /bin/
 
 EXPOSE 26656 26657 26658
--- a/_docs/kustomize/akash-provider/run.sh
+++ b/_docs/kustomize/akash-provider/run.sh
@@ -2,13 +2,13 @@
 
 set -e
 
-if [ -n "${AKASH_BID_PRICE_SCRIPT_CONTENTS}" ]; then
+pricing_script_path="/boot/pricing.sh"
+pricing_script_size=$(stat -c%s ${pricing_script_path?})
+
+if [ $pricing_script_size > 0 ]; then
   echo 'setting up bid price script'
-  echo "${AKASH_BID_PRICE_SCRIPT_CONTENTS}" > /bin/pricing.sh
-  chmod 555 /bin/pricing.sh
-  AKASH_BID_PRICE_SCRIPT_PATH=/bin/pricing.sh
-  export AKASH_BID_PRICE_SCRIPT_PATH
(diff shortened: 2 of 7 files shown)


_build/Dockerfile.akash
```diff
@@ -1,5 +1,7 @@
 FROM debian:buster
 
+RUN apt-get update && apt-get -y install python3 && apt-get clean
```
troian, line 3: add `--no-install-recommends`

hydrogen18: Done!

_docs/kustomize/akash-provider/run.sh
```diff
@@ -1,7 +1,16 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 
+pricing_script_path="/boot/pricing.sh"
+pricing_script_size=$(stat -c%s ${pricing_script_path?})
+
+if [ $pricing_script_size > 0 ]; then
```
Ropes: If the script is present, this will default to always running the script right?

_docs/kustomize/akash-provider/run.sh
```diff
@@ -1,7 +1,16 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 
+pricing_script_path="/boot/pricing.sh"
+pricing_script_size=$(stat -c%s ${pricing_script_path?})
+
+if [ "$pricing_script_size" -gt 0 ]; then
```
Ropes, line 8: I think this might be too aggressive. If I'm understanding this right, if the shell script is present, then other configuration flags are ignored?

I think it will cause confusion if we include the script in the default kustomize, and then don't provide a way to use other configuration methods even when the script is present.

hydrogen18: OK, then what is suggested here as an alternative?

Ropes: I'd propose removing these `run.sh` changes.(sorry for the slack confusion)

Then providers could use the normal means of kustomize -> configmap settings and overrides to configure the provider, without the need to also know about this behavior here. 

If you want to use the shell script, it's configured in the same place as the other bidding behavior:
```
pricing_script_path="/boot/pricing.sh"
bid_price_strategy="shellScript"
```

Seem reasonable? That way the script can exist in the configuration and container, but not have a hard change in behavior based on existance. It will make tuning the provider's configuration easier, and I think less prone to mistakes.

hydrogen18:
@boz testing this got lost in the noise I think. We still want to merge this, correct?

PR #949, opened 2020-11-27
feat(provider): Set limits when configuring containers in kubernetes

hydrogen18:
This bumps up the limit & enforces it on kubernetes containers. There is no such feature as a storage limit, only ephemeral storage limits. We don't create actual kubernetes storage that I see anywhere.


hydrogen18:
After shelling into the kube container and doing `cat /dev/zero > file` into did this  

evicted_container

looks like it is working

PR #954, opened 2020-11-29
fix(provider): Remove nodeport check in the deployment monitor, only …

hydrogen18:
We don't need to check the nodeport presence on a service it isn't strictly necessary and has some bugs. 

My deployment seems to be stable after this change

Screenshot from 2020-11-28 17-58-07


Ropes (approved):

boz (approved):
🚀

PR #965, opened 2020-12-01
network policies are blocking inter-pod connections

Ropes:
This change simply disables producing the list of Network policies to
apply, so the builder will effectively do nothing, and not quarantine
entwork traffic.

This change should be reverted once accurate network polices isolate
namespaces without blocking traffic.


boz (approved):

boz:
superseded by #966

PR #982, opened 2020-12-02
Cleanup provider error response handling

hydrogen18:
Changes

1. Suppress showing the usage when we have an internal failure (like 404 from the remote). Still shows up if the user does a `help`command
2. Refactor the way the gateway router handles errors & check for all the special cases I could find
3. Add tests around the gateway router
4. Refactor the way the provider client handles errors from the remote. They get returned to cobra, but beforehand the actual HTTP body is captured and displayed to the user. This should contain useful information from the provider i.e. `send-manifest` fails due to an invalid manifest.
5. Fix some errors that I think came in from master?
6. Rename `ensureManger` to `ensureManager` !

### DIFF
--- a/client/broadcaster/serial.go
+++ b/client/broadcaster/serial.go
@@ -27,7 +27,7 @@ var (
 
 	// errors are of the form:
 	// "account sequence mismatch, expected 25, got 27: incorrect account sequence"
-	recoverRegexp = regexp.MustCompile("^account sequence mismatch, expected (\\d+), got (\\d+):")
+	recoverRegexp = regexp.MustCompile(`^account sequence mismatch, expected (\\d+), got (\\d+):`)
 )
 
 type SerialClient interface {
--- a/client/broadcaster/serial.go
+++ b/client/broadcaster/serial.go
@@ -27,7 +27,7 @@ var (
 
 	// errors are of the form:
 	// "account sequence mismatch, expected 25, got 27: incorrect account sequence"
-	recoverRegexp = regexp.MustCompile(`^account sequence mismatch, expected (\\d+), got (\\d+):`)
+	recoverRegexp = regexp.MustCompile(`^account sequence mismatch, expected (\d+), got (\d+):`)
 )
 
 type SerialClient interface {
--- a/provider/gateway/client.go
+++ b/provider/gateway/client.go
@@ -83,6 +83,17 @@ func (c *client) Status(ctx context.Context, host string) (*provider.Status, err
 	return &obj, nil
 }
 
+func createClientResponseErrorIfNotOK(resp *http.Response, responseBuf *bytes.Buffer) error {
+	if resp.StatusCode == http.StatusOK {
+		return nil
+	}
+
+	return ClientResponseError{
+		Status:  resp.StatusCode,
+		Message: responseBuf.String(),
+	}
+}
+
 func (c *client) SubmitManifest(ctx context.Context, host string, mreq *manifest.SubmitRequest) error {
 	uri, err := makeURI(host, submitManifestPath(mreq.Deployment))
 	if err != nil {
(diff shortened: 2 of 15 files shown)


provider/manifest/service.go
```diff
@@ -106,9 +106,19 @@ func (s *service) Submit(ctx context.Context, mreq *SubmitRequest) error {
 	case <-ctx.Done():
 		return ctx.Err()
 	case s.mreqch <- req:
-		return <-ch
```
hydrogen18, line 109: @troian  I think this would wait for ever for something to come back on the channel ? I refactored it below. I am unsure if this is the problem you were hitting however.

client/broadcaster/serial.go
```diff
@@ -27,7 +27,7 @@ var (
 
 	// errors are of the form:
 	// "account sequence mismatch, expected 25, got 27: incorrect account sequence"
-	recoverRegexp = regexp.MustCompile("^account sequence mismatch, expected (\\d+), got (\\d+):")
+	recoverRegexp = regexp.MustCompile(`^account sequence mismatch, expected (\\d+), got (\\d+):`)
```
troian, line 30: 🚀

boz: not sure this makes a difference here :shrug: 

does `\` still need to be escaped?  I guess so if the tests are passing.

provider/gateway/client.go
```diff
@@ -92,13 +104,20 @@ func (c *client) SubmitManifest(ctx context.Context, host string, mreq *manifest
 	if err != nil {
 		return err
 	}
-	_, _ = io.Copy(ioutil.Discard, resp.Body)
+	responseBuf := &bytes.Buffer{}
+	_, err = io.Copy(responseBuf, resp.Body)
+	if err != nil {
+		return err
```
boz, line 110: I believe we always want to `resp.Body.Close()` - perhaps do it as a `defer` before the `io.Copy`.  Same goes for later instances of this change in this PR.

provider/gateway/client.go
```diff
@@ -194,7 +221,19 @@ func (c *client) ServiceLogs(ctx context.Context,
...
+	if err == websocket.ErrBadHandshake {
+		buf := &bytes.Buffer{}
+		_, err = io.Copy(buf, response.Body)
+		if err != nil {
+			return nil, err
+		}
+		return nil, ClientResponseError{
+			Status:  response.StatusCode,
+			Message: buf.String(),
+		}
```
boz, line 234: perhaps encapsulate this pattern in a utility function?

provider/manifest/service.go
boz: Yes, it would.  Hope this wasn't the problem, but it could be.

The proposed fix would especially help if there was deadline set on the context in the caller (web server).

provider/manifest/service.go
```diff
@@ -198,7 +208,7 @@ loop:
 				break
 			}
 
-			manager, err := s.ensureManger(req.value.Deployment)
+			manager, err := s.ensureManager(req.value.Deployment)
```
boz, line 211: omg I never noticed that.  I didn't even follow what you were saying on slack about it.

boz (changes requested):
Looks great!  A couple comments that would be nice to address:

1. always close `resp.Body`
1. make a utility function for constructing the error from `resp.Body`

provider/gateway/client.go
hydrogen18: I have refactored everything to defer

client/broadcaster/serial.go
hydrogen18: It's been changed now. This change comes from the code linters requirements btw

provider/gateway/client.go
hydrogen18: This has been done

Ropes (approved):
I'm afraid of potentially leaking sensitive information from k8s logs, but you've created nice control mechanisms around returning data to the user, and inputs declaring the query against the API are still controlled by the provider.

troian (approved):

PR #986, opened 2020-12-07
feat(provider): Add deployment information to running containers

hydrogen18:
This just adds in the deployment information for the running containers. Not required but can be useful in some circumstances.

### DIFF
--- a/provider/cluster/kube/builder.go
+++ b/provider/cluster/kube/builder.go
@@ -297,11 +297,11 @@ func (b *deploymentBuilder) container() corev1.Container {
 }
 
 const (
-	envVarAkashGroupSequence = "AKASH_GROUP_SEQUENCE"
-	envVarAkashDeploymentSequence = "AKASH_DEPLOYMENT_SEQUENCE"
-	envVarAkashOrderSequence = "AKASH_ORDER_SEQUENCE"
-	envVarAkashOwner = "AKASH_OWDER"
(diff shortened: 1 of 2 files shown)


provider/cluster/kube/builder.go
```diff
@@ -293,6 +296,34 @@ func (b *deploymentBuilder) container() corev1.Container {
 	return kcontainer
 }
 
+const envVarAkashGroupSequence = "AKASH_GROUP_SEQUENCE"
+const envVarAkashDeploymentSequence = "AKASH_DEPLOYMENT_SEQUENCE"
+const envVarAkashOrderSequence = "AKASH_ORDER_SEQUENCE"
+const envVarAkashOwner = "AKASH_OWDER"
+const envVarAkashProvider = "AKASH_PROVIDER"
+const envVarAkashClusterPublicHostname = "AKASH_CLUSTER_PUBLIC_HOSTNAME"
```
boz, line 304: use
```go
const (
  foo = "val"
  bar = "val"
)
```
format please

boz (changes requested):
Looks great.  One small format change request.

provider/cluster/kube/builder.go
hydrogen18: I need to get our code linter to prefer this, I keep forgetting it.

troian (approved):

boz (approved):

PR #992, opened 2020-12-10
Allow testing via K8S running in vagrant

hydrogen18:
I have a companion PR here https://github.com/ovrclk/kubespray/pull/3

When both of these are put together it does the following

1. Gives up the ability to provision a single master K8S cluster in Vagrant
2. Configures the cluster using our example
3. Gives us a way to automatically run the e2e tests against the K8S cluster

The K8S cluster is a single master with two other nodes. I'm running it in VirtualBox, you need about 6 GB of free memory for this to work. Theoretically we can point Vagrant at a VMWare cluster or AWS if we want to run e2e tests against a huge cluster for some reason.

The main drawback here is we can't run this automatically inside of Github CI.

Once this is merged we need to work on adding more e2e tests as there is substantially more we can cover in those tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ GOBIN                 := $(shell go env GOPATH)/bin
 
 KIND_APP_IP           ?= $(shell make -sC _run/kube kind-k8s-ip)
 KIND_APP_PORT         ?= $(shell make -sC _run/kube app-http-port)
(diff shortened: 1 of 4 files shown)


boz (approved):
:rocket:

troian (approved):

PR #1000, opened 2020-12-14
fix(audit): rename validator to auditor

### LINKED ISSUE #970: audit: field names seem backwards
In the audit below, the provider address is `akash1ds82uk3pzawavlasc2qd88luewzye2qrcwky7t` and the signer address is `akash1vz375dkt0c60annyp6mkzeejfq0qpyevhseu05`, but they are labeled oppositely.

```yaml
- attributes:
  - key: organization
    value: ovrclk.com
  - key: region
    value: us-west
  owner: akash1ds82uk3pzawavlasc2qd88luewzye2qrcwky7t
  validator: akash1vz375dkt0c60annyp6mkzeejfq0qpyevhseu05
```

troian:
fixes #970


boz (approved):
:rocket: :pray:

boz:
Maybe with the proto errors we should cut this as a new minor release and reset testnet?  We never had a non-rc `v0.9.0` :joy: .  What do you think we should do here @troian?

troian:
yah, good point about release. 0.9.0-rc deserved to be born 😁

Issue #999, opened 2020-12-14
the "provider send-manifest" response should indicate what the provider did with the manifest


hydrogen18:
Right now the `provider send-manifest` sends a manifest to a provider. If the manifest is valid, the provider responds with 200. We should also add information as to if the manifest matched a deployment that the provider has a lease for. It's easy to accidentally do `provider send-manifest` to the wrong provider & have nothing at all happen. This isn't a good experience for users.

boz:
@hydrogen18 should we close this?

hydrogen18:
I merged a fix for this, closing.

PR #1001, opened 2020-12-14
fix(provider): Reply with 404 whenever a manifest is received that has no leases

hydrogen18:
This changes the provider to send back 404 if the manifest is valid but there is no lease. I also added a test for this.


provider/manifest/manager.go
```diff
@@ -187,6 +188,8 @@ loop:
 
 			if err := result.Error(); err != nil {
 				m.log.Error("error fetching data", "err", err)
+				// Fetching data failed, all requests are now in an error state
+				m.fillAllRequests(err)
```
boz, line 192: great catch

provider/manifest/manager.go
```diff
@@ -259,8 +260,25 @@ func (m *manager) maybeScheduleStop() bool { // nolint:golint,unparam
...
+	}
+	m.requests = nil
+}
+
 func (m *manager) emitReceivedEvents() {
-	if m.data == nil || len(m.leases) == 0 || len(m.manifests) == 0 {
+	if len(m.leases) == 0 {
+		m.log.Debug("emit received events skips due to no leases", "data", m.data, "leases", len(m.leases), "manifests", len(m.manifests))
+		m.fillAllRequests(ErrNoLeaseForDeployment)
+		return
```
boz, line 279: Again good catch.  The idea here was that there are cases where the manifest might be shared before the lease is obtained, but I think it's better that we shed all of these historical ideas and simplify as much as possible.  To that end, I'd be open to slimming down this service significantly - I've never been happy with how difficult it is to follow (and I wrote it).

boz (approved):
Looks great.

One reason why it might be nice to store the whole request in `pendingRequests` is that we _could_ pass a context in with the request and the server loop _could_ discard requests whose context has already been cancelled.

I think it's fine as-is for now tho.

boz:
refs #999

troian (approved):

PR #1012, opened 2020-12-21
feat(provider): Add basic resource overcommit in the kubernetes cluster

hydrogen18:
The main issue I found while testing things using my local kubernetes cluster is that the provider continues to bid on things even after the available resources are consumed in the cluster. The provider was setting a limit only in the container pods (which also becomes the request in that case). Kubernetes enforces this, refusing to spinup new pods if available resources aren't there. The provider was continuing to bid because the "inventory" was determined by getting values from the metrics server. For almost all deployment, utilized CPU is less than the CPU limit enforced in the kubernetes pod. It's been changed to just measure the commit CPU by going through all the pods in the cluster.

After this changes, the provider just stops bidding once the available CPU is no longer enough to meet the requested amount. 

The tests in this wound up being a bit verbose, due to the way the kubernetes API works.

This makes the following changes

1. Measure requested resources in the kubernetes cluster for inventory, not the values reported by the metrics server
2. Specify requests & limits when creating a kubernetes pod. By default requests == limits. When overcommit is enabled, requests is less than limits
3. Add overcommit percentage options to the provider. By default overcommit is zero for all resources. Specifying an overcommit of 50 would result in 150% of available CPU in the cluster being requested by the pods
4. If a deployment fails, stop the associated manager immediately
5. If a deployment fails to be cleaned up, retry it with delay between retries up to 5 times. This should help avoid leaking resources (although I don't think that was a problem in testnet).

### DIFF
--- a/provider/cluster/kube/client.go
+++ b/provider/cluster/kube/client.go
@@ -2,6 +2,7 @@ package kube
 
 import (
 	"context"
+	"k8s.io/apimachinery/pkg/runtime"
 	"os"
 	"path"
 
@@ -11,6 +12,7 @@ import (
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -25,6 +27,7 @@ import (
 	"github.com/ovrclk/akash/provider/cluster"
 	"github.com/ovrclk/akash/types"
 	mtypes "github.com/ovrclk/akash/x/market/types"
+	"k8s.io/client-go/tools/pager"
 )
 
 var (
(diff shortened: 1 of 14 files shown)


provider/cluster/kube/client.go
```diff
@@ -401,99 +402,158 @@ func (c *client) ServiceStatus(ctx context.Context, lid mtypes.LeaseID, name str
...
 	knodes, err := c.kc.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
 	if err != nil {
 		return nil, err
 	}
 
-	retnodes := make(map[string]*corev1.Node)
+	listOptions := metav1.ListOptions{
+		FieldSelector: "status.phase!=Failed,status.phase!=Succeeded",
+	}
+	pods, err := c.kc.CoreV1().Pods("").List(ctx, listOptions)
```
boz, line 492: Do we need to paginate here?

Why is it better to do this than to use `metrics-server`?

provider/cluster/manager.go
```diff
@@ -126,17 +129,41 @@ loop:
...
+				// Check to see if the limit on the number of attepmts has been reached
+				if teardownAttempts == maxTeardownAttempts {
+					dm.log.Error("could not teardown lease")
+					break loop
+				}
+
+				// Wait before trying again
+				delay := time.Second * time.Duration(1+math.Exp2(float64(teardownAttempts)))
+				dm.log.Info("Trying teardown again", "delay", delay)
+				time.Sleep(delay)
```
boz, line 162: We can't `Sleep` in any of these state machine loops.  Can you separate this retry change into a different PR?

boz (commented):
Looking great!  Some comments inline.

provider/cluster/manager.go
boz: hmm, where you _can_ sleep is in the routines that are run asynchronously: https://github.com/ovrclk/akash/blob/b0c296d802cd4969c02efb7042edaaf546398d5e/provider/cluster/manager.go#L197-L205

provider/cluster/kube/client.go
hydrogen18: The metrics-server only returns actual CPU usage on a node, rather than the amount allocated to each pod in the node. For CPU this number is usually zero, which is why the provider would continue to bid on things it couldn't start up.

boz: I see, that makes sense.  Thanks.

I agree with an earlier sentiment that you wrote - it probably shouldn't do this so frequently.  Maybe it can be _yet another flag_ or just an environment variable (`caarlos0/env` or similar).

I think that if we do things this way that we might be able to simplify things a bit.  Previously we would:

1. use `metrics-server` to collect current unused capacity.
2. initially reserve capacity in _pending_; available would be `available = unused_capacity - pending`.
3. once a workload was "healthy" aka _active_, it was assumed that it was included in _used_ capacity from the metrics server; hence it would always be that `available = unused_capacity - pending`, not `available = unused_capacity - pending - active`.

All of this assumed that `Resources.Requests` was ignored, and as you found, that is not the case.

Nothing actionable here, just thinking out loud that maybe some complexity can be removed.

provider/cluster/kube/client.go
```diff
@@ -401,99 +402,158 @@ func (c *client) ServiceStatus(ctx context.Context, lid mtypes.LeaseID, name str
...
 func (c *client) Inventory(ctx context.Context) ([]ctypes.Node, error) {
+
 	// Load all the nodes
 	knodes, err := c.activeNodes(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	// Load all the node metrics
-	nodeMetrics, err := c.metc.MetricsV1beta1().NodeMetricses().List(ctx, metav1.ListOptions{})
```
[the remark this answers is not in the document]
hydrogen18: The script wait has been removed

provider/cluster/kube/client.go
hydrogen18: Pagination has been added here

provider/cluster/manager.go
hydrogen18: This has been removed from this PR

boz (approved):
Looks great!

I think that we should probably extract this into a separate service soon.

Using the k8s controller pattern where we listen to changes with a periodic reset would be best (there's a lot of support for this in the k8s libs) here.

But this is perfect for now.  So glad this was resolved!

PR #1013, opened 2020-12-21
fix(provider): Reject SDL describing invalid env. var. names

hydrogen18:
If an environmental variable name is invalid, this code change causes the SDL to fail validation.


boz (approved):
👍

boz (approved):
:ship:

PR #1015, opened 2020-12-22
feat(provider): Run containers as non-root user in deployments

hydrogen18:
For the testnet we disabled a bunch of stuff including restrictions around the user ID used inside a container.

This does the following

1. Require containers to run as non root
2. Set the container user ID (specified via CLI)
3. Set the container group ID (specified via CLI)
4. Set the pod security policy, including allowed user IDs, group IDs, and storage types allowed

**Note**: I have a PR for the ovrclk/demo-app project that I need to merge and then push a new tag into quay.io for this to pass. I don't have permission to do that yet. The automated tests won't pass until I do that

### DIFF
--- a/_docs/kustomize/networking/namespace.yaml
+++ b/_docs/kustomize/networking/namespace.yaml
@@ -4,5 +4,5 @@ metadata:
   name: akash-services
   labels:
     name: akash-services
-    akash.network: true
+    akash.network: "true"
     akash.network/name: akash-services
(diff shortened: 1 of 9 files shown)


boz:
If this requires a change to the image, then we probably can't use it right now.  

We can't require containers to run as non-root user.  99% of images don't do this right now AFAIK.  If it is a runtime configuration that doesn't impact running most containers then it might be doable.

hydrogen18:
Alright, I'll put this on the back burner then and keep working on the network security stuff for now.

hydrogen18:
Closing until we decide how & when we want to make this change.

PR #1016, opened 2020-12-23
Update SDK to v0.40.0

akhilkumarpilli:


hydrogen18:
I think to get the tests to pass you'll need to be on a version of cosmos that has this fix btw:

https://github.com/cosmos/cosmos-sdk/pull/8225

boz:
Can we go straight to `0.40.0` now?

boz:
The failing test looks like something that's easy to fix; let's try to get this merged asap @akhilkumarpilli

akhilkumarpilli:
> The failing test looks like something that's easy to fix; let's try to get this merged asap @akhilkumarpilli

Actually there is some issue with `sdk.Coin`. It is accepting decimals and normalizing it into integer. But when you compare normalized integer with expected integer, there hash values are different.

boz:
I see that @akhilkumarpilli , but I think that we should expect that case to fail.  That is how it looks on my branch, at least.

https://github.com/ovrclk/akash/blob/c6cf3e02e88809dc990ec2945301f766a8402df4/sdl/coin_test.go#L19-L31

akhilkumarpilli:
Yeah, but now it's not throwing error. Before, it used to throw error when we sent decimals as input. That's the reason, we expect it should be fail.

boz:
I see, okay.

PR #1017, opened 2020-12-28
feat: implement mTLS authentication

### LINKED ISSUE #988: Authentication between provider & deployment owner
We need to implement a form of authentication between the provider & the deployment owner. This needs to be bidirectional.

This should be possible by placing the public part of a cert. on the blockchain for each entity & using mTLS to secure the connection established to the provider.

troian:
fixes #988

- [x] proto API
- [x] query api
- [x] cli create certificate
- [x] cli revoke certificate
- [x] server to validate client certificate
- [x] client to validate server certificate
- [x] fix existing integration tests
- [x] validate dns name in certificate
- [x] allow whitelisted server ip addresses
- [x] `x/cert` tests 
- [ ] cli integration tests 
- [x] gateway obtain owner id from context
- [x] make certificate routines into shared utils
- [x] generate certificates into genesis
- [x] provider status endpoint is TLS without client authentication

### testing
`_run/lite` is set to play with `mTLS`
generated certificates are located in `home` dir
default key for deployment operations is `main`

1. `make init` - along with other init stuff also creates certificate for each key and adds it to genesis. `provider` certificate has both `Server` and `Client` ExtKeyUsage
2. __t2__ `make node-run`
3. `make query-certificates` query onchain certificates
3. `make provider-create`
4. __t3__ `make provider-run`
5. `make provider-status` obtain provider status via HTTPS with mTLS authentication on client side.
6. `make deployment-create` create deployment
7. `make send-manifest KEY_NAME=validator` errors
```
provider error messsage:
no lease for deployment
```
8. `make send-manifest` manifest sent successfully
9. `make provider-lease-status KEY_NAME=validator` no lease will be found
```
uri: https://localhost:8080/lease/5/1/1/status
{
  "services": null,
  "forwarded-ports": null
}
```
10. `make provider-lease-status` lease status obtained successfully
```
uri: https://localhost:8080/lease/5/1/1/status
{
  "services": {
    "web": {
      "name": "web",
      "available": 2,
      "total": 2,
      "uris": null,
      "observed-generation": 0,
      "replicas": 0,
      "updated-replicas": 0,
      "ready-replicas": 0,
      "available-replicas": 0
    }
  },
  "forwarded-ports": null
}
```
11. `make provider-service-status`
```
uri: https://localhost:8080/lease/5/1/1/service/web/status
{
  "name": "",
  "available": 0,
  "total": 0,
  "uris": null,
  "observed-generation": 0,
  "replicas": 0,
  "updated-replicas": 0,
  "ready-replicas": 0,
  "available-replicas": 0
}
```
12. `make provider-service-logs`
```
service web does not have running pods
```


provider/cmd/leaseStatus.go
```diff
@@ -23,37 +22,45 @@ func leaseStatusCmd() *cobra.Command {
...
 
-	addr, err := mcli.ProviderFromFlagsWithoutCtx(cmd.Flags())
+	prov, err := providerFromFlags(cmd.Flags())
 	if err != nil {
 		return err
 	}
 
-	pclient := pmodule.AppModuleBasic{}.GetQueryClient(cctx)
-	res, err := pclient.Provider(context.Background(), &ptypes.QueryProviderRequest{Owner: addr.String()})
+	dseq, gseq, oseq, err := parseLeaseFromFlags(cmd.Flags())
```
boz, line 41: Can you fold this into `x/market/client/cli/flags.go` and `x/deployment/cli/flags.go`?

I am worried about constructing a `LeaseID` below that doesn't have `Owner` populated.

boz (commented):
Looks great @troian

hydrogen18:
@troian I ran through your steps and everything worked as expected. My only remark would be when we run

```
make provider-lease-status KEY_NAME=validator

```

It should probably return 403 as the status code. Since the deployment info is valid but the key does not authorize retrieval of the data.

troian:
@hydrogen18 this behavior is a bit confusing but here is why  :)
on the client certificate is loaded from the account id specified in `--from` and passed to the server for authentication.
on the server account id for lease is extracted from the certificate so server is actually trying to find lease for `<validator account>/1/1/1` not `<main account id>/1/1/1`.

hydrogen18:
OK, in that case it seems like it should be 404 status code? Since it doesn't exist at all.

troian:
yah, you're right

_run/lite/provider.yaml
```diff
@@ -1,4 +1,4 @@
-host: http://localhost:8080
+host: https://localhost:8080
```
hydrogen18, line 1: maybe we should move this to port 8443 to be consistent?

provider/cmd/run.go
```diff
@@ -290,6 +293,11 @@ func doRunCmd(ctx context.Context, cmd *cobra.Command, _ []string) error {
 
 	gwaddr := viper.GetString(FlagGatewayListenAddress)
 
+	cert, err := cutils.LoadCertificateFromFrom(cctx.HomeDir, cctx.FromAddress, txFactory.Keybase())
```
hydrogen18, line 296: This seems like it is named weird ? It ends with "FromFrom" ?

provider/cmd/run.go
```diff
@@ -418,7 +438,8 @@ func showErrorToUser(err error) error {
 	// If the error has a complete message associated with it then show it
 	clientResponseError, ok := err.(gateway.ClientResponseError)
 	if ok && 0 != len(clientResponseError.Message) {
-		fmt.Fprintf(os.Stderr, "provider error messsage:\n%v\n", clientResponseError.Message)
+		_, _ = fmt.Fprintf(os.Stderr, "provider error messsage:\n%v\n", clientResponseError.Message)
+		err = nil
```
hydrogen18, line 442: What does suppressing the error on this line do?

provider/cmd/serviceLogs.go
```diff
@@ -26,56 +25,44 @@ func serviceLogsCmd() *cobra.Command {
...
-
-	bid, err := mcli.BidIDFromFlagsWithoutCtx(cmd.Flags())
+	outputFormat, err := cmd.Flags().GetString("format")
 	if err != nil {
 		return err
 	}
 
-	lid := mtypes.MakeLeaseID(bid)
+	if outputFormat != "text" && outputFormat != "json" {
+		return errors.Errorf("invalid output format %s", outputFormat)
```
hydrogen18, line 64: It'd be helpful to indicate what the valid output formats are in this error message.

provider/gateway/client.go
```diff
@@ -38,22 +44,66 @@ type ServiceLogMessage struct {
...
+	OnClose <-chan string
 }
 
+type ClientOptions func(Client)
+
 // NewClient returns a new Client
-func NewClient() Client {
-	return &client{
-		hclient: http.DefaultClient,
+func NewClient(cctx cclient.Context, addr sdk.Address, certs []tls.Certificate) (Client, error) {
```
hydrogen18, line 54: Is it worth checking in the code to see if someone passes `certs` with `len(certs) == 0`? No one should, but it is a common problem.

provider/gateway/client.go
```diff
@@ -38,22 +44,66 @@ type ServiceLogMessage struct {
...
+	cl := &client{
+		host:    res.Provider.HostURI,
+		addr:    addr,
+		qclient: cmodule.AppModuleBasic{}.GetQueryClient(cctx),
+	}
+
+	tlsConfig := &tls.Config{
+		ServerName:            uri.Host,
+		Certificates:          certs,
+		InsecureSkipVerify:    true,
```
hydrogen18, line 76: Shouldn't this be false? Otherwise we can get MITM attack'd

provider/gateway/client.go
```diff
@@ -38,22 +44,66 @@ type ServiceLogMessage struct {
...
+
+	cl.hclient = &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsConfig,
+		},
+	}
+
+	cl.wsclient = &websocket.Dialer{
+		Proxy:            http.ProxyFromEnvironment,
+		HandshakeTimeout: 45 * time.Second,
```
hydrogen18, line 90: This should probably be configurable, although 45 seconds is a fine default.

provider/gateway/server.go
```diff
@@ -2,19 +2,105 @@ package gateway
...
 			return ctx
 		},
 	}
+
+	srv.TLSConfig = &tls.Config{
+		Certificates:       certs,
+		ClientAuth:         tls.RequestClientCert,
+		InsecureSkipVerify: true,
+		MinVersion:         tls.VersionTLS13,
+		VerifyPeerCertificate: func(certificates [][]byte, _ [][]*x509.Certificate) error {
```
hydrogen18, line 41: This seems very similar to the another method named `verifyPeerCertificate`. Can we reduce the duplication?

provider/gateway/client.go
```diff
@@ -69,33 +119,92 @@ func (err ClientResponseError) ClientError() string {
...
+	var prov sdk.Address
+	if prov, err = sdk.AccAddressFromBech32(cert.Subject.CommonName); err != nil {
+		return errors.Wrap(err, "invalid certificate")
+	}
+
+	if !c.addr.Equals(prov) {
+		return errors.Errorf("tls: hijacked certificate")
+	}
+
+	// 2. serial number must be in
```
hydrogen18, line 147: This comment doesn't make sense.

provider/gateway/client.go
```diff
@@ -69,33 +119,92 @@ func (err ClientResponseError) ClientError() string {
...
+		return errors.Wrap(err, "invalid certificate")
+	}
+
+	return nil
 }
 
-func createClientResponseErrorIfNotOK(resp *http.Response, responseBuf *bytes.Buffer) error {
-	if resp.StatusCode == http.StatusOK {
-		return nil
+func (c *client) verifyConnection(_ tls.ConnectionState) error {
```
hydrogen18, line 188: I think you can just set `verifyConnection: nil` in the TLS configuration if this method does not need to do anything

provider/gateway/client.go
troian: `certs` can be empty if there is no need in authenticated request, for example provider status.
point to update the doc

provider/gateway/client.go
troian: No, `InsecureSkipVerify` must be set `true` as root of trust is build in `VerifyPeerCertificate` and it won't be invoked when `InsecureSkipVerify: false` as TLS library will go through normal certificate validation process and fail.
Basically we do normal certificate verification in `VerifyPeerCertificate`

provider/gateway/server.go
troian: No, they work differently. Client certificate not always supplied

provider/gateway/client.go
```diff
@@ -157,6 +266,7 @@ func (c *client) ServiceStatus(ctx context.Context, host string, id mtypes.Lease
 }
 
 func (c *client) getStatus(ctx context.Context, uri string, obj interface{}) error {
+	fmt.Printf("uri: %s\n", uri)
```
hydrogen18, line 269: I think this can be removed.

provider/gateway/client.go
```diff
@@ -282,3 +414,12 @@ func (c *client) ServiceLogs(ctx context.Context,
 
 	return logs, nil
 }
+
+func parseCloseMessage(msg string) string {
+	errmsg := strings.SplitN(msg, ": ", 3)
```
hydrogen18, line 419: It'd be helpful to have a comment showing an example of what is being parsed out here

testutil/cert.go
```diff
@@ -0,0 +1,84 @@
...
+	"encoding/pem"
+	"math/big"
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/ovrclk/akash/x/cert/types"
+)
+
+var AuthVersionOID = asn1.ObjectIdentifier{2, 23, 133, 2, 6}
```
hydrogen18, line 20: It'd be nice to have a comment explaining what these numbers mean

x/cert/genesis.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// InitGenesis initiate genesis state and return updated validator details
+func InitGenesis(ctx sdk.Context, keeper keeper.Keeper, data *types.GenesisState) []abci.ValidatorUpdate {
+	for _, record := range data.Certificates {
+		id, _ := types.ToCertID(record.ID)
+
+		err := keeper.CreateCertificate(ctx, id, record.Certificate.Cert, record.Certificate.Pubkey)
+		if err != nil {
+			ctx.Logger().Error("error init genesis cert: %s", err)
```
hydrogen18, line 32: Should we just `panic` here?

x/cert/query/querier.go
```diff
@@ -0,0 +1,113 @@
+package query
```
hydrogen18, line 1: In this file there is lots of usage of `path[2]`,`path[3:]`, `path[3]` etc. It'd be easier to understand if those were named constants in the file.

x/cert/types/cert.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/pkg/errors"
+)
+
+func (m *CertificateID) String() string {
+	return fmt.Sprintf("%s/%s", m.Owner, m.Serial)
+}
+
+func (m *CertificateID) Equals(val CertificateID) bool {
+	return (m.Owner == val.Owner) && (m.Serial == m.Serial)
```
hydrogen18, line 16: I think you have a bug here, it should be checking `m.Serial == val.Serial`

x/cert/utils/utils.go
```diff
@@ -0,0 +1,90 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/pkg/errors"
+
+	ctypes "github.com/ovrclk/akash/x/cert/types"
+)
+
+// LoadPEMFromFrom load certificate/private key from file named as
+// account name supplied in FlagFrom
+// file must contain two PEM blocks, certificate followed by a private key
+func LoadPEMFromFrom(homedir string, key sdk.Address, keyring keyring.Keyring) ([]byte, []byte, []byte, error) {
```
hydrogen18, line 23: It seems like it would be easier to return some sort of struct rather than 3 byte values here.

provider/cmd/run.go
troian: In this case command will not print usage

testutil/cert.go
troian: thats defined here https://portal.etsi.org/pnns/asn1oids

provider/cmd/run.go
troian: Changed to `LoadPEMForAccount`

hydrogen18 (approved):

_run/kube/Makefile
```diff
@@ -3,8 +3,8 @@ include ../common-commands.mk
 include ../common-kind.mk
 
 GATEWAY_HOSTNAME ?= localhost
-GATEWAY_HOST     ?= localhost:8080
-GATEWAY_ENDPOINT ?= http://$(GATEWAY_HOST)
+GATEWAY_HOST     ?= $(GATEWAY_HOSTNAME):8443
+GATEWAY_ENDPOINT ?= https://$(GATEWAY_`HOST)
```
boz, line 7: wild ```"`"```

_run/common.mk
```diff
@@ -49,6 +52,17 @@ node-init-genesis: init-dirs
 		jq -rM '(..|objects|select(has("mint_denom"))).mint_denom |= "$(CHAIN_TOKEN_DENOM)"' > \
 		"$(GENESIS_PATH)"
 
+.PHONY: node-init-genesis-certs
+node-init-genesis-certs: $(patsubst %,node-init-genesis-client-cert-%,$(CLIENT_CERTS)) $(patsubst %,node-init-genesis-server-cert-%,$(SERVER_CERTS))
+
+.PHONY: node-init-genesis-client-cert-%
+node-init-genesis-client-cert-%:
+	$(AKASHD) $(KEY_OPTS) tx cert create client --to_genesis=true --from=$(@:node-init-genesis-client-cert-%=%)
```
boz, line 60: :nail_care: `--to-genesis`

provider/cluster/kube/client.go
```diff
@@ -208,6 +212,10 @@ func (c *client) TeardownLease(ctx context.Context, lid mtypes.LeaseID) error {
 
 func (c *client) ServiceLogs(ctx context.Context, lid mtypes.LeaseID,
 	_ string, follow bool, tailLines *int64) ([]*ctypes.ServiceLog, error) {
+	if err := c.leaseExists(ctx, lid); err != nil {
```
boz, line 215: :pray:

provider/cmd/run.go
```diff
@@ -371,12 +412,22 @@ func doRunCmd(ctx context.Context, cmd *cobra.Command, _ []string) error {
...
+		ctx,
+		log,
+		service,
+		cquery,
+		gwaddr,
+		cctx.FromAddress,
+		[]tls.Certificate{cert},
+	)
+	if err != nil {
+		return group.Wait()
```
boz, line 429: `return err`?  the one above should probably be that too.

provider/gateway/rest/middleware.go
```diff
@@ -46,7 +43,41 @@ func requestService(req *http.Request) string {
...
+			// at this point client certificate has been validated
+			// so only thing left to do is get account id stored in the CommonName
+			owner, err := sdk.AccAddressFromBech32(r.TLS.PeerCertificates[0].Subject.CommonName)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusUnauthorized)
+				return
+			}
+
+			context.Set(r, ownerContextKey, owner)
+			next.ServeHTTP(w, r)
```
boz, line 75: :fire:

boz (approved):
some minor comments inline.  otherwise, :ship: :fire:

Issue #749, opened 2020-07-15
Assert EventDeployment* Version field's length


Ropes:
Originated from #565's [PR comment](https://github.com/ovrclk/akash/pull/748/files/f6970e3921590fac8140f839eec05a69715c0238#diff-fb685344e0d5164bd860526ed7e82e41) which identified we should assert then length of the Deployment's Version field.

Asserting it was equal to 64(which was the length reported by tests) returned the error which subtly broke the `EventDeploymentUpdate` message and updating functionality. It seems there's additional nuance to reporting vs the actual encoded length. Fortunately integration tests caught the problem, so it was not merged into master.

**Update the following assertions to the correct Version length**
* [ ] https://github.com/ovrclk/akash/blob/master/x/deployment/types/msgs.go#L45-L47
* [ ] https://github.com/ovrclk/akash/blob/master/x/deployment/types/msgs.go#L70-L72
* [ ] Write tests around the ValidateBasic methods for these types.

PR #1018, opened 2020-12-30
fix: assert deployment version length

akhilkumarpilli:
Fixes: #749

### DIFF
--- a/x/deployment/types/types.go
+++ b/x/deployment/types/types.go
@@ -12,13 +12,18 @@ import (
 
 type attributesMatching map[string]types.Attributes
 
-// DefaultOrderBiddingDuration is the default time limit for an Order being active.
-// After the duration, the Order is automatically closed.
-// ( 24(hr) * 3600(seconds per hour) ) / 7s-Block
-const DefaultOrderBiddingDuration = int64(12342)
+const (
+	// ManifestVersionLength is the length of manifest version
+	ManifestVersionLength = 32
 
-// MaxBiddingDuration is roughly 30 days of block height
-const MaxBiddingDuration = DefaultOrderBiddingDuration * int64(30)
+	// DefaultOrderBiddingDuration is the default time limit for an Order being active.
+	// After the duration, the Order is automatically closed.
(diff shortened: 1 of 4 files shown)


boz (changes requested):
Looks great.

Can you define a constant in the file that generates the hash and use that constant here?

https://github.com/ovrclk/akash/blob/6ee9e6f06fc022595b2e0ced400c61b46f1aa72a/sdl/sdl.go#L120-L134

akhilkumarpilli:
> Can you define a constant in the file that generates the hash and use that constant here?

Sorry, I didn't get it.

hydrogen18:
I think @boz is suggesting you add a constant like `const ManifestVersionLength = 32` in `sdl/sdl.go` and then do `if len(version) != sdl.ManifestVersionLength` in the validation code.

akhilkumarpilli:
@boz , Should I use this code to generate hash and use that constant in `ManifestVersion`?

https://github.com/ovrclk/akash/blob/master/testutil/ids.go#L41-L47

troian (approved):

x/deployment/types/types.go
```diff
@@ -12,13 +12,18 @@ import (
 
 type attributesMatching map[string]types.Attributes
 
-// DefaultOrderBiddingDuration is the default time limit for an Order being active.
-// After the duration, the Order is automatically closed.
-// ( 24(hr) * 3600(seconds per hour) ) / 7s-Block
-const DefaultOrderBiddingDuration = int64(12342)
+const (
+	// ManifestVersionLength is the length of manifest version
+	ManifestVersionLength = 32
```
boz, line 17: It'd be nice to define this in the same place that the hash is calculated - `sdl/sdl.go`.  Are there import cycle issues if you do this?

If nothing else, you can use `sha256.Size`

akhilkumarpilli: Yes, there are import cycle issues if we add in `sdl/sdl.go`. That's the reason, I moved it to types.

boz (approved):
:ship:

PR #1021, opened 2021-01-04
feat(provider): Run containers in deployments with a pod security policy

hydrogen18:
This set of changes is based off the work Josh did, but only contains the following restrictions

1. Disallow access to kubernetes persistent volumes (we have no support of this at present)
2. Disallow direct access to host network adapter 
3. Disallow host IPC communication
4. Disallow host PID usage
5. Disallow access to any devices on the host

### DIFF
--- a/_run/kube/deployment.yaml
+++ b/_run/kube/deployment.yaml
@@ -6,7 +6,6 @@ services:
     image: quay.io/ovrclk/demo-app
     expose:
       - port: 80
-        as: 80
         accept:
           - hello.localhost
         to:
--- a/_run/single/deployment.yaml
+++ b/_run/single/deployment.yaml
@@ -6,7 +6,6 @@ services:
     image: quay.io/ovrclk/demo-app
     expose:
       - port: 80
-        as: 80
         accept:
           - hello.localhost
         to:
--- a/provider/cluster/kube/apply.go
+++ b/provider/cluster/kube/apply.go
@@ -57,23 +57,22 @@ func applyNetPolicies(ctx context.Context, kc kubernetes.Interface, b *netPolBui
 	return err
 }
 
-// TODO: re-enable.  see #946
-// func applyRestrictivePodSecPoliciesToNS(ctx context.Context, kc kubernetes.Interface, p *pspRestrictedBuilder) error {
-// 	obj, err := kc.PolicyV1beta1().PodSecurityPolicies().Get(ctx, p.name(), metav1.GetOptions{})
-// 	switch {
-// 	case err == nil:
-// 		obj, err = p.update(obj)
-// 		if err == nil {
-// 			_, err = kc.PolicyV1beta1().PodSecurityPolicies().Update(ctx, obj, metav1.UpdateOptions{})
-// 		}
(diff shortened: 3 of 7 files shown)


_run/kube/deployment.yaml
```diff
@@ -5,7 +5,7 @@ services:
   web:
     image: quay.io/ovrclk/demo-app
     expose:
-      - port: 80
+      - port: 8080
```
boz, line 8: Do we still need this?

boz (commented):
Looks great!

Can you add a flag for this (default on is fine) like is done with the network policies?

_run/kube/deployment.yaml
hydrogen18: I thought I had removed this change, I'll double check.

hydrogen18:
OK, I was able to fix the issues identified. But what i am seeing is each time a deployment is created a pod security policy is added. It doesn't get deleted when the deployment is closed. So eventually it would run kubernetes out of some resource.

Apparenty pod security policies do not exist within a kubernetes namespace? So I think we just need to create one policy and make sure it applies to all deployments. I'm still doing more work to figure this out.

troian:
pod-security-policy is cluster-level resource but you can set a namespace to which it applies to

hydrogen18:
Alright, it sounds like I need to make sure it applies to the deployment namespace & make sure it is removed when the deployment is closed.

hydrogen18:
So after doing some more research on this I've been able to establish

1. Pod security policies do not exist within a namespace
2. I haven't found a way to target a pod security policy to a namespace
3. The idea behind a pod security policy is that it would be created by an administrator, then a kubernetes service account should be authorized to use it. We are just doing everything as an admin presently, which doesn't really align with this
4. There is no way to verify what pod security policy authorized a pod to be created 
5. There is no way to specify a pod security policy when creating a deployment
6. There is some discussion that pod security policies may be deprecated in the future: https://www.antitree.com/2020/11/pod-security-policies-are-being-deprecated-in-kubernetes/

boz:
> So after doing some more research on this I've been able to establish
> 
> 1. Pod security policies do not exist within a namespace
> 2. I haven't found a way to target a pod security policy to a namespace
> 3. The idea behind a pod security policy is that it would be created by an administrator, then a kubernetes service account should be authorized to use it. We are just doing everything as an admin presently, which doesn't really align with this
> 4. There is no way to verify what pod security policy authorized a pod to be created
> 5. There is no way to specify a pod security policy when creating a deployment
> 6. There is some discussion that pod security policies may be deprecated in the future: https://www.antitree.com/2020/11/pod-security-policies-are-being-deprecated-in-kubernetes/

Okay, so it doesn't make much sense for us then?  We can drop it.  Otherwise, the code looks fine; if the policy is (manually) deleted along with the namespace then I'm okay with trying it.

hydrogen18:
Putting this on the back burner for now as it doesn't seem simple to enforce policies like this. It's doable, but probably 
 quite a bit more work.

Maybe kubernetes will get its security policies upgraded from beta to full feature status in the near future & make things simpler.

PR #1024, opened 2021-01-06
feat(provider): Validate hostnames as part of a manifest before accepting it

hydrogen18:
In this pull request I did the following

1. If "DeploymentIngressStaticHosts" is set, we require all deployment groups to have at least on hostname specified
if they are going to be used for HTTP via a Kubernetes Ingress. No restrictions are added to node port stuff.
2. Checked to see if hostnames are in use before accepting a manifest. If it is in use, reject the manifest
3. Add an option to block a hostname or domains. Hostnames can be specified by using a string like "bobsdefi.com". An entire
domain can be blocked by using a string like ".bobsdefi.com". This is specified a string slice in the code, and the command line option
can be given multiple times to add multiple entries.

Implementation notes:

1. I added something as part of `provider/cluster` that I am calling the hostname service. The inventory service is used when
determining to bid on a lease, so I opted not to make this part of the inventory service. This new services just tracks
the hostnames in use, as well as checking to see if a hostname is blocked by the provider. When a hostname is "in use" it is associated with a given deployment ID. This means that later deployment updates can expand the use of hostnames without having to pick an entirely new set.
2. The deployment manager reserves & releases hostnames whenever it starts up and stops a deployment.
3. The manifest service checks hostnames availability before accepting a manifest.
4. I had to add a list of known leases & group names to the manifest service, so it can check the correct deployment groups hostnames.
5. The manifest service is responsible for determining if we should accept or reject a manifest. As a result I did some strange
stuff to avoid it blocking on a single manifest while validating the hostnames are available. It's spinning up a new goroutine
to wait on an answer from the hostname service.
6. I nuked some lingering usage of `env.Parse` for configuration, I don't think we were setting that environment variable anyways

On point #5, if we push out the responsiblity of accepting or rejecting a manifest into the manifest manager it might make
some of the code simpler. But at present we onlt start the manifest manager if the manifest is valid. I didn't want to change that at this time.

Test notes:
 I added and E2E test that updates an existing deployment with additional hostnames.

### DIFF
--- a/provider/cluster/hostname.go
+++ b/provider/cluster/hostname.go
@@ -1,0 +1,232 @@
+package cluster
+
+import (
+	"context"
+	"fmt"
+	lifecycle "github.com/boz/go-lifecycle"
+	dtypes "github.com/ovrclk/akash/x/deployment/types"
+	"github.com/pkg/errors"
+	"strings"
+	"sync"
+)
+
+type reserveRequest struct {
+	hostnames []string
+	result    chan<- error
+	doReserve bool
+	dID       dtypes.DeploymentID
+}
+
+type HostnameServiceClient interface {
+	ReserveHostnames(hostnames []string, did dtypes.DeploymentID) <-chan error
+	ReleaseHostnames(hostnames []string)
+	CanReserveHostnames(hostnames []string, did dtypes.DeploymentID) <-chan error
+}
+
+type SimpleHostnames struct {
+	Hostnames map[string]dtypes.DeploymentID
+	lock      sync.Mutex
+} /* Used in test code */
+
+func (sh *SimpleHostnames) ReserveHostnames(hostnames []string, did dtypes.DeploymentID) <-chan error {
+	sh.lock.Lock()
+	defer sh.lock.Unlock()
+	ch := make(chan error, 1)
+	for _, hostname := range hostnames {
+		_, inUse := sh.Hostnames[hostname]
+		if inUse {
+			ch <- fmt.Errorf("%w: host %q in use", errHostnameNotAllowed, hostname)
+			return ch
+		}
+		sh.Hostnames[hostname] = did
+	}
+
+	ch <- nil
+	return ch
+}
+
+func (sh *SimpleHostnames) CanReserveHostnames(hostnames []string, did dtypes.DeploymentID) <-chan error {
+	sh.lock.Lock()
+	defer sh.lock.Unlock()
+	ch := make(chan error, 1)
+
+	for _, hostname := range hostnames {
+		usedByDid, inUse := sh.Hostnames[hostname]
+		if inUse && !usedByDid.Equals(did) {
+			ch <- fmt.Errorf("%w: host %q in use", errHostnameNotAllowed, hostname)
+			return ch
+		}
+	}
+
+	ch <- nil
+	return ch
+}
+
+func (sh *SimpleHostnames) ReleaseHostnames(hostnames []string) {
+	sh.lock.Lock()
+	defer sh.lock.Unlock()
+
+	for _, hostname := range hostnames {
+		delete(sh.Hostnames, hostname)
+	}
+}
+
+type hostnameService struct {
+	inUse map[string]dtypes.DeploymentID
+
+	requests chan reserveRequest
+	releases chan []string
+	lc       lifecycle.Lifecycle
+
+	blockedHostnames []string
+	blockedDomains   []string
+}
+
+const HostnameSeparator = '.'
+
+func newHostnameService(ctx context.Context, cfg Config) *hostnameService {
+
+	blockedHostnames := make([]string, 0)
+	blockedDomains := make([]string, 0)
+	for _, name := range cfg.BlockedHostnames {
+
+		if len(name) != 0 && name[0] == HostnameSeparator {
+			blockedDomains = append(blockedDomains, name)
+			blockedHostnames = append(blockedHostnames, name[1:])
+		} else {
+			blockedHostnames = append(blockedHostnames, name)
+		}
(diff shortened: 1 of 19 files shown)


boz (commented):
Looking pretty good.  I think that we don't need to use async style for this.  It is okay if manifest processing is a bit slow, and the hostname lock service isn't listening on a queue or anything.

We probably went too far with reserving resources too ( https://github.com/ovrclk/akash/blob/d0788a813e72d1253ca8248f1ab6036d58e2c26f/provider/bidengine/order.go#L167-L170 ) but that folded in pretty cleanly compared to this.

Another thing to address:  this should handle updating manifests with new hostnames.

hydrogen18:
@boz  I think I addressed most of the points you had with a refactor. I pushed all the manifest checks into the manager, the service just starts it immediately now.

I think I still need to do some more testing around updating a deployment, I changed the logic the manager is using to check if manifest versions are valid.

provider/manifest/manager.go
```diff
@@ -333,14 +341,58 @@ func (m *manager) validateRequests() {
...
 func (m *manager) validateRequest(req manifestRequest) error {
 	// ensure that an uploaded manifest matches the hash declared on
 	// the Akash Deployment.Version
 	version, err := sdl.ManifestVersion(req.value.Manifest)
 	if err != nil {
 		return err
 	}
-	if !bytes.Equal(version, m.data.Deployment.Version) {
+
+	var versionExpected []byte
```
hydrogen18, line 387: @boz @troian  I refactored the code here that checks the manifest version that is sent vs. what comes from the blockchain.

Previously, only `m.data.Deployment.Version` was checked. This was the first value received. The value `m.versions` was never checked. After this change, it checks the last version (which should be the most recent version broadcast on the blockchain) seen.

hydrogen18:
I made some additional changes here and added a test to cover the deployment update case.

provider/manifest/service.go
```diff
@@ -201,13 +225,60 @@ loop:
...
+
+		case pendingReq := <-pendingManifests:
+			req := pendingReq.req // Get original request
+
+			// Check that the hostnames are available
+			if pendingReq.hostnamesCheckResult != nil {
+				req.ch <- pendingReq.hostnamesCheckResult
+				break
+			}
+
```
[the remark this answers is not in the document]
hydrogen18: After my last refactor I have pushed out everything into the manifest manager, so the service isn't really handling anything anymore. This probably is easier to work with since everything happens in the manifest.

The code that matches up group names has been moved into it's own function and just returns immediately if  it finds a problem.

provider/cluster/hostname.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+func (hs *hostnameService) ReleaseHostnames(hostnames []string) {
+	hs.releases <- hostnames
+}
+
+func (hs *hostnameService) CanReserveHostnames(hostnames []string, did dtypes.DeploymentID) <-chan error {
+	returnValue := make(chan error, 1) // Buffer of one so service does not block
+	request := reserveRequest{
+		hostnames: hostnames,
```
boz, line 212: Maybe `strings.ToLower` it here so the burden is on the calling threads.  Probably doesn't make much of a difference tho.

provider/manifest/manager.go
```diff
@@ -333,14 +341,58 @@ func (m *manager) validateRequests() {
...
+func (m *manager) checkHostnamesForManifest(requestManifest manifest.Manifest, groupNames []string) error {
+	allHostnames := make([]string, 0)
+
+	for _, mgroup := range requestManifest.GetGroups() {
+		for _, groupName := range groupNames {
+			// Only check leases with a matching deployment ID & group name
+			if groupName == mgroup.GetName() {
+				allHostnames = append(allHostnames, util.AllHostnamesOfManifestGroup(mgroup)...)
+				// For each service that exposes via an Ingress, then require a hsotname
+				if m.config.HTTPServicesRequireAtLeastOneHost {
```
boz, line 355: remove two levels of nesting:
```go
if groupName != mgroup.GetName() {
  continue
}
allHostNames = foo
if !m.config.HTTPServiceRequireAtLeastOneHost {
  continue
}
```

provider/manifest/config.go
```diff
@@ -1,7 +1,5 @@
 package manifest
 
-import "time"
-
-type config struct {
-	ManifestLingerDuration time.Duration `env:"AKASH_MANIFEST_LINGER_DURATION" envDefault:"5m"`
+type ServiceConfig struct {
+	HTTPServicesRequireAtLeastOneHost bool
```
boz, line 4: `IngressExposeHostRequired`?  Just seems a bit verbose.

provider/manifest/manager.go
boz: Thanks.  This is screaming for a rewrite.

provider/manifest/manager_test.go
```diff
@@ -3,79 +3,291 @@ package manifest
...
+	did := lid.DeploymentID()
+	dgroups, err := sdl2.DeploymentGroups()
+	require.NoError(t, err)
+
+	// Tell the service that a lease has been won
+	dgroup := &dtypes.Group{
+		GroupID: dtypes.GroupID{
+			Owner: lid.Owner,
+			DSeq:  lid.DSeq,
+			GSeq:  lid.GSeq,
```
boz, line 162: `lid.GroupID()`

boz (commented):
Looks amazing.

provider/manifest/config.go
hydrogen18: Yeah, its verbose. I was just trying to be clear with it.

PR #1029, opened 2021-01-13
fix(lint): Disable network when running the linters

hydrogen18:
For some reason when running commands like `make test-sublinters` it is equivalent to unplugging & plugging in the network cable on the host each time docker is run.

This disables the network access for the linters run in docker, which don't need it anyways.


troian:
does it happen on your machine particularly? i was running linters intensively these days and it didn't seems to happen. wonder if it is host specific issue. does it happen on other docker commands within makefile?

boz (approved):

troian (approved):

PR #1032, opened 2021-01-14
feat(provider): Add the deploy tool

hydrogen18:
I made the following changes from the work that @anilCSE started

1. If we get a deployment closed event, just bail
2. If we get a lease closed event, just bail
3. Don't return the error for context cancelled if all leases are reporting as OK by providers
4. Run service status after all leases are up and dump the result to standard out.
5. Don't log the same service status multiple times
6. Use a context with a deadline, tie everything to that. Poll "forever" or until the deadline is reached

### DIFF
--- a/deploy/cmd/create.go
+++ b/deploy/cmd/create.go
@@ -1,0 +1,343 @@
+package cmd
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	mtypes "github.com/ovrclk/akash/x/market/types"
+	"os"
+	"sync"
+	"time"
+
+	"github.com/avast/retry-go"
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	ctypes "github.com/ovrclk/akash/provider/cluster/types"
+	gateway "github.com/ovrclk/akash/provider/gateway/rest"
+	dcli "github.com/ovrclk/akash/x/deployment/client/cli"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	// FlagTimeout represents max amount of time for lease status checking process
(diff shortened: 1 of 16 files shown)


deploy/cmd/create.go
```diff
@@ -0,0 +1,341 @@
...
+// createCmd represents the create command
+func createCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create [sdl-file]",
+		Args:  cobra.ExactArgs(1),
+		Short: "Create a deployment on the akash network",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			timeoutDuration := viper.GetDuration(FlagTimeout)
+			endAt := time.Now().Add(timeoutDuration)
+			ctx, cancel := context.WithDeadline(context.Background(), endAt)
```
troian, line 49: i think it is better to use `cmd.Context()`

deploy/cmd/create.go
```diff
@@ -0,0 +1,341 @@
...
+		if err != nil {
+			cancel()
+			return nil, err
+		}
+
+		servicesChecked := make(map[string]bool)
+
+		return func() error {
+			err = retry.Do(func() error {
+				ls, err := gclient.LeaseStatus(context.Background(), leaseID)
```
troian, line 261: `ctx` from parameters?

deploy/cmd/deployment-data.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	mani, err := sdlSpec.Manifest()
+	if err != nil {
+		return nil, err
+	}
+	ver, err := sdl.ManifestVersion(mani)
+	if err != nil {
+		return nil, err
+	}
+	id, err := dcli.DeploymentIDFromFlags(flags, clientCtx.GetFromAddress().String())
```
troian, line 114: use `DeploymentIDFromFlagsForOwner` instead. considering `--from` is required to load client certificate `--owner` is not necessary

deploy/cmd/create.go
hydrogen18: Good point, I'll give that a try

hydrogen18:
OK, I think I've addressed most of the comments now

troian (approved):

boz (approved):
:rocket:

PR #1033, opened 2021-01-15
test(cert): cases with expired and not started certificates

troian:


provider/gateway/rest/router_test.go
```diff
@@ -85,6 +102,187 @@ func runRouterTest(t *testing.T, authClient bool, fn func(*routerTest)) {
...
+	buf, err := json.Marshal(mani)
+	require.NoError(t, err)
+
+	req, err := http.NewRequest("PUT", uri, bytes.NewBuffer(buf))
+	require.NoError(t, err)
+
+	req.Header.Set("Content-Type", contentTypeJSON)
+
+	_, err = test.gclient.hclient.Do(req)
+	require.Error(t, err)
```
hydrogen18, line 133: It seems like we should assert some specific error here, or something in the text string at least.

hydrogen18 (approved):

provider/gateway/rest/router_test.go
troian: https://github.com/ovrclk/akash/pull/1036

PR #1042, opened 2021-01-19
fix(cert): query account certificates

troian:


troian:
@akhilkumarpilli @anilCSE is this a right way to turn offset into page for iterator?

x/cert/keeper/grpc_query.go
```diff
@@ -75,6 +75,10 @@ func (q Querier) Certificates(c context.Context, req *types.QueryCertificatesReq
 				limit = uint(req.Pagination.Limit)
 			}
 
+			if limit == 0 && offset == 0 {
+				offset = 1
```
anilCSE, line 79: Default offset should be 0. May be we can set default limit but I believe there's a default limit from the SDK already

troian: if `offset` is 0 `KVStorePrefixIteratorPaginated` exits immediately

x/cert/client/cli/query.go
```diff
@@ -46,14 +48,18 @@ func cmdGetCertificates() *cobra.Command {
...
-					Serial: cmd.Flag("serial").Value.String(),
-					State:  "",
-				},
 				Pagination: pageReq,
 			}
 
+			if value := cmd.Flag("serial").Value.String(); value != "" {
+				val, valid := new(big.Int).SetString(value, 10)
+				if !valid {
+					return errors.Errorf("invalid value in serial flag. expected integer")
```
hydrogen18, line 57: Since this is the same error message each time, it makes more sense to use an assignment like `var errInvalidCertificate = errors.new("invalid value in serial flag. expected integer")`

boz (approved):
:rocket:

PR #1043, opened 2021-01-19
feat(docs): Add executable for generating Markdown documentation

hydrogen18:
This adds an executable that dumps the markdown documentation for the `akash` command into whatever destination pass on the command line.

### DIFF
--- a/make/releasing.mk
+++ b/make/releasing.mk
@@ -11,6 +11,10 @@ build:
 akash:
 	$(GO) build $(BUILD_FLAGS) ./cmd/akash
 
+.PHONY: akash_docgen
+akash_docgen:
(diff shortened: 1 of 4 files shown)


make/releasing.mk
```diff
@@ -11,6 +11,10 @@ build:
 akash:
 	$(GO) build $(BUILD_FLAGS) ./cmd/akash
 
+.PHONY: akash_docgen
```
troian, line 14: `cmddoc` ?

hydrogen18: Are you suggesting that as the name of the executable we create?

boz (approved):
:100:

troian (approved):

PR #1045, opened 2021-01-21
Ryan/fix dashes

### LINKED ISSUE #984: provider api: remove dashes from field names
They're really cumbersome to use with `jq`, unfortunately.

`cluster-public-hostname` -> `cluster_public_hostname`, etc...

rsum1:
fixes #984 

replaced dashes in JSON field names with underscores.


_run/single/kustomize/akash-provider/kustomization.yaml
```diff
@@ -22,7 +22,7 @@ configMapGenerator:
     behavior: merge
     literals:
       - ingress-domain=app.localhost
-      - cluster-public-hostname=app.localhost
+      - cluster_public_hostname=app.localhost
```
hydrogen18, line 25: I don't really understand the motivation for this change, but if we're going change this argument then all of the remaining ones as well.

@boz What are we trying to accomplish by changing dashes to underscores?

troian: i think it relates to issue with `jq`.  in it's queries `-` has to be escaped or quoted

boz: I prefer `-`, but with `jq` it makes things more cumbersome.  You have to do something special to handle it.

```sh
jq '.["foo-bar"] | ."baz-boo"'
```
vs

```sh
jq '.foo_bar | .baz_boo'
```

It's also inconsistent with the style that is in tendermint and cosmos-sdk, maybe for this reason.

hydrogen18: I guess one thing I don't understand is why the command line arguments are being manipulated with `jq` ? 

But other than that, we'd need to change all of the flags for this to make sense.

boz: OH sorry.  Yes, these don't need to be changed.  Thank you.

@rsum1 : It's only the struct fields that have an annotation like `json:"foo-bar"` that need to be changed - this makes the field name be `"foo-bar"` in the JSON output/input.

boz (approved):
:rocket: :pray:

boz (approved):
:rocket:

boz (approved):
:rocket:

PR #1050, opened 2021-01-25
fix(provider): Manage context correctly, so the provider can shutdown

hydrogen18:
I couldn't get my provider to shutdown when I had a bunch of open deployments that the provider was trying to bid on.

I ended up making the following changes to get this to work

1. When the bidengine service starts, the service queries for existing orders. But it does not query for the bids
2. When the bidengine order starts, it will query for an existing bid _if an order already existed when the provider was started_. So this query-ing happens in parallel rather than serially in a loop
3. Don't attempt to store the bid in the bidengine order handler, instead just keep a flag indicating if it has bid already. This avoids some issues where the bidengine order handler thought a bid existed with an ID of `0/0/0` which is obviously invalid
4. Don't ignore errors when querying for the existing bid _unless_ the error message says "bid not found"
5. When the bid engine order handler stops and the bid _is not won_ always cancel the bid. This previously did not happen
6. I chased down places we were just using `context.Background()` and then refactored them into a managed context within the same file that is cancelled when the goroutine gets a shutdown signal.
7. Increased the test coverage around the bid engine order handler.


provider/cluster/manager.go
```diff
@@ -94,6 +94,7 @@ func (dm *deploymentManager) teardown() error {
 
 func (dm *deploymentManager) run() {
 	defer dm.lc.ShutdownCompleted()
+	ctx, cancel := context.WithCancel(context.Background())
```
boz, line 97: The thing with this one is that it's not clear that we should actually cancel operations when things start to shut down.

For instance if it's in the middle of closing a lease and we bounce the provider, we don't really want to cancel the api call to delete the namespace.

provider/bidengine/order.go
```diff
@@ -74,35 +76,75 @@ func newOrderInternal(svc *service, oid mtypes.OrderID, bid *mtypes.Bid, pricing
 	return order, nil
 }
 
-func (o *order) run() {
+var matchBidNotFound = regexp.MustCompile("^.+bid not found.+$")
+
+func (o *order) run(checkForExistingBid bool) {
 	defer o.lc.ShutdownCompleted()
+	// TODO -get context from parent
```
boz, line 83: It's okay - this gets cancelled when we're supposed to shut down.

provider/bidengine/order.go
```diff
@@ -74,35 +76,75 @@ func newOrderInternal(svc *service, oid mtypes.OrderID, bid *mtypes.Bid, pricing
...
 
+		case queryBid := <-queryBidCh:
+			err := queryBid.Error()
+			bidFound := true
+			if err != nil {
+				if matchBidNotFound.MatchString(err.Error()) {
+					bidFound = false
+				} else {
+					o.session.Log().Error("could not get existing bid", "err", err, "errtype", fmt.Sprintf("%T", err))
+					break
```
boz, line 136: `break loop`?  can we recover from this?

provider/bidengine/order.go
```diff
@@ -74,35 +76,75 @@ func newOrderInternal(svc *service, oid mtypes.OrderID, bid *mtypes.Bid, pricing
...
+				}
+			}
+
+			if bidFound {
+				o.bidPlaced = true
+				_ = queryBid.Value().(*mtypes.QueryBidResponse) // TODO do I need me?
```
boz, line 142: no.

provider/bidengine/order.go
```diff
@@ -228,6 +267,8 @@ loop:
 			}
 
 			// Fulfillment placed.
+			o.bidPlaced = true
+			result.Value()
```
boz, line 271: I don't think it's needed

provider/bidengine/order_test.go
```diff
@@ -141,13 +142,35 @@ func makeOrderForTest(t *testing.T, bid *mtypes.Bid, pricing BidPricingStrategy)
...
+			},
+		}
+		scaffold.queryClient.On("Bid", mock.Anything, queryBidRequest).Return(response, nil)
+	}
+	/**
 	if bid != nil {
 		bid.BidID = mtypes.MakeBidID(scaffold.orderID, scaffold.testAddr)
 		bid.Price = testutil.AkashCoin(t, 1)
-	}
+	}**/
```
boz, line 170: :scissors:

provider/cluster/monitor.go
```diff
@@ -176,10 +177,10 @@ func (m *deploymentMonitor) doCheck() (bool, error) {
 	return badsvc == 0, nil
 }
 
-func (m *deploymentMonitor) runCloseLease() <-chan runner.Result {
+func (m *deploymentMonitor) runCloseLease(ctx context.Context) <-chan runner.Result {
 	return runner.Do(func() runner.Result {
 		// TODO: retry, timeout
-		err := m.session.Client().Tx().Broadcast(context.Background(), &mtypes.MsgCloseBid{
+		err := m.session.Client().Tx().Broadcast(ctx, &mtypes.MsgCloseBid{
```
boz, line 183: we might have to do a reaper task that closes leases in addition to the one we've talked about that cleans up namespaces.

boz (commented):
Looks great, some comment inline.

provider/cluster/manager.go
hydrogen18: Yeah, I agree we'd rather wait on that to finish (even if it takes a while). I'll go back over this and try and make sure I'm not messing anything up like that.

provider/bidengine/order.go
hydrogen18: I was kind of wondering that as well. There are two possible outcomes when we get an error here 

1. Try and recover - provider might send another bid created event, even if one already exists. What happens in that case?
2. Bail completely - provider misses out on bidding on this deployment?

I'm not sure which one makes the most sense.

Option 3 - retry (with a delay) forever until we get past the error?

provider/bidengine/order.go
hydrogen18: Yeah I was leaning to that, I'll kill this line.

provider/bidengine/order.go
hydrogen18: Yeah the `result.Value()` isn't doing anything anymore. I'll kill it

hydrogen18:
@boz I think I've addressed all your comments now

boz (approved):
:pray: :rocket:

PR #1051, opened 2021-01-26
fix(provider): fix _run/single to work with certificates

troian:
load certificate to kustomize


boz (approved):
:rocket:

_docs/kustomize/akash-node/kustomization.yaml
```diff
@@ -35,7 +35,6 @@ configMapGenerator:
       - fast_sync=true
       - db_backend=goleveldb
       - db_dir=data
-      - log_level=main:info,state:info,*:error
```
hydrogen18, line 38: Don't we want to keep this as a configuration option? It still works if you set `info` from what I could tell

troian: we can keep it. `info` is a default level though

hydrogen18 (approved):

PR #1056, opened 2021-01-29
feat(provider): refactor gateway api

### LINKED ISSUE #1035: Allow streaming events from a provider for a lease
After a provider picks up a lease, the owner should optionally be able to connect via websocket **before** doing a `send-manifest` command. Then when things happen for the lease like the deployment to the cluster they should be able to get events. If it fails, whoever created the deployment at least knows what happened. We don't want to send too much information as it might be private stuff about the kubernetes cluster.

troian:
example of output

fixes #1035

<details>
  <summary>events output example</summary>
```
❯ make provider-lease-events
../../akash --home "./cache" --node "tcp://akash.localhost:80" "--keyring-backend=test" provider lease-events \
        --follow \
        --dseq "1"     \
        --from "main" \
        --provider "akash1e4adtnd546z49cvuw6g3x2cp5jcq00y5e6x39f"
{
  "type": "Normal",
  "time": "0001-01-01T00:00:00Z",
  "reason": "Pulled",
  "object": {
    "kind": "Pod",
    "namespace": "q48li8e671c01aff7hkg642hk515u7r8efmuajq73pthu",
    "name": "web-5f9ff4bfc8-xrrz6"
  }
}
{
  "type": "Normal",
  "time": "0001-01-01T00:00:00Z",
  "reason": "Sync",
  "object": {
    "kind": "Ingress",
    "namespace": "q48li8e671c01aff7hkg642hk515u7r8efmuajq73pthu",
    "name": "web"
  }
}
{
  "type": "Normal",
  "time": "0001-01-01T00:00:00Z",
  "reason": "Scheduled",
  "object": {
    "kind": "Pod",
    "namespace": "q48li8e671c01aff7hkg642hk515u7r8efmuajq73pthu",
    "name": "web-5f9ff4bfc8-xrrz6"
  }
}
{
  "type": "Normal",
  "time": "0001-01-01T00:00:00Z",
  "reason": "SuccessfulCreate",
  "object": {
    "kind": "ReplicaSet",
    "namespace": "q48li8e671c01aff7hkg642hk515u7r8efmuajq73pthu",
    "name": "web-5f9ff4bfc8"
  }
}
{
  "type": "Normal",
  "time": "0001-01-01T00:00:00Z",
  "reason": "Pulling",
  "object": {
    "kind": "Pod",
    "namespace": "q48li8e671c01aff7hkg642hk515u7r8efmuajq73pthu",
    "name": "web-5f9ff4bfc8-xrrz6"
  }
}
{
  "type": "Normal",
  "time": "0001-01-01T00:00:00Z",
  "reason": "Created",
  "object": {
    "kind": "Pod",
    "namespace": "q48li8e671c01aff7hkg642hk515u7r8efmuajq73pthu",
    "name": "web-5f9ff4bfc8-xrrz6"
  }
}
{
  "type": "Normal",
  "time": "0001-01-01T00:00:00Z",
  "reason": "Started",
  "object": {
    "kind": "Pod",
    "namespace": "q48li8e671c01aff7hkg642hk515u7r8efmuajq73pthu",
    "name": "web-5f9ff4bfc8-xrrz6"
  }
}
{
  "type": "Normal",
  "time": "0001-01-01T00:00:00Z",
  "reason": "ScalingReplicaSet",
  "object": {
    "kind": "Deployment",
    "namespace": "q48li8e671c01aff7hkg642hk515u7r8efmuajq73pthu",
    "name": "web"
  }
}
{
  "type": "Normal",
  "time": "0001-01-01T00:00:00Z",
  "reason": "Sync",
  "object": {
    "kind": "Ingress",
    "namespace": "q48li8e671c01aff7hkg642hk515u7r8efmuajq73pthu",
    "name": "web"
  }
}
```
</details>

### DIFF
--- a/provider/cluster/kube/client.go
+++ b/provider/cluster/kube/client.go
@@ -2,12 +2,15 @@ package kube
 
 import (
 	"context"
-	"k8s.io/client-go/util/flowcontrol"
+	"fmt"
 	"os"
 	"path"
 
+	"k8s.io/client-go/util/flowcontrol"
+
 	kubeErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
 
 	ctypes "github.com/ovrclk/akash/provider/cluster/types"
 	"github.com/ovrclk/akash/provider/cluster/util"
@@ -15,6 +18,7 @@ import (
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	eventsv1 "k8s.io/api/events/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
@@ -211,13 +215,94 @@ func (c *client) TeardownLease(ctx context.Context, lid mtypes.LeaseID) error {
 	return c.kc.CoreV1().Namespaces().Delete(ctx, lidNS(lid), metav1.DeleteOptions{})
 }
 
-func (c *client) ServiceLogs(ctx context.Context, lid mtypes.LeaseID,
-	_ string, follow bool, tailLines *int64) ([]*ctypes.ServiceLog, error) {
+func newEventsFeedList(ctx context.Context, events []eventsv1.Event) ctypes.EventsWatcher {
+	wtch := ctypes.NewEventsFeed(ctx)
+
+	go func() {
+		defer wtch.Shutdown()
+
+	done:
+		for _, evt := range events {
+			evt := evt
+			if !wtch.SendEvent(&evt) {
+				break done
+			}
+		}
+	}()
+
+	return wtch
+}
+
+func newEventsFeedWatch(ctx context.Context, events watch.Interface) ctypes.EventsWatcher {
+	wtch := ctypes.NewEventsFeed(ctx)
+
+	go func() {
+		func() {
+			events.Stop()
+			wtch.Shutdown()
+		}()
+
+	done:
+		for {
+			select {
+			case obj := <-events.ResultChan():
+				evt := obj.Object.(*eventsv1.Event)
+				if !wtch.SendEvent(evt) {
+					break done
+				}
+			case <-wtch.Done():
+				break done
+			}
+		}
+	}()
+
+	return wtch
+}
+
+func (c *client) LeaseEvents(ctx context.Context, lid mtypes.LeaseID, services string, follow bool) (ctypes.EventsWatcher, error) {
 	if err := c.leaseExists(ctx, lid); err != nil {
 		return nil, err
 	}
 
-	pods, err := c.kc.CoreV1().Pods(lidNS(lid)).List(ctx, metav1.ListOptions{})
+	listOpts := metav1.ListOptions{}
+	if len(services) != 0 {
+		listOpts.LabelSelector = fmt.Sprintf(akashManifestServiceLabelName+" in (%s)", services)
+	}
+
+	var wtch ctypes.EventsWatcher
+	if follow {
+		watcher, err := c.kc.EventsV1().Events(lidNS(lid)).Watch(ctx, listOpts)
+		if err != nil {
+			return nil, err
+		}
+
+		wtch = newEventsFeedWatch(ctx, watcher)
+	} else {
+		list, err := c.kc.EventsV1().Events(lidNS(lid)).List(ctx, listOpts)
+		if err != nil {
+			return nil, err
+		}
+
+		wtch = newEventsFeedList(ctx, list.Items)
+	}
+
+	return wtch, nil
+}
+
+func (c *client) LeaseLogs(ctx context.Context, lid mtypes.LeaseID,
+	services string, follow bool, tailLines *int64) ([]*ctypes.ServiceLog, error) {
+	if err := c.leaseExists(ctx, lid); err != nil {
+		return nil, err
+	}
+
+	listOpts := metav1.ListOptions{}
+	if len(services) != 0 {
+		listOpts.LabelSelector = fmt.Sprintf(akashManifestServiceLabelName+" in (%s)", services)
+	}
+
+	c.log.Error("filtering pods", "labelSelector", listOpts.LabelSelector)
+
+	pods, err := c.kc.CoreV1().Pods(lidNS(lid)).List(ctx, listOpts)
 	if err != nil {
 		c.log.Error(err.Error())
 		return nil, errors.Wrap(err, ErrInternalError.Error())
@@ -259,7 +344,6 @@ func (c *client) LeaseStatus(ctx context.Context, lid mtypes.LeaseID) (*ctypes.L
 			AvailableReplicas:  deployment.Status.AvailableReplicas,
 		}
 		serviceStatus[deployment.Name] = status
-
 	}
 
 	ingress, err := c.kc.NetworkingV1().Ingresses(lidNS(lid)).List(ctx, metav1.ListOptions{})
@@ -280,7 +364,8 @@ func (c *client) LeaseStatus(ctx context.Context, lid mtypes.LeaseID) (*ctypes.L
 		if !found {
 			continue
 		}
-		hosts := []string{}
+
+		hosts := make([]string, 0, len(ing.Spec.Rules)+(len(ing.Status.LoadBalancer.Ingress)*2))
 
 		for _, rule := range ing.Spec.Rules {
 			hosts = append(hosts, rule.Host)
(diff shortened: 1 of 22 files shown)


provider/cluster/kube/client.go
```diff
@@ -210,6 +212,95 @@ func (c *client) TeardownLease(ctx context.Context, lid mtypes.LeaseID) error {
 	return c.kc.CoreV1().Namespaces().Delete(ctx, lidNS(lid), metav1.DeleteOptions{})
 }
 
+// type eventsFeed struct {
```
hydrogen18, line 215: We should be be able to remove this.

provider/cluster/types/types.go
```diff
@@ -71,3 +74,55 @@ type ServiceLog struct {
...
+}
+
+var _ EventsWatcher = (*EventsFeed)(nil)
+
+func NewEventsFeed(ctx context.Context) *EventsFeed {
+	ctx, cancel := context.WithCancel(ctx)
+	return &EventsFeed{
+		ctx:    ctx,
+		cancel: cancel,
+		quit:   make(chan struct{}),
```
hydrogen18, line 99: it looks like this is used for signaling shutdown. Does it make sense to just use the lifecycle object here?

provider/cmd/leaseEvents.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	result, err := gclient.LeaseEvents(cmd.Context(), bid.LeaseID(), follow)
+	if err != nil {
+		return showErrorToUser(err)
+	}
+
+	for res := range result.Stream {
+		buf, err := json.MarshalIndent(res, "", "  ")
+		if err != nil {
```
hydrogen18, line 73: There is a method in `cmd/common/util.go` called PrintJSON that does this printing logic

provider/gateway/rest/client.go
```diff
@@ -293,6 +308,103 @@ func (c *client) LeaseStatus(ctx context.Context, id mtypes.LeaseID) (*cltypes.L
...
+	endpoint, err := url.Parse(c.host.String() + "/" + leaseEventsPath(id))
+	if err != nil {
+		return nil, err
+	}
+
+	switch endpoint.Scheme {
+	case "wss", "https":
+		endpoint.Scheme = "wss"
+	default:
+		return nil, errors.Errorf("invalid uri scheme \"%s\"", endpoint.Scheme)
```
hydrogen18, line 321: You can use `%q` in place of the quotes and `%s` modifier.

provider/gateway/rest/client.go
```diff
@@ -293,6 +308,103 @@ func (c *client) LeaseStatus(ctx context.Context, id mtypes.LeaseID) (*cltypes.L
...
+			return nil, ClientResponseError{
+				Status:  response.StatusCode,
+				Message: buf.String(),
+			}
+		}
+
+		return nil, err
+	}
+
+	streamch := make(chan *eventsv1.Event)
```
hydrogen18, line 343: This should probably be buffered, if someone is trying to dump the output into a file it should run faster since Go doesn't have to context switch so often.

provider/gateway/rest/client.go
```diff
@@ -407,6 +518,20 @@ func (c *client) ServiceLogs(ctx context.Context,
 		OnClose: onclose,
 	}
 
+	if err = conn.SetReadDeadline(time.Now().Add(pingWait)); err != nil {
+		onclose <- parseCloseMessage(err.Error())
+		return nil, err
+	}
+
+	conn.SetPingHandler(func(string) error {
```
hydrogen18, line 526: This looks identical to another ping handler, can this be made a struct with the needed variables & a receiver function?

provider/gateway/rest/router.go
```diff
@@ -340,6 +380,95 @@ func wsLogWriter(ctx context.Context, ws *websocket.Conn, cfg wsLogsConfig) {
...
+	}
+
+	ws.SetPongHandler(func(string) error {
+		return ws.SetReadDeadline(time.Time{})
+	})
+
+	go func() {
+		var err error
+
+		defer func() {
```
hydrogen18, line 425: This is really odd, why wouldn't we just put this after the `for` loop?

provider/gateway/rest/router.go
```diff
@@ -340,6 +380,95 @@ func wsLogWriter(ctx context.Context, ws *websocket.Conn, cfg wsLogsConfig) {
...
+			websocket.FormatCloseMessage(websocketLeaseNotFound, ""))
+		if err != nil {
+			cfg.log.Error("couldn't push control message through websocket", "error", err.Error())
+		}
+		return
+	}
+
+	defer evts.Stop()
+
+	if err = ws.SetReadDeadline(time.Time{}); err != nil {
```
hydrogen18, line 414: This is unusual, shouldn't this be setting a time in the future?

troian: Set initial deadline to 0 till ping timer fires

hydrogen18 (approved):

provider/cluster/kube/client.go
```diff
@@ -279,7 +346,7 @@ func (c *client) LeaseStatus(ctx context.Context, lid mtypes.LeaseID) (*ctypes.L
 		if !found {
 			continue
 		}
-		hosts := []string{}
+		var hosts []string
```
boz, line 349: `hosts := make([]string,0,len(ing.Spec.Rules))`?

provider/cluster/types/types.go
```diff
@@ -71,3 +75,70 @@ type ServiceLog struct {
...
+}
+
+// EventsWatcher
+type EventsWatcher interface {
+	Stop()
+	ResultChan() <-chan *eventsv1.Event
+}
+
+// EventsFeed
+type EventsFeed struct {
```
boz, line 101: I'd prefer to only expose an interface and implement this with a private struct.

Also, instead of writing to `NomNom()` with the fallback every time, how about a method that does that, like `Write()` or `Send()`?

provider/cluster/types/types.go
```diff
@@ -71,3 +75,70 @@ type ServiceLog struct {
...
+		close(e.feed)
+	}
+}
+
+func (e *EventsFeed) Done() <-chan struct{} {
+	return e.ctx.Done()
+}
+
+func (e *EventsFeed) NomNom() chan<- *eventsv1.Event {
+	return e.feed
```
boz, line 139: This will blow up if you use the pattern from earlier.  You're not guaranteed to read from `Done()` before writing to `NomNom()` even if `Stop()` has been called.  Writing to a closed channel will `panic`.

I think it's probably best to use two channels here - one for reading and one for writing.

boz (commented):
Quickly read through it.  Looks great!  Couple comments; I'll take a deeper look when I can.

provider/cluster/types/types.go
```diff
@@ -71,3 +75,70 @@ type ServiceLog struct {
...
+	e.cancel()
+}
+
+func (e *EventsFeed) Shutdown() {
+	e.Stop()
+	select {
+	case <-e.quit:
+	default:
+		close(e.quit)
+		close(e.feed)
```
boz, line 130: close, but if two threads call `Shutdown()` at the same time, they will both `close(quit)` & whoever happens to be second will blow up.

sorry :grimacing:

provider/gateway/rest/client.go
```diff
@@ -295,6 +311,103 @@ func (c *client) LeaseStatus(ctx context.Context, id mtypes.LeaseID) (*cltypes.L
...
+			if _, ok := err.(*websocket.CloseError); ok {
+				onclose <- parseCloseMessage(err.Error())
+			} else {
+				onclose <- err.Error()
+			}
+		}
+	}
+
+	if err = conn.SetReadDeadline(time.Now().Add(pingWait)); err != nil {
+		processOnCloseErr(err)
```
boz, line 364: I think it's good to just return `nil, err` here.

provider/gateway/rest/client.go
```diff
@@ -409,6 +521,20 @@ func (c *client) ServiceLogs(ctx context.Context,
 		OnClose: onclose,
 	}
 
+	if err = conn.SetReadDeadline(time.Now().Add(pingWait)); err != nil {
+		onclose <- parseCloseMessage(err.Error())
```
boz, line 525: same here - can just return `nil, err`, no?

boz (approved):
Honestly I don't really understand this one.  Merge if it works; can look at it later.

provider/cluster/kube/client.go
```diff
@@ -211,6 +214,73 @@ func (c *client) TeardownLease(ctx context.Context, lid mtypes.LeaseID) error {
...
+				evt := obj.Object.(*eventsv1.Event)
+				if !wtch.SendEvent(evt) {
+					break done
+				}
+			case <-wtch.Done():
+				break done
+			}
+		}
+
+		events.Stop()
```
hydrogen18, line 252: Should we use `defer` for these?

PR #1057, opened 2021-01-29
docs(_run/single): update readme

rsum1:
updated single example readme


troian (changes requested):
change the commit message to following format
`docs(_run/single)`: update readme`

hydrogen18 (approved):

PR #1080, opened 2021-02-11
fix(provider): disable kubernetes client API rate limiting

hydrogen18:
I couldn't find any docs on this, but switching this over should result in the client always making the request immediately rather than throttling at the application level.


boz (approved):
:ship: :yolo:

PR #1084, opened 2021-02-12
fix(deployment): check certificate exists before creating deployment

troian:


boz (approved):
:+1:

hydrogen18:
I think the `deploy` tool we added needs the same check before it starts doing things. Otherwise `send-manifest` is going to fail for that as well

troian:
superseded by #1085

PR #1092, opened 2021-02-15
fix(deploy-cmd): Create a lease once bids are in

hydrogen18:
This is the smallest set of changes I could make to the deploy tool to get it running again after the escrow refactor.


deploy/cmd/create.go
```diff
@@ -178,10 +194,110 @@ func createCmd() *cobra.Command {
...
+	return lhs.ID.Provider < rhs.ID.Provider
+}
+
+func (bl bidList) Swap(i, j int) {
+	bl[i], bl[j] = bl[j], bl[i]
+}
+
+func (wfb *waitForBids) run(ctx context.Context, cancel context.CancelFunc, clientCtx client.Context, flags *pflag.FlagSet) error {
+	var bidDeadline <-chan time.Time
+	allBids := make(map[uint32]bidList)
```
boz, line 237: I think you can just use `map[mtypes.BidID]bidList` here.

boz (approved):
:rocket:

troian (approved):

PR #1099, opened 2021-02-16
docs(cert): certificates cookbook

troian:


boz (approved):
awesome.  we should get all of these over to github.com/ovrclk/docs soon.

PR #1109, opened 2021-02-22
feat(provider): daily automatic withdrawal of lease escrow

hydrogen18:
This makes the following changes

1. Run a lease escrow withdrawal once every 24 hours automatically
2. If the provider balance dips down below 2 * min bid deposit run an escrow withdrawal automatically for all leases.


provider/cluster/lease_withdraw.go
```diff
@@ -0,0 +1,96 @@
...
+func (dw *deploymentWithdrawal) run() {
+	defer dw.lc.ShutdownCompleted()
+	ctx, cancel := context.WithCancel(context.Background())
+
+	events, err := dw.bus.Subscribe()
+	if err != nil {
+		dw.log.Error("Could not subscribe to events", "err", err)
+	}
+
+	const withdrawalPeriod = 24 * time.Hour
```
boz, line 55: I really miss having https://github.com/caarlos0/env for settings so that we could change these kinds of things without creating a new release.

boz (approved):
:fire:

PR #1114, opened 2021-02-24
fix(provider): don't try and get ingress data for a service that has none 

hydrogen18:
What I changed here

1. Some better error messages in the logs
2. For service status, pull the manifest from the CRD in kubernetes
3. Only check the ingress status if CRD says the service should have an ingress object
4. Add some tests using mocks

### DIFF
--- a/provider/cluster/kube/client.go
+++ b/provider/cluster/kube/client.go
@@ -43,6 +43,7 @@ var (
 	ErrNoDeploymentForLease     = errors.New("kube: no deployments for lease")
 	ErrNoGlobalServicesForLease = errors.New("kube: no global services for lease")
 	ErrInternalError            = errors.New("kube: internal error")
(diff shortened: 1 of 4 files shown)


provider/cluster/kube/client.go
```diff
@@ -304,7 +305,7 @@ func (c *client) LeaseLogs(ctx context.Context, lid mtypes.LeaseID,
 
 	pods, err := c.kc.CoreV1().Pods(lidNS(lid)).List(ctx, listOpts)
 	if err != nil {
-		c.log.Error(err.Error())
+		c.log.Error("listing pods", "err", err.Error())
```
boz, line 308: I *think* you can use `err` instead of `err.Error()`

boz (approved):
:rocket:

PR #1115, opened 2021-02-24
feat(provider): allow specifying kubernetes config file path from any…

hydrogen18:
This adds a flag for specifying the kubernetes configuration path, but doesn't change the existing behavior.

### DIFF
--- a/provider/cmd/run.go
+++ b/provider/cmd/run.go
@@ -72,6 +72,7 @@ const (
 	FlagOvercommitPercentStorage         = "overcommit-pct-storage"
 	FlagDeploymentBlockedHostnames       = "deployment-blocked-hostnames"
 	FlagAuthPem                          = "auth-pem"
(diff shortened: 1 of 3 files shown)


provider/cmd/run.go
```diff
@@ -72,6 +72,7 @@ const (
 	FlagOvercommitPercentStorage         = "overcommit-pct-storage"
 	FlagDeploymentBlockedHostnames       = "deployment-blocked-hostnames"
 	FlagAuthPem                          = "auth-pem"
+	FlagKubeConfig                       = "kube-config"
```
troian, line 75: shall we declare flag as `kubeconfig` instead? this way it will match to env variable `AKASH_KUBECONFIG` (which is pretty close to original `KUBECONFIG`) instead of `AKASH_KUBE_CONFIG`

hydrogen18: Yeah, I'll make this change.

PR #1116, opened 2021-02-24
build: update build env to use go 1.16

troian:


hydrogen18 (approved):

hydrogen18 (approved):

.github/workflows/k8s-integration.yaml
```diff
@@ -17,9 +17,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - shell: bash
+        run: |
+          echo "loading env from .makerc"
+          while IFS='' read -r LINE || [ -n "${LINE}" ]; do
+            echo "$LINE" >> $GITHUB_ENV;
+          done < .makerc
```
boz, line 25: `cat .makerc >> "$GITHUB_ENV`?

tools.go
```diff
@@ -4,9 +4,6 @@ package tools
 
 //nolint
 import (
-	_ "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger"
-	_ "github.com/regen-network/cosmos-proto/protoc-gen-gocosmos"
-	_ "github.com/vektra/mockery"
 	_ "golang.org/x/tools/cmd/stringer"
```
boz, line 7: I think we don't use stringer at all anymore.

boz (approved):
:+1:

.github/workflows/k8s-integration.yaml
troian: that won't work sadly.
changed to use `c-py/action-dotenv-to-setenv@v3` instead

PR #1117, opened 2021-02-24
fix(provider): validate the ingress hostname

hydrogen18:
This puts basic validation on the content of the ingress hostname. Only runs on startup.


hydrogen18:
I am going to move the logic I have here for sanity checking a hostname into a utility function and make sure it doesn't use any regexes

hydrogen18:
@troian updated this PR to use the function from the Golang source repo

troian (approved):

PR #1119, opened 2021-03-02
fix: Switch to versions of Cosmos & Tendermint with backports & patches

hydrogen18:
this switches over to our temporary forks of Cosmos & Tendermint to get the following changes

1. The address prefix fix backported into Cosmos SDK
2. Prevent dropping of events if the channel is full in Tendermint
3. Enable Websocket Ping Pong behavior by default in Tendermint


boz (approved):
:pray: :100:

PR #1135, opened 2021-03-04
feat(provider): Allow specification of runtime class name, so gvisor …

hydrogen18:
This is a boring PR to fix some problems & allow specification of a runtime class name so we can use gVisor in production if we want to.


boz (approved):
:rocket:

boz:
@hydrogen18 check the length of the commit message's first line - looks like it goes over 50 chars.

troian (approved):

PR #1136, opened 2021-03-05
fix: raise up akash cosmos fork version to get new multisig fix

hydrogen18:
This has the latest version of our fix. Moving to this tag:

https://github.com/ovrclk/cosmos-sdk/tree/v0.41.4-akash-2


anilCSE (approved):
utACK

PR #1149, opened 2021-03-09
fix(cmd): silence usage print on command error

### LINKED ISSUE #1145: akash shows help for any error
If there is an error not related to flag or argument issues, it shouldn't show help.

I'm fine if we just make it so that it never shows help unless it's asked for explicitly.

troian:
fixes #1145


boz (approved):
:rocket:

PR #1155, opened 2021-03-10
fix(provider): Remove a race condition around the logger

hydrogen18:
Just remove this log statement for now, it's a debug level anyways. We shouldn't be dependent on it.

### DIFF
--- a/provider/balance_checker.go
+++ b/provider/balance_checker.go
@@ -51,7 +51,8 @@ func (bc *balanceChecker) doCheck(ctx context.Context) (bool, error) {
 	}
 
 	balance := result.Balance.Amount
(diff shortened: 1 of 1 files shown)


troian (approved):

provider/balance_checker.go
```diff
@@ -51,7 +51,8 @@ func (bc *balanceChecker) doCheck(ctx context.Context) (bool, error) {
 	}
 
 	balance := result.Balance.Amount
-	bc.log.Debug("provider acct balance", "balance", balance)
+	// TODO - find a way to use the logger without a race
```
boz, line 54: maybe

```go
log := bc.log.With()
go func() { log.Debug("foo") }()
```

boz (approved):
:+1:

PR #1159, opened 2021-03-10
fix: parse CLI flags for all commands

hydrogen18:
I don't know why we had this turned off, but I did a smoke test and it fixed the issue for me where it was trying to parse the home directory that is stored at `~/.akash`. I don't see any reason to have this disabled.


provider/balance_checker_test.go
```diff
@@ -49,7 +49,7 @@ func balanceCheckerForTest(t *testing.T, balance int64) (*scaffold, *balanceChec
```
hydrogen18, line 49: @troian I am still getting some races reported around this but only in the test mock code, so I have refactored the test now.

troian: 👍

troian (approved):

PR #1161, opened 2021-03-12
Add supply queries

akhilkumarpilli:


hydrogen18:
@akhilkumarpilli why is this a new project and not just more code in the akash project?

boz:
@hydrogen18 the idea was to make it an independent sdk module so that other projects could use it easier.

boz (approved):
:+1:

PR #1162, opened 2021-03-12
fix(provider): only bid on deployments that attributes have been audited

hydrogen18:
I think I made this way more complex than it needed to be, but it seems that everything is working as it should now. The provider only bids on deployments if has those attributes & the correct signature  is on the blockchain.

### DIFF
--- a/provider/bidengine/order.go
+++ b/provider/bidengine/order.go
@@ -3,6 +3,7 @@ package bidengine
 import (
 	"context"
 	"fmt"
+	atypes "github.com/ovrclk/akash/x/audit/types"
 	"regexp"
 
 	lifecycle "github.com/boz/go-lifecycle"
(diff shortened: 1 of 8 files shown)


provider/bidengine/order.go
```diff
@@ -316,20 +334,53 @@ loop:
...
+	if signatureRequirements.Size() != 0 {
+		// Check that the signature requirements are met for each attribute
+		var provAttr []atypes.Provider
+		ownAttrs := atypes.Provider{
+			Owner:      o.session.Provider().Owner,
+			Auditor:    "",
+			Attributes: o.session.Provider().Attributes,
+		}
+		provAttr = append(provAttr, ownAttrs)
+		auditors := make([]string, 0)
```
boz, line 355: to quiet lints maybe

```go
auditors := make([]string,0,len(x)+len(y))
```

boz (approved):
:+1:

I'd prefer to have us start to use an external cache for this stuff to keep the complexity down, but this is great given what we actually have.

As a final note, the reason this came up is apparently that bids which fail are not unreserving inventory.  This is an issue we should fix in general - there are and will be other reasons for the bid to fail (lease already created, order closed, too many existing bids, cosmic rays, etc...) ; we should have a catch-all in addition to this.

PR #1169, opened 2021-03-15
fix(deploy): check client certificate exists before creating deployment

### LINKED ISSUE #1164: `akash deploy` does not require/create a client certificate
Like `akash tx deployment create` does.

troian:
fixes #1164

### DIFF
--- a/deploy/cmd/create.go
+++ b/deploy/cmd/create.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/ovrclk/akash/cmd/common"
+	cutils "github.com/ovrclk/akash/x/cert/utils"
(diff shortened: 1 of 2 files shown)


deploy/cmd/create.go
```diff
@@ -75,6 +76,15 @@ func createCmd() *cobra.Command {
 				return err
 			}
 
+			if _, err = cutils.LoadCertificateForAccount(clientCtx, clientCtx.Keyring); err != nil {
+				if os.IsNotExist(err) {
```
hydrogen18, line 80: Shouldn't we also check that the certificate is on chain?

troian: 💯

troian: fixed

boz (approved):
:rocket:

PR #1173, opened 2021-03-15
feat(provider): Allow provider pricing to be a decimal value

hydrogen18:
This switches to using a decimal type for the provider pricing calculation when 'scale' is used.

If you use a bunch of tiny values, they get rounded up to the current minimum of 1uakt. If we come up with a fractional scheme in the near future we can just eliminate the rounding.


provider/bidengine/pricing.go
```diff
@@ -69,81 +70,70 @@ func (fp scalePricing) calculatePrice(ctx context.Context, gspec *dtypes.GroupSp
...
-	scale.SetUint64(fp.endpointScale)
-	endpointTotal.Mul(endpointTotal, scale)
+	storageTotal = storageTotal.Div(mebibytes)
+	storageTotal = storageTotal.Mul(fp.storageScale)
 
-	memoryTotalInt := ceilBigRatToBigInt(memoryTotal)
-	storageTotalInt := ceilBigRatToBigInt(storageTotal)
+	endpointTotal = endpointTotal.Mul(fp.endpointScale)
 
+	maxAllowedValue := decimal.NewFromInt(0x7fffffffffffffff)
```
boz, line 110: I think that there's a `math.MaxInt64` or something.

provider/bidengine/pricing_test.go
```diff
@@ -79,7 +80,7 @@ func defaultGroupSpec() *dtypes.GroupSpec {
 }
 
 func Test_ScalePricingFailsOnOverflow(t *testing.T) {
-	pricing, err := MakeScalePricing(math.MaxUint64, 0, 0, 0)
+	pricing, err := MakeScalePricing(decimal.New(math.MaxInt64, 2), decimal.Zero, decimal.Zero, decimal.Zero)
```
boz, line 83: yeah you have it here

boz (approved):
:+1:

provider/bidengine/pricing.go
hydrogen18: Fixed.

troian (approved):

PR #1175, opened 2021-03-17
fix: use a script and fetch the kubernetes code generator

hydrogen18:
I changed the kubernetes code generator to use a temporary `GOPATH` location & `go.mod`.

I bumped the version downwards on the protoc swagger stuff, to the last  version I could find that has that executable we want in it.

I still have to manually run `sudo npm install -g swagger-combine` to get the `swagger-combine` on my path. There are some npm commands in the scripts, but they appear to be ineffective. The build process still relies on `swagger-combine` being available globally.


boz (commented):
oof I really don't want to do anything with `npm` as root.  but... okay.

I still don't understand why this wasn't sufficient:

https://github.com/ovrclk/akash/blob/aa2ad8a31a37bc9a63f109a9fd4cc7763946d388/make/setup-cache.mk#L92-L99

but if it's not let's get rid of it.

PR #1179, opened 2021-03-22
feat(provider): create gspec validate endpoint

### LINKED ISSUE #1167: ability to query pricing before creating order
The client should be able to ask providers if the max price is acceptable to them.  Extending this, a bit further, we can have a `/validate` endpoint that validates not only price but everything else.

The flow would then be:

1. client queries a sample of providers that match its requirements from the blockchain.
2. client queries each provider in the sample to see if the provider will accept their group spec.
3. if **2.** passes, create the deployment.  If **2.** fails, exit with error message from providers.

troian:
fixes #1167


provider/gateway/rest/client.go
```diff
@@ -263,6 +265,55 @@ func (c *client) Status(ctx context.Context) (*provider.Status, error) {
...
+	uri, err := makeURI(c.host, validatePath())
+	if err != nil {
+		return provider.Validate{}, err
+	}
+
+	if err = gspec.ValidateBasic(); err != nil {
+		return provider.Validate{}, err
+	}
+
+	var obj provider.Validate
```
hydrogen18, line 278: Can this be declared down below, where it is used?

troian: 💯

provider/types.go
```diff
@@ -13,3 +13,7 @@ type Status struct {
 	Manifest              *manifest.Status  `json:"manifest"`
 	ClusterPublicHostname string            `json:"cluster_public_hostname,omitempty"`
 }
+
+type Validate struct {
```
hydrogen18, line 17: Maybe `ValidateResult` is a better name?

troian: `ValidateGroupSpecResult` ?

boz (approved):
:100:

PR #1193, opened 2021-03-29
feat(sdkutil): retry in broadcast-mode block

boz:
Rewrite of https://github.com/cosmos/cosmos-sdk/blob/ca98fda6eae597b1e7d468f96d030b6d905748d7/client/tx/tx.go#L29
with a different method for broadcast-mode=block:

1. use BroadcastSync
2. query for the tx until timeout or the tx is found in a block.

### DIFF
--- a/sdkutil/broadcast.go
+++ b/sdkutil/broadcast.go
@@ -1,0 +1,177 @@
+package sdkutil
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/client/input"
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authclient "github.com/cosmos/cosmos-sdk/x/auth/client"
+	"github.com/spf13/pflag"
+	ttypes "github.com/tendermint/tendermint/types"
+)
+
+const (
+	broadcastBlockRetryTimeout = 30 * time.Second
+	broadcastBlockRetryPeriod  = time.Second
+
+	// sadface.
+
+	// Only way to detect the timeout error.
+	// https://github.com/tendermint/tendermint/blob/46e06c97320bc61c4d98d3018f59d47ec69863c9/rpc/core/mempool.go#L124
+	timeoutErrorMessage = "timed out waiting for tx to be included in a block"
+
+	// Only way to check for tx not found error.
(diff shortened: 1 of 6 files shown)


sdkutil/broadcast.go
```diff
@@ -0,0 +1,155 @@
...
+	lctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	for lctx.Err() == nil {
+
+		// wait up to one second
+		select {
+		case <-lctx.Done():
+			return syncres, err
+		case <-time.After(broadcastBlockRetryPeriod):
```
hydrogen18, line 104: Should we use exponential backoff here? We could cap the frequency at something like 3 seconds.

sdkutil/broadcast.go
```diff
@@ -0,0 +1,155 @@
...
+		// check transaction
+		res, err := authclient.QueryTx(ctx, syncres.TxHash)
+		if err == nil {
+			return res, nil
+		}
+
+		fmt.Fprintf(os.Stderr, "error querying for transaction: %v.  retrying...", err)
+	}
+
+	return syncres, lctx.Err()
```
hydrogen18, line 116: Isn't it kind of weird to return the initial response here **along with an error** ?

hydrogen18:
Only concern I have with this is we may have changed the meaning of the name "block" from the normal cosmos SDK name? Perhaps we should call it "blockPolling" or similar ?

sdkutil/broadcast.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return err
+	}
+
+	txBytes, err := ctx.TxConfig.TxEncoder()(txb.GetTx())
+	if err != nil {
+		return err
+	}
+
+	res, err := doBroadcast(ctx, broadcastBlockRetryTimeout, txBytes)
```
hydrogen18, line 70: Is it possible to pickup `broadcastBlockRetryTimeout` from a flag and have it set on the context? That way if it needs to be increased end users have some control over it?

sdkutil/broadcast.go
boz: I didn't think that it warranted it - it should happen within 20 seconds and the user is there waiting for a response.

(note: this is a quick hack; people need more feedback for sure)

I'm open to it if you feel strongly about it.

sdkutil/broadcast.go
boz: Wanted to allow access to the transaction hash.  I can take it out.

sdkutil/broadcast.go
boz: I just changed how `block` is implemented because the previous implementation wasn't working.  I'd rather not add a fourth option and confuse people.

troian (approved):

boz:
@hydrogen18 @troian I changed this so that it uses the original method (`block`) first.

Issue #1174, opened 2021-03-17
Fractional uakt bid pricing


hydrogen18:
The AKT token price has gotten high enough that a 1 uakt per block bid can be pretty pricey on a daily basis. For small deployments this is a problem.

Bid prices need to be capable of being expressed in fractional uakt, use the decimal coin type.

We'll need to do the following

1. Convert the protobuf for bid pricing to use `DecCoin` instead of `Coin`
2. Convert the protobuf for payments (part of escrow) to use `DecCoin` instead of `Coin`
3. Allow the manifest to specify fractional uakt
4. Only settle an escrow payment when it is >1 uakt. Since this still needs to be settled in whole uakt, payments are going to be tracked to "fractional block height" where this fractional height is less than the current block height during settlement. The combination of fractional uakt and fractional block height settlement (tracked on a per payment basis) gives us whole uakt which can be deducted from the escrow account.
5. Do a cosmos upgrade to convert the blockchain over, the basic explanation of this is here https://pkg.go.dev/github.com/cosmos/cosmos-sdk/x/upgrade
6. When a deployment is closed or a lease terminates, check to see if the escrow account has ever made a payment. If it hasn't, make sure the end-of-lease settlement pays at least 1uakt. This prevents deployment abuse that could give "free" delployments.

hydrogen18:
@boz Thoughts on what I've outlined here?

boz:
Looks good.  Some thoughts on escrow:

Settlement happens for basically any operation that modifies an escrow account or its payments.  It "moves" tokens from the account to the payment object such that everything is calculated up to the block during which the settlement is running.

If a new payment is added, from a new lease for instance, we _have_ to settle.  Unless everything is done in a different way of course.

I think that a reasonable way to do it would be to keep things in `DecCoin` until withdraw.  For instance, have `Payment.Balance` be a `DecCoin` but have `Payment.Withdrawn` be a `Coin` https://github.com/ovrclk/akash/blob/aa2ad8a31a37bc9a63f109a9fd4cc7763946d388/x/escrow/keeper/keeper.go#L189-L190

This way partial `uakt` can accumulate in `Payment.Balance` over multiple account settlements.

When a payment is closed, any un-transferred `Payment.Balance` (note: it should be less than `1uakt`) can be "transferred" back to the `Account.Balance` (removed from `Account.Transferred`).

Thoughts?

hydrogen18:
Yeah, that might be an easier strategy to implement

[...]

PR #1201, opened 2021-04-08
feat: Allow fractional uakt for settlement

hydrogen18:
Stuff I changed

1. For escrow accounts, use a decimal coin
2. For bid pricing and deployment max price, use a decimal coin
3. Rework some of the settlement code, since with decimal types it isn't really as complex
4. Migrate all of our types to v1beta2 
5. Keep the v1beta1 code for now, in a package called `typesv1beta` where appropriate
6. Make some of the logic around provider attribute types code use an interface instead, so they can run on any version
6. Add a migration that reads the binary values, parses them, and writes them back as v1beta2 version
7. In our examples, lower the proposal time down to 15 seconds so we can actually do proposals when testing

### DIFF
--- a/.github/workflows/proto.yaml
+++ b/.github/workflows/proto.yaml
@@ -10,6 +10,12 @@ jobs:
     timeout-minutes: 5
     steps:
       - uses: actions/checkout@v2
+      - uses: c-py/action-dotenv-to-setenv@v3
+        with:
+          env-file: .makerc
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "${{ env.GOLANG_VERSION }}"
       - name: go mod vendor
         run: make modvendor
       - name: lint


boz:
bump

boz:
refs #1174

proto/akash/audit/v1beta2/query.proto
```diff
@@ -0,0 +1,79 @@
+syntax = "proto3";
```
boz, line 1: We can use a mix of versions - not everything necessarily needs to move to `v1beta2`.

provider/bidengine/pricing.go
```diff
@@ -108,34 +108,40 @@ func (fp scalePricing) CalculatePrice(_ context.Context, gspec *dtypes.GroupSpec
 
 	endpointTotal = endpointTotal.Mul(fp.endpointScale)
 
-	maxAllowedValue := decimal.NewFromInt(math.MaxInt64)
+	decimal.NewFromInt(math.MaxInt64)
```
boz, line 111: :scissors:

x/escrow/keeper/keeper.go
```diff
@@ -392,11 +418,13 @@ func (k *keeper) doAccountSettle(ctx sdk.Context, id types.AccountID) (types.Acc
 	account, payments, amountRemaining = accountSettleDistributeWeighted(
 		account, payments, blockRate, amountRemaining)
 
+	/**
```
boz, line 421: :scissors:

x/escrow/keeper/keeper.go
```diff
@@ -530,18 +562,19 @@ func accountSettleFullblocks(
...
 
 	account.Transferred = account.Transferred.Add(transferred)
 	account.Balance = account.Balance.Sub(transferred)
 
 	remaining := account.Balance
 	overdrawn := true
 	if numFullBlocks.Equal(heightDelta) {
-		remaining.Amount = sdk.ZeroInt()
+		remaining.Amount = sdk.ZeroDec()
 		overdrawn = false
```
boz, line 578: does this still work if we're using `TruncateInt()` when calculating `numFullBlocks`?

x/escrow/keeper/keeper.go
```diff
@@ -550,31 +583,32 @@ func accountSettleFullblocks(
...
-	actualTransferred := sdk.ZeroInt()
+	actualTransferred := sdk.ZeroDec()
 
 	// distribute remaining balance weighted by rate
+
 	for idx := range payments {
 		payment := payments[idx]
 		// amount := (rate / blockrate) * remaining
 		amount := amountRemaining.Amount.
 			Mul(payment.Rate.Amount).
```
boz, line 603: uh... does this mutate `amountRemaining.Amount`?

boz (commented):
Looking good.  Sorry for the delay.  Is there any chance that we can only upgrade the package versions for proto defs that need to be updated for this?

Note: we'll need to revive the `mainnet/main` branch and release from there before merging this.

hydrogen18:
> Looking good. Sorry for the delay. Is there any chance that we can only upgrade the package versions for proto defs that need to be updated for this?
> 
> Note: we'll need to revive the `mainnet/main` branch and release from there before merging this.

If we stagger versions like that, how would we know what versions of what types to use?

boz:
> If we stagger versions like that, how would we know what versions of what types to use?

```go
import xv1beta1 "akash/proto/v1beta1"
import yv1beta2 "akash/proto/v1beta2"
```

x/escrow/keeper/keeper.go
hydrogen18: I don't actually know, I'll have to look at the implementation.

hydrogen18:
Ok I will work on figuring out how to stagger the versions then

boz:
superseded by #1394

PR #1203, opened 2021-04-09
feat(provider): cancel bids if no lease is created

hydrogen18:
After a bid is created, just start a timer. If the timer is reached, cancel the bid and unreserve capacity.

This prevents a bunch of bids from taking up all capacity without any actual use of it.

### DIFF
--- a/provider/cmd/run.go
+++ b/provider/cmd/run.go
@@ -75,6 +75,7 @@ const (
 	FlagAuthPem                          = "auth-pem"
 	FlagKubeConfig                       = "kubeconfig"
 	FlagDeploymentRuntimeClass           = "deployment-runtime-class"
+	FlagBidTimeout                       = "bid-timeout"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


provider/cmd/run.go
```diff
@@ -239,6 +240,11 @@ func RunCmd() *cobra.Command {
 	if err := viper.BindPFlag(FlagDeploymentRuntimeClass, cmd.Flags().Lookup(FlagDeploymentRuntimeClass)); err != nil {
 		return nil
 	}
+
+	cmd.Flags().Duration(FlagBidTimeout, 20*time.Minute, "time after which bids are cancelled if no lease is created")
```
boz, line 244: how about 5 minutes?

boz (approved):
:fire:

provider/cmd/run.go
hydrogen18: Yeah I can change the default, I just picked something

troian (approved):

PR #1210, opened 2021-04-12
Cancel bid if a manifest is not received

hydrogen18:
This starts a watchdog per deployment ID that either is:

1. Cancelled when a manifest arrives 
2. Closes the bid when a timeout is reached

This should save AKT of folks who for someone can't send a valid manifest to the provider

### DIFF
--- a/provider/manifest/watchdog.go
+++ b/provider/manifest/watchdog.go
@@ -1,0 +1,71 @@
+package manifest
+
+import (
+	"context"
+	"github.com/boz/go-lifecycle"
+	"github.com/ovrclk/akash/provider/session"
+	dtypes "github.com/ovrclk/akash/x/deployment/types"
+	"github.com/ovrclk/akash/x/market/types"
+	"github.com/tendermint/tendermint/libs/log"
+	"time"
+)
+
+type watchdog struct {
+	leaseID types.LeaseID
+	timeout time.Duration
+	lc      lifecycle.Lifecycle
+	sess    session.Session
+	ctx     context.Context
+	log     log.Logger
+}
+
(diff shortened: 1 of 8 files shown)


provider/manifest/watchdog.go
```diff
@@ -0,0 +1,72 @@
...
+		sess:    sess,
+		ctx:     ctx,
+		log:     sess.Log().With("leaseID", leaseID),
+	}
+
+	go func() {
+		result.lc.WatchChannel(parent)
+		cancel()
+		result.lc.ShutdownAsync(nil)
+	}()
```
boz, line 37: this can just be 

```go
go result.lc.WatchChannel(parent)
```

Which will trigger a shutdown.

The newer context can be created in `run()`, but it looks like it might not really be necessary to create a new one.

boz (commented):
Love it, looks great.  One small comment re using `go-lifecycle`.

provider/manifest/watchdog.go
hydrogen18: If I create a new context in `run()`, how can I cancel it when the parent channel closes?

boz: Do you need to create a new context?

In `run()`, you can do it in a defer and in the select I suppose.

Or do it like this, but the `ShutdownAsync()` is unnecessary - `WatchChannel` will do that.

hydrogen18: I'm passing it while doing the broadcast, presumably that should be cancelled when the parent closes. Otherwise it could block for a really long time.

boz: okay

boz (approved):
:rocket:

PR #1216, opened 2021-04-15
fix(sdl): use service name to find expose endpoints

### LINKED ISSUE #1215: sdl: compute profile name assumed to be service name
`GetDeploymentGroups()` [uses the profile name](https://github.com/ovrclk/akash/blob/9b11cf65d6ec58ef31a66b639fc20f9d825b795d/sdl/v2.go#L123) when determining endpoints to include in a group instead of the service name.

This causes sdl files like the one below to not have any endpoints in their groups or orders.  Wen sending the manifest, validation fails because of an endpoint count mismatch ([`Manifest()`](https://github.com/ovrclk/akash/blob/9b11cf65d6ec58ef31a66b639fc20f9d825b795d/sdl/v2.go#L170) apparently does the right thing).

```yaml
---
version: "2.0"
services:
  web:
    image: ourwebapp
    expose:
      - port: 8080
        as: 80
        to:
          - global: true
profiles:
  compute:
    frontend:
      resources:
        cpu:
          units: 1
        memory:
          size: 2Gi
        storage:
          size: 32Gi
  placement:
    westcoast:
      pricing:
        frontend:
          denom: uakt
          amount: 100

deployment:
  web:
    westcoast:
      profile: frontend
      count: 1
```

troian:
fixes #1215

### DIFF
--- a/sdl/_testdata/deployment-svc-mismatch.yaml
+++ b/sdl/_testdata/deployment-svc-mismatch.yaml
@@ -1,0 +1,45 @@
+---
+version: "2.0"
+services:
+  web:
+    image: nginx
+    expose:
+      - port: 80
+        accept:
+          - ahostname.com
+        to:
+          - global: true
+      - port: 12345
+        to:
+          - global: true
+        proto: udp
+profiles:
+  compute:
+    web:
+      resources:
+        cpu:
+          units: "100m"
+        memory:
+          size: "128Mi"
+        storage:
+          size: "1Gi"
+  placement:
+    westcoast:
+      attributes:
+        region: us-west
+      signedBy:
+        anyOf:
+          - 1
+          - 2
+        allOf:
+          - 3
+          - 4
+      pricing:
+        web:
+          denom: uakt
+          amount: 50
+deployment:
+  webapp:
+    westcoast:
+      profile: web
+      count: 2
--- a/sdl/v2_test.go
+++ b/sdl/v2_test.go
@@ -1,8 +1,9 @@
 package sdl
 
 import (
-	"github.com/ovrclk/akash/validation"
 	"testing"
+
+	"github.com/ovrclk/akash/validation"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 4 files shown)


boz (commented):
mind adding a test for this?

troian:
test added

sdl/_testdata/deployment-svc-mismatch.yaml
```diff
@@ -0,0 +1,45 @@
+---
+version: "2.0"
```
hydrogen18, line 2: What is the error in this file that leads to the test case? I don't see it immediately.

sdl/v2_test.go
```diff
@@ -192,3 +193,10 @@ func Test_v1_Parse_ProfileNameNotServiceName(t *testing.T) {
 	require.NoError(t, err)
 	assert.Len(t, mani.GetGroups(), 1)
 }
+
+func Test_v2_Parse_DeploymentNameServiceNameMismatch(t *testing.T) {
+	sdl, err := ReadFile("./_testdata/deployment-svc-mismatch.yaml")
+	require.Nil(t, sdl)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "no service profile named")
```
boz, line 201: The problem was that it was not registering endpoints, everything else worked fine.  It was looking for endpoints by compute profile name, not service name.

```yaml
services:
  web:
    # ...
profiles:
  compute:
    bew:
      # ...
```

boz (changes requested):

hydrogen18 (approved):

sdl/v2_test.go
troian: i probably missed point here. fix makes endpoints iteration via service name. if deployment points to non-existing service it fails on `ReadFile`

valid
```yaml
services:
  web:
    # ...
profiles:
  compute:
    bew:
      # ...
  placement:
    westcoast:
      # ...
deployment:
  web:
    westcoast:
      profile: bew
```

invalid
```yaml
services:
  web:
    # ...
profiles:
  compute:
    bew:
      # ...
  placement:
    westcoast:
      # ...
deployment:
  bew:
    westcoast:
      profile: bew
```

boz: It wasn't failing it just wasn't picking up the endpoints.

boz: and yeah, your first example would have triggered it.

troian: updated test

boz (approved):
:ship:

PR #1217, opened 2021-04-15
fix(provider): query certificate status prior using gateway

### LINKED ISSUE #1208: cli: check client cert validity before doing actions that require it
People are able to run `send-manifest` if they have a "revoked" certificate.  See [here](https://discordapp.com/channels/747885925232672829/771909909335506955/830917478027886652).

`send-manifest`, `tx deployment create`, and other commands should check not only that a certificate exists, but that it is useable.

troian:
fixes #1208

### DIFF
--- a/deploy/cmd/create.go
+++ b/deploy/cmd/create.go
@@ -4,15 +4,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/ovrclk/akash/cmd/common"
-	cutils "github.com/ovrclk/akash/x/cert/utils"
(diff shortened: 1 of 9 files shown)


boz (commented):
Do the other places that use `LoadCertificateForAccount` need to be changed too?

troian:
fixed all

hydrogen18 (approved):

boz (commented):
Can we get rid of the old function then?

boz (approved):
:rocket:

PR #1225, opened 2021-04-21
fix(provider): query provider if flag is not supplied for send-manifest

### LINKED ISSUE #1177: send-manifest requires --provider
You should be able to only pass `--dseq`.

troian:
fixes #1177

output examples
**text**
```text
provider: akash1f6gmtjpx4r8qda9nxjwq26fp5mcjyqmaq5m6j7
    status: PASS
provider: akash10cl5rm0cqnpj45knzakpa4cnvn5amzwp4lhcal
    status: PASS
```

**yaml**
```yaml
- provider: akash1f6gmtjpx4r8qda9nxjwq26fp5mcjyqmaq5m6j7
  status: PASS
- provider: akash10cl5rm0cqnpj45knzakpa4cnvn5amzwp4lhcal
  status: PASS
```

**json**
```json
[{
	"provider": "akash1f6gmtjpx4r8qda9nxjwq26fp5mcjyqmaq5m6j7",
	"status": "PASS"
}, {
	"provider": "akash10cl5rm0cqnpj45knzakpa4cnvn5amzwp4lhcal",
	"status": "PASS"
}]
```

### DIFF
--- a/provider/cmd/helpers.go
+++ b/provider/cmd/helpers.go
@@ -1,24 +1,41 @@
 package cmd
 
 import (
+	"context"
+
+	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/flags"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	akashclient "github.com/ovrclk/akash/client"
+	dtypes "github.com/ovrclk/akash/x/deployment/types"
+	mtypes "github.com/ovrclk/akash/x/market/types"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
 	"github.com/ovrclk/akash/app"
 )
 
+const (
+	FlagService  = "service"
+	FlagProvider = "provider"
+	FlagDSeq     = "dseq"
+	FlagGSeq     = "gseq"
+	FlagOSeq     = "oseq"
+)
+
+const (
+	outputText = "text"
+	outputYAML = "yaml"
+	outputJSON = "json"
+)
+
 func addCmdFlags(cmd *cobra.Command) {
-	cmd.Flags().String("provider", "", "provider")
-	cmd.Flags().Uint64("dseq", 0, "deployment sequence")
+	cmd.Flags().String(FlagProvider, "", "provider")
+	cmd.Flags().Uint64(FlagDSeq, 0, "deployment sequence")
 	cmd.Flags().String(flags.FlagHome, app.DefaultHome, "the application home directory")
 	cmd.Flags().String(flags.FlagFrom, "", "name or address of private key with which to sign")
 	cmd.Flags().String(flags.FlagKeyringBackend, flags.DefaultKeyringBackend, "select keyring's backend (os|file|kwallet|pass|test)")
-
-	if err := cmd.MarkFlagRequired(FlagProvider); err != nil {
-		panic(err.Error())
-	}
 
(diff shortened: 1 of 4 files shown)


provider/cmd/helpers.go
```diff
@@ -58,3 +71,49 @@ func providerFromFlags(flags *pflag.FlagSet) (sdk.Address, error) {
 
 	return addr, nil
 }
+
+func providersFromFlagsOrQuery(flags *pflag.FlagSet, ctx context.Context, cctx client.Context, did dtypes.DeploymentID) ([]sdk.Address, error) {
+	var providers []string
+
+	if !flags.Changed(FlagProvider) {
+		cclient := akashclient.NewQueryClientFromCtx(cctx)
+		resp, err := cclient.Leases(ctx, &mtypes.QueryLeasesRequest{
```
boz, line 80: It should filter leases by whatever you pass.  An owner and dseq is required, but oseq, gseq, provider can be used to specify it further.

It shouldn't accept a random provider like is done below.

boz (changes requested):

provider/cmd/helpers.go
```diff
@@ -58,3 +71,49 @@ func providerFromFlags(flags *pflag.FlagSet) (sdk.Address, error) {
...
+				DSeq:  did.DSeq,
+				State: mtypes.Lease_State_name[int32(mtypes.LeaseActive)],
+			},
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		for _, lease := range resp.Leases {
+			providers = append(providers, lease.Lease.LeaseID.Provider)
```
hydrogen18, line 92: Could we wind up with the same provider address multiple times here?

provider/cmd/manifest.go
```diff
@@ -57,15 +66,75 @@ func doSendManifest(cmd *cobra.Command, sdlpath string) error {
...
+		return err
+	}
+
+	type result struct {
+		Provider sdk.Address `json:"provider" yaml:"provider"`
+		Status   string      `json:"status" yaml:"status"`
+		Error    error       `json:"error,omitempty" yaml:"error,omitempty"`
+	}
+
+	results := make([]result, 0, len(providers))
```
hydrogen18, line 84: can this just do `make([]result, len(providers))` and use `results[i]` in the loop rather than `append` ?

provider/cmd/manifest.go
```diff
@@ -57,15 +66,75 @@ func doSendManifest(cmd *cobra.Command, sdlpath string) error {
...
-	gclient, err := gwrest.NewClient(akashclient.NewQueryClientFromCtx(cctx), prov, []tls.Certificate{cert})
+	_, err = fmt.Fprint(cmd.OutOrStdout(), buf.String())
+
 	if err != nil {
 		return err
 	}
 
-	return showErrorToUser(gclient.SubmitManifest(context.Background(), dseq, mani))
+	if submitFailed {
+		return errors.Errorf("submit manifest to some providers has been failed")
```
hydrogen18, line 136: This can be assigned statically I think like `var errSubmitManifestFailed = errors.New(...)`

provider/cmd/manifest.go
boz: I tend to go with the three-argument version to make the loops independent of how it's allocated or if you're actually appending, etc...

provider/cmd/helpers.go
troian: there shouldn't be multiple leases for one dseq on provider

hydrogen18 (approved):

boz (approved):
:rocket:

PR #1232, opened 2021-04-28
Add prometheus metrics to provider

hydrogen18:
This adds a bunch of metrics around critical stuff. To enable it on the provider you just pass a command like arg like `--metrics-listener 0.0.0.0:3331` and it starts. If not specified, it doesn't start the metrics listener

### DIFF
--- a/provider/bidengine/order.go
+++ b/provider/bidengine/order.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	atypes "github.com/ovrclk/akash/x/audit/types"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"regexp"
 	"time"
 
@@ -14,6 +16,7 @@ import (
 	"github.com/ovrclk/akash/provider/event"
 	"github.com/ovrclk/akash/provider/session"
 	"github.com/ovrclk/akash/pubsub"
+	metricsutils "github.com/ovrclk/akash/util/metrics"
 	"github.com/ovrclk/akash/util/runner"
 	dtypes "github.com/ovrclk/akash/x/deployment/types"
 	mtypes "github.com/ovrclk/akash/x/market/types"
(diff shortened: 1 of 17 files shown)


jcudit (approved):
The project may benefit by expanding on `kubeCallsCounter` someday to incorporate durations of calls as well as their absolute counts.  End-to-end latency may be an important factor to end users and knowing kube operation runtime would help to spot slowdowns over time.

Great addition overall! 📈

hydrogen18:
Mostly done, need to add in gauges for # of running deployment monitors, etc.

_run/kube/Makefile
```diff
@@ -17,7 +17,7 @@ provider-run:
 		--cluster-node-port-quantity 100 \
 		--cluster-public-hostname "$(GATEWAY_HOSTNAME)" \
 		--bid-price-strategy "randomRange" \
-		--deployment-runtime-class "none"
+		--deployment-runtime-class "none"
```
boz, line 20: :scissors:

provider/cluster/kube/apply.go
```diff
@@ -138,22 +223,44 @@ func prepareEnvironment(ctx context.Context, kc kubernetes.Interface, ns string)
 			},
 		}
 		_, err = kc.CoreV1().Namespaces().Create(ctx, obj, metav1.CreateOptions{})
+		label := "success"
+		if err != nil  {
+			label = "fail"
+		}
+		kubeCallsCounter.WithLabelValues("namespaces-create", label).Inc()
```
boz, line 230: It'd be nice to put this whole pattern into a function.

```go
func incrementCounterWithLabel(counter Interface,name string,err error) {
   label := "success"
   if err != nil {
    label = "fail"
  }
  counter.WithLabelValues(name,label).Inc()
}
```

boz (approved):
Looks great, small suggestion inline.

provider/cluster/kube/apply.go
hydrogen18: Yeah, that's a good point. Also it solves some other code organization concerns I had. I'll make a pass to do this.

provider/bidengine/order.go
```diff
@@ -212,31 +269,37 @@ loop:
...
 				v := runner.NewResult(o.cluster.Reserve(o.orderID, group))
 				return v
-			})
+			}, reservationDuration))
 
 		case result := <-clusterch:
 			clusterch = nil
 
 			if result.Error() != nil {
+				reservationCounter.WithLabelValues("open", "fail")
```
troian, line 296: "open", "fail" etc are repeated often. const?

troian (changes requested):

provider/bidengine/order.go
```diff
@@ -82,6 +126,16 @@ func newOrderInternal(svc *service, oid mtypes.OrderID, cfg Config, pass Provide
 
 var matchBidNotFound = regexp.MustCompile("^.+bid not found.+$")
 
+func observeRunner(f func () runner.Result, observer prometheus.Histogram) func() runner.Result{
```
boz, line 129: kinda feel like this can be added to a `runner.Do`-ish variant but we can... _do_ that later if we find that we want this elsewhere.

provider/cluster/kube/apply.go
```diff
@@ -14,16 +14,32 @@ import (
...
+	kubeCallsCounter.WithLabelValues("namespaces-get", label).Inc()
+
 	switch {
 	case err == nil:
 		obj, err = b.update(obj)
 		if err == nil {
 			_, err = kc.CoreV1().Namespaces().Update(ctx, obj, metav1.UpdateOptions{})
+			label := "success"
+			if err != nil {
+				label = "fail"
```
boz, line 30: it's really annoying that go doesn't have the tertiary operator.

boz: for reals I know it's lame but I would package this up into a little helper.

```go
kubeCallsCounter.WithLabelValues("namespaces-update", rlabel(err)).Inc()
```

provider/manifest/service.go
```diff
@@ -269,13 +268,13 @@ func (s *service) maybeRemoveWatchdog(deploymentID dtypes.DeploymentID) {
...
-		s.handleLease(lease)
+		s.handleLease(lease, false)
 	}
 }
 
-func (s *service) handleLease(ev event.LeaseWon) {
+func (s *service) handleLease(ev event.LeaseWon, isNew bool) {
 	// Only run this if configured to do so
-	if s.config.ManifestTimeout > time.Duration(0) {
+	if isNew && s.config.ManifestTimeout > time.Duration(0) {
```
boz, line 277: this is the accidental purge fix right?

boz (commented):
looks great.

You mind putting the bugfix in a separate PR?  It's an important commit and I'd like to have all the reporting around it and what not be clean if possible.

boz (approved):
Looks like the bugfix _was_ included in another PR.

provider/manifest/service.go
hydrogen18: yes

hydrogen18:
All comments have been addressed

boz (approved):
:+1:

PR #1237, opened 2021-04-30
feat(provider): set AKASH_OWNER env. var. for bid pricing script

hydrogen18:
Add an env. var. of AKASH_OWNER to the bid pricing script

### DIFF
--- a/provider/bidengine/pricing.go
+++ b/provider/bidengine/pricing.go
@@ -6,8 +6,10 @@ import (
 	"crypto/rand"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"math"
 	"math/big"
+	"os"
 	"os/exec"
 	"time"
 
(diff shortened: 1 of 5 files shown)


troian (approved):

provider/bidengine/pricing.go
```diff
@@ -293,6 +295,10 @@ func (ssp shellScriptPricing) CalculatePrice(ctx context.Context, gspec *dtypes.
 	outputBuf := &bytes.Buffer{}
 	cmd.Stdout = outputBuf
 
+	subprocEnv := os.Environ()
+	subprocEnv = append(subprocEnv, fmt.Sprintf("AKASH_OWNER=%s", owner))
```
troian, line 299: can we use prefix from config instead of direct `AKASH`?

troian: does subprocess also support multiple addresses in `AKASH_OWNER`

hydrogen18: I could make this configurable, but I guess why would it need to be? I was trying to keep the same format like I used here https://github.com/ovrclk/akash/blob/master/provider/cluster/kube/builder.go#L318

PR #1245, opened 2021-05-06
feat: update SDK to v0.42.4 and add upgrade handler

akhilkumarpilli:
* Updated cosmos-sdk version to v0.42.4
* Added upgrade handler with name `v0.42-upgrade`

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -63,7 +63,6 @@ import (
 	stakingkeeper "github.com/cosmos/cosmos-sdk/x/staking/keeper"
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 
-	"github.com/ovrclk/akash/app/migrations"
 	"github.com/ovrclk/akash/x/audit"
 	"github.com/ovrclk/akash/x/cert"
 	escrowkeeper "github.com/ovrclk/akash/x/escrow/keeper"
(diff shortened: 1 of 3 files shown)


troian (approved):

app/app.go
```diff
@@ -263,11 +262,8 @@ func NewApp(
 
 	app.keeper.upgrade = upgradekeeper.NewKeeper(skipUpgradeHeights, app.keys[upgradetypes.StoreKey], appCodec, homePath)
 
-	app.keeper.upgrade.SetUpgradeHandler("akashnet-2-upgrade-1", func(ctx sdk.Context, plan upgradetypes.Plan) {
-		migrations.MigrateAkashnet2Upgrade1(ctx, app.keeper.acct, app.keeper.bank, app.keeper.staking)
+	app.keeper.upgrade.SetUpgradeHandler("v0.42-upgrade", func(ctx sdk.Context, plan upgradetypes.Plan) {
 
-		// Set staking HistoricalEntries to 10000, required positive value for ibc
-		app.GetSubspace(stakingtypes.ModuleName).Set(ctx, stakingtypes.KeyHistoricalEntries, uint32(10000))
```
boz, line 270: Is there anything to upgrade?

akhilkumarpilli: We don't have anything to upgrade here. Just we need a handler to upgrade to v0.42.4. We might need to add some migrator functions when upgrading to v0.43. 

We can keep this PR as draft until v0.43 release. Once after that, we can modify this handler and add migrator functions if needed any.

PR #1249, opened 2021-05-06
fix(provider): Don't cancel the context until the code is done

hydrogen18:


boz (approved):
that'll do it.

:ship:

PR #1251, opened 2021-05-07
fix(provider): Allow the hostname with a dash in SDL

hydrogen18:
Just fixed a missing escape in my regex

### DIFF
--- a/validation/manifest.go
+++ b/validation/manifest.go
@@ -11,12 +11,11 @@ import (
 	"github.com/ovrclk/akash/manifest"
 	"github.com/ovrclk/akash/types"
 	dtypes "github.com/ovrclk/akash/x/deployment/types"
+	k8svalidation "k8s.io/apimachinery/pkg/util/validation"
 )
 
 var (
 	serviceNameValidationRegex = regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`)
-	envVarValidatorRegex       = regexp.MustCompile(`^[-._a-zA-Z][-._a-zA-Z0-9]*$`)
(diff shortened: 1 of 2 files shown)


validation/manifest.go
```diff
@@ -16,7 +16,7 @@ import (
 var (
 	serviceNameValidationRegex = regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`)
 	envVarValidatorRegex       = regexp.MustCompile(`^[-._a-zA-Z][-._a-zA-Z0-9]*$`)
-	hostnameRegex              = regexp.MustCompile(`^[[:alnum:],-,\.]+\.[[:alpha:]]{2,}$`)
+	hostnameRegex              = regexp.MustCompile(`^[[:alnum:],\-,\.]+\.[[:alpha:]]{2,}$`)
```
boz, line 19: This is strange - I think it shouldn't need to be escaped inside of a `[]`?  `envVarValidatorRegex` doesn't seem to need to be.

Anyways, I think that we should use the validation that comes from k8s for this stuff:

* https://github.com/kubernetes/apimachinery/blob/8daf28983e6ecf28bd8271925ee135c1179ad13a/pkg/util/validation/validation.go#L208
* https://github.com/kubernetes/apimachinery/blob/8daf28983e6ecf28bd8271925ee135c1179ad13a/pkg/util/validation/validation.go#L408

boz (commented):
Some comments inline

validation/manifest.go
hydrogen18: Done.

hydrogen18:
Gonna merge as soon as tests are passing

PR #1253, opened 2021-05-12
refactor(provider/cli): show logs/events for all leases by default

### LINKED ISSUE #1250: lease-logs, lease-events: show for all leases by default
For `akash provider {lease-logs,lease-events}`, show logs and events for all leases that match the filtering that is passed in.

* `--dseq` is required
* `--gseq`,`--oseq`,`--provider` all optionally specify a lease

This is similar to what was done recently with `akash provider send-manifest` in #1177

troian:
fixes #1250

lease events example
```json
{
  "type": "Warning",
  "time": "0001-01-01T00:00:00Z",
  "reason": "Failed",
  "note": "Error: ImagePullBackOff",
  "object": {
    "kind": "Pod",
    "namespace": "4a7d0egn6o6n277feui0o8457nvr4gt6k3qjvjm0qsdi0",
    "name": "web-7654dc49f7-5mhv6"
  },
  "lease_id": {
    "owner": "akash1zlqr03j0mk04guz9k9t8p0w4tvdqxyzkz2d0pw",
    "dseq": 1,
    "gseq": 1,
    "oseq": 1,
    "provider": "akash1lxl3a62vn0rkzxacyz4zky679jmcwl3n2eqgcw"
  }
}
```

log output as json
```json
{
  "name": "web-68d66d985b-k426m",
  "message": "2021-05-13T07:00:59+0000 ERROR An error is usually an exception that has been caught and not handled.",
  "lease_id": {
    "owner": "akash1zlqr03j0mk04guz9k9t8p0w4tvdqxyzkz2d0pw",
    "dseq": 2,
    "gseq": 1,
    "oseq": 1,
    "provider": "akash1lxl3a62vn0rkzxacyz4zky679jmcwl3n2eqgcw"
  }
}
```

log output as text
```
[akash1zlqr03j0mk04guz9k9t8p0w4tvdqxyzkz2d0pw/2/1/1/akash1lxl3a62vn0rkzxacyz4zky679jmcwl3n2eqgcw][web-68d66d985b-k426m] 2021-05-13T07:00:15+0000 INFO This is less important than debug log and is often used to provide context in the current task.
```

### DIFF
--- a/provider/cmd/leaseEvents.go
+++ b/provider/cmd/leaseEvents.go
@@ -2,15 +2,19 @@ package cmd
 
 import (
 	"crypto/tls"
+	"sync"
 
 	sdkclient "github.com/cosmos/cosmos-sdk/client"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	cmdcommon "github.com/ovrclk/akash/cmd/common"
+	cltypes "github.com/ovrclk/akash/provider/cluster/types"
+	dtypes "github.com/ovrclk/akash/x/deployment/types"
+	mtypes "github.com/ovrclk/akash/x/market/types"
 	"github.com/spf13/cobra"
 
 	akashclient "github.com/ovrclk/akash/client"
-	cmdcommon "github.com/ovrclk/akash/cmd/common"
 	gwrest "github.com/ovrclk/akash/provider/gateway/rest"
 	cutils "github.com/ovrclk/akash/x/cert/utils"
-	mcli "github.com/ovrclk/akash/x/market/client/cli"
 )
 
 func leaseEventsCmd() *cobra.Command {
@@ -36,12 +40,20 @@ func doLeaseEvents(cmd *cobra.Command) error {
 		return err
 	}
 
-	prov, err := providerFromFlags(cmd.Flags())
+	cert, err := cutils.LoadAndQueryCertificateForAccount(cmd.Context(), cctx, cctx.Keyring)
 	if err != nil {
 		return err
 	}
 
-	bid, err := mcli.BidIDFromFlagsForOwner(cmd.Flags(), cctx.FromAddress)
+	dseq, err := dseqFromFlags(cmd.Flags())
+	if err != nil {
+		return err
+	}
+
+	leases, err := leasesForDeployment(cmd.Context(), cctx, cmd.Flags(), dtypes.DeploymentID{
+		Owner: cctx.GetFromAddress().String(),
+		DSeq:  dseq,
+	})
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


provider/cmd/leaseEvents.go
```diff
@@ -36,54 +41,88 @@ func doLeaseEvents(cmd *cobra.Command) error {
...
 	}
 
-	follow, err := cmd.Flags().GetBool("follow")
+	svcs, err := cmd.Flags().GetString(FlagService)
 	if err != nil {
 		return err
 	}
 
-	cert, err := cutils.LoadAndQueryCertificateForAccount(cmd.Context(), cctx, cctx.Keyring)
+	follow, err := cmd.Flags().GetBool("follow")
```
hydrogen18, line 67: Should we try and make this flag a const variable somewhere?

provider/cmd/leaseEvents.go
```diff
@@ -36,54 +41,88 @@ func doLeaseEvents(cmd *cobra.Command) error {
...
 
-	for res := range result.Stream {
-		if err = cmdcommon.PrintJSON(cctx, res); err != nil {
-			break
+	streams := make([]result, 0, len(leases))
+
+	for _, lid := range leases {
+		var stream result
+		prov, _ := sdk.AccAddressFromBech32(lid.Provider)
+		gclient, err := gwrest.NewClient(akashclient.NewQueryClientFromCtx(cctx), prov, []tls.Certificate{cert})
```
hydrogen18, line 87: Can this code just call `akashclient.NewQueryClientFromCtx(cctx)` once outside the loop rather on each iteration of the loop?

provider/cmd/leaseEvents.go
```diff
@@ -36,54 +41,88 @@ func doLeaseEvents(cmd *cobra.Command) error {
...
+		if err == nil {
+			stream.stream, stream.error = gclient.LeaseEvents(ctx, lid, svcs, follow)
+			stream.lid = lid
```
hydrogen18, line 90: This could be set in an initializer up at the top of the loop like `stream := result{ lid: lid }` so that it is always present in the resulting slice.

provider/cmd/leaseEvents.go
```diff
@@ -36,54 +41,88 @@ func doLeaseEvents(cmd *cobra.Command) error {
...
+
+		streams = append(streams, stream)
 	}
 
-	select {
-	case msg, ok := <-result.OnClose:
-		if ok && msg != "" {
-			_ = cctx.PrintString(msg + "\n")
+	var wgStreams sync.WaitGroup
+	wgStreams.Add(len(streams))
```
hydrogen18, line 99: The usage of this is kind of weird, I think it would be simpler like this 

```
var wgStreams sync.WaitGroup

for _, stream := range streams {

if stream.error == nil {
wgStreams.Add(1)
go func (stream result) {
  defer wgStreams.Done()
....
```

That way you can avoid the unusual pattern of `if stream.error != nil { wgStreams.Done() }`

provider/cmd/leaseEvents.go
```diff
@@ -36,54 +41,88 @@ func doLeaseEvents(cmd *cobra.Command) error {
...
+	follow, err := cmd.Flags().GetBool("follow")
 	if err != nil {
 		return err
 	}
 
-	gclient, err := gwrest.NewClient(akashclient.NewQueryClientFromCtx(cctx), prov, []tls.Certificate{cert})
-	if err != nil {
-		return err
-	}
+	ctx, cancel := context.WithCancel(cmd.Context())
```
hydrogen18, line 72: Is there a signal handler that comes along and cancels this in the case of the user pressing `CTRL + C` ?

provider/cmd/leaseEvents.go
troian: 💯

provider/cmd/leaseEvents.go
troian: there can be a few different leases on different providers for same deployment

provider/cmd/leaseEvents.go
troian: 💯

hydrogen18 (approved):

provider/cmd/leaseEvents.go
```diff
@@ -36,54 +40,84 @@ func doLeaseEvents(cmd *cobra.Command) error {
...
 	}
 
-	select {
-	case msg, ok := <-result.OnClose:
-		if ok && msg != "" {
-			_ = cctx.PrintString(msg + "\n")
+	var wgStreams sync.WaitGroup
+
+	for _, stream := range streams {
+		if stream.error == nil {
```
boz, line 98: save an indentation & make it easier to read:
```go
if steam.error != nil {
    continue
}
```

provider/cmd/leaseEvents.go
```diff
@@ -36,54 +40,84 @@ func doLeaseEvents(cmd *cobra.Command) error {
...
+						cltypes.LeaseEvent `json:",inline"`
+						Lid                mtypes.LeaseID `json:"lease_id"`
+					}
+
+					evt := &logEntry{
+						LeaseEvent: res,
+						Lid:        stream.lid,
+					}
+
+					_ = cmdcommon.PrintJSON(cctx, evt)
```
boz, line 114: does `cmdcommon.PrintJSON` have a mutex?  If not I think that different prints may be interleaved together.

you could change this to

```go
outch <- evt
```

and have something like:

```go
go func(){
  for evt := range outch {
    _cmdcommon.PrintJSON(cctx,evt)
  }
}()
```

and

```go
wgstreams.Wait()
close(outch)
```

provider/cmd/leaseEvents.go
boz: I think you might have forgotten this particular `"follow"` - you used the var elsewhere.

provider/cmd/leaseLogs.go
```diff
@@ -78,47 +89,65 @@ func doServiceLogs(cmd *cobra.Command) error {
...
+
+		streams = append(streams, stream)
 	}
 
-	printFn := func(msg gwrest.ServiceLogMessage) error {
-		fmt.Printf("[%s] %s\n", msg.Name, msg.Message)
-		return nil
+	var wgStreams sync.WaitGroup
+
+	printFn := func(lid mtypes.LeaseID, msg gwrest.ServiceLogMessage) {
```
boz, line 118: you could do a

```go
mtx.Lock()
defer mtx.Unlock()
```

here to prevent interleaving.

provider/cmd/leaseLogs.go
```diff
@@ -78,47 +89,65 @@ func doServiceLogs(cmd *cobra.Command) error {
...
 	}
 
-	for res := range result.Stream {
-		err = printFn(res)
-		if err != nil {
-			return err
-		}
-	}
+	for _, stream := range streams {
+		if stream.error == nil {
```
boz, line 138: again please get past the preconditions quickly and without bumping indentation if you can.

boz (commented):
Looks great!  Couple comments.

provider/cmd/leaseEvents.go
troian: good point, stdout (default stream) does not guarantee threadsafety

boz (approved):
:rocket: :ship:

PR #1257, opened 2021-05-19
feat(provider): Add a label for the Kubernetes ingress controller nod…

hydrogen18:
This just requires the NGINX ingress host to have a label of `akashRole: ingress`


_run/ingress-nginx.yaml
```diff
@@ -401,6 +401,7 @@ spec:
               memory: 90Mi
       nodeSelector:
         kubernetes.io/os: linux
+        akashRole: ingress
```
boz, line 404: Let's use `akash.network/` prefix for all of our labels

boz (commented):
looks great, small label name request inline.

PR #1264, opened 2021-05-27
ryan/blocks remaining v2

### LINKED ISSUE #1252: query escrow account status
write Akash command to return estimated time until escrow account state=overdrawn

rsum1:
fixes #1252


x/escrow/client/cli/query.go
```diff
@@ -1,7 +1,142 @@
...
+			}
+
+			leases := make([]marketTypes.Lease, 0)
+			for _, lease := range leasesResponse.Leases {
+				leases = append(leases, lease.Lease)
+			}
+
+			// Fetch the balance of the escrow account
+			deploymentClient := deploymentTypes.NewQueryClient(clientCtx)
+			// fmt.Printf("length of leases, %d \n", len(leases))
```
hydrogen18, line 82: We can get rid of this line now.

x/escrow/client/cli/query.go
```diff
@@ -1,7 +1,142 @@
...
+				amount := lease.Price.Amount
+				balance := res.EscrowAccount.Balance.Amount
+				settledAt := res.EscrowAccount.SettledAt
+				blockchainHeight, err := cli.CurrentBlockHeight(clientCtx)
+				if err != nil {
+					return err
+				}
+				balanceRemain := balance.Int64() - ((int64(blockchainHeight) - settledAt) * (amount.Int64()))
+				blocksRemain := balanceRemain / amount.Int64()
+				blocksPerDay := 86400 / 6.5
```
hydrogen18, line 101: We should add a comment to indicate that the 6.5 is the estimated block time. It'd be neat to query the blockchain to try and estimate this value, but a clearly labeled constant like 

```
const secondsPerBlock = 6.5
```

Should be fine for now

x/escrow/client/cli/query.go
```diff
@@ -1,7 +1,142 @@
...
+				daysRemain := blocksRemain / int64(blocksPerDay)
+
+				output := struct {
+					BalanceRemain int64 `json:"balance_remaining" yaml:"balance_remaining"`
+					BlocksRemain  int64 `json:"blocks_remaining" yaml:"blocks_remaining"`
+					DaysRemain    int64 `json:"days_remaining" yaml:"days_remaining"`
```
hydrogen18, line 107: Maybe we should label this as "estimated_days_remaining" since the blocks are exact but the days are just an estimate

x/escrow/client/cli/query.go
```diff
@@ -1,7 +1,142 @@
...
+				outputSlice = append(outputSlice, output)
+
+			}
+			outputType, err := cmd.Flags().GetString("output")
+			if outputType == "json" {
+				data, err := json.MarshalIndent(outputSlice, " ", "\t")
+				if err != nil {
+					return err
+				}
+				clientCtx.PrintBytes(data)
```
hydrogen18, line 122: There is a slight optimization here, you can declare the code like this and just set the values for data & err inside the the if / else and then use them outside the statement

```
var data []byte
if outputType =="json" {
  data, err = json.MarshalIndent
} else {
  data, err = yaml.Marshal
}

// Check for an error serializing
if err != nil {
  return err
}


//Write out the result
clientCtx.PrintBytes(data)
```

x/escrow/client/cli/query.go
```diff
@@ -1,7 +1,142 @@
...
+				res, err := deploymentClient.Deployment(ctx, &deploymentTypes.QueryDeploymentRequest{
+					ID: lease.LeaseID.DeploymentID(),
+				})
+				if err != nil {
+					return err
+				}
+				amount := lease.Price.Amount
+				balance := res.EscrowAccount.Balance.Amount
+				settledAt := res.EscrowAccount.SettledAt
+				blockchainHeight, err := cli.CurrentBlockHeight(clientCtx)
```
boz, line 95: this can be done once, outside of the loop.

x/escrow/client/cli/query.go
boz: Yeah or make `blocksPerDay` a constant

```go
const blocksPerDay = int(24 * 60 * 60 / 6.5)
```
or something if you can.

x/escrow/client/cli/query.go
```diff
@@ -1,7 +1,142 @@
...
+
+			// Fetch the balance of the escrow account
+			deploymentClient := deploymentTypes.NewQueryClient(clientCtx)
+			// fmt.Printf("length of leases, %d \n", len(leases))
+			outputSlice := make([]interface{}, 0)
+			for _, lease := range leases {
+				//  Fetch the time of last settlement
+				res, err := deploymentClient.Deployment(ctx, &deploymentTypes.QueryDeploymentRequest{
+					ID: lease.LeaseID.DeploymentID(),
+				})
```
boz, line 88: this should be done once, outside of the loop.

x/escrow/client/cli/query.go
```diff
@@ -1,7 +1,142 @@
...
+				}
+				amount := lease.Price.Amount
+				balance := res.EscrowAccount.Balance.Amount
+				settledAt := res.EscrowAccount.SettledAt
+				blockchainHeight, err := cli.CurrentBlockHeight(clientCtx)
+				if err != nil {
+					return err
+				}
+				balanceRemain := balance.Int64() - ((int64(blockchainHeight) - settledAt) * (amount.Int64()))
+				blocksRemain := balanceRemain / amount.Int64()
```
boz, line 100: this only works if there is one lease.   `balance` here is the balance of the escrow account that funds all of the leases.

It'd be best to `sum(amount from every lease)` to get the aggregated price per block.

boz (changes requested):
Some small comments inline.  The blocker is that it doesn't work for deployments with more than one lease (unless I'm missing something).

hydrogen18 (approved):

x/escrow/client/cli/query.go
```diff
@@ -1,7 +1,153 @@
...
+
+			owner, err := cmd.Flags().GetString("owner")
+			if err != nil {
+				return err
+			}
+
+			dseq, err := cmd.Flags().GetUint64("dseq")
+			if err != nil {
+				return err
+			}
```
boz, line 55: [`cli.DeploymentIDFromFlags()`](https://github.com/ovrclk/akash/blob/86436f8a8b5852ea22d54b8dec55bb5125c82873/x/deployment/client/cli/flags.go#L29)

x/escrow/client/cli/query.go
```diff
@@ -1,7 +1,153 @@
...
+			totalLeaseAmount := cosmosTypes.NewInt(0)
+			blockchainHeight, err := cli.CurrentBlockHeight(clientCtx)
+			if err != nil {
+				return err
+			}
+			if 0 == len(leases) {
+				return errNoLeaseMatches
+			}
+			for _, lease := range leases {
+				//  Fetch the time of last settlement
```
boz, line 94: :scissors:

x/escrow/client/cli/query.go
```diff
@@ -1,7 +1,153 @@
...
+				ID: deploymentTypes.DeploymentID{Owner: owner, DSeq: dseq},
+			})
+			if err != nil {
+				return err
+			}
+			balance := res.EscrowAccount.Balance.Amount
+			settledAt := res.EscrowAccount.SettledAt
+			if err != nil {
+				return err
+			}
```
boz, line 110: ?

x/escrow/client/cli/query.go
```diff
@@ -1,7 +1,153 @@
...
+			balanceRemain := balance.Int64() - ((int64(blockchainHeight) - settledAt) * (totalLeaseAmount.Int64()))
+			blocksRemain := balanceRemain / totalLeaseAmount.Int64()
+			const secondsPerDay = 24 * 60 * 60
+			const secondsPerBlock = 6.5
+			// Calculate blocks per day by using 6.5 seconds as average block-time
+			blocksPerDay := (secondsPerDay / secondsPerBlock)
```
boz, line 116: towards the top of the file:

```go
const (
  secondsPerDay = xxx
  secondsPerBlock = yyy
  blocksPerDay = secondsPerDay / secondsPerBlock
)
```

x/escrow/client/cli/query.go
```diff
@@ -1,7 +1,153 @@
...
+			estimatedDaysRemain := blocksRemain / int64(blocksPerDay)
+			output := struct {
+				BalanceRemain       int64 `json:"balance_remaining" yaml:"balance_remaining"`
+				BlocksRemain        int64 `json:"blocks_remaining" yaml:"blocks_remaining"`
+				EstimatedDaysRemain int64 `json:"estimated_days_remaining" yaml:"estimated_days_remaining"`
```
boz, line 121: I think that this should be a [`time.Duration`](https://golang.org/pkg/time/#Duration) and that the default [`String()`](https://golang.org/pkg/time/#Duration.String) formatting is good for now.

boz (changes requested):
Looking good.  A couple more changes requested inline.

x/escrow/client/cli/query.go
```diff
@@ -1,7 +1,144 @@
...
+				data, err = json.MarshalIndent(output, " ", "\t")
+			} else {
+				data, err = yaml.Marshal(output)
+			}
+
+			if err != nil {
+				return err
+			}
+
+			return clientCtx.PrintBytes(data)
```
boz, line 135: does this print is as hex bytes or as readable output?

boz (approved):
:rocket:

x/escrow/client/cli/query.go
rsum1: sample output looks like this:

```
balance_remaining: 4.888378e+06
blocks_remaining: 2.444189e+06
estimated_time_remaining: 4413h7m8s
```

PR #1265, opened 2021-05-27
fix(provider): eliminate resource leaks causing OOM

hydrogen18:
We were running out of memory in prod.

Using the pprof package I could see the number of goroutines in pubsub code kept going up. Also a bunch of them sitting on `WatchChannel` in the `go-lifecycle` package

I changed the following.

1. Make sure we call `.Close()` on all pubsub subscribers
2. Remove a pubsub subscriber we aren't using
3. Move call to `.ShutdownInitiated` to outside of a loop, so that the `WatchChannel` always has the opportunity to terminate


boz (approved):
looks great

PR #1273, opened 2021-06-07
fix(provider): add more debugging around inventory

hydrogen18:
This just breaks out the reservations metrics into pending & active. Also added some logging around the number of reserve & unreserve calls that succeed.

### DIFF
--- a/provider/cluster/inventory.go
+++ b/provider/cluster/inventory.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"sync/atomic"
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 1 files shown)


provider/cluster/inventory.go
```diff
@@ -270,30 +280,51 @@ func (is *inventoryService) updateInventoryMetrics(inventory []ctypes.Node) {
...
+	pendingStorageTotal := 0.0
+	pendingEndpointsTotal := 0.0
 
-	cpuTotal := 0.0
-	memoryTotal := 0.0
-	storageTotal := 0.0
-	endpointsTotal := 0.0
 	allocated := 0.0
 	for _, reservation := range reservations {
+		cpuTotal := &pendingCPUTotal
```
boz, line 297: why are pointers needed here?

hydrogen18: To select the destination of what set of values to update.

boz: ah, I see.  thanks.

boz (approved):

PR #1279, opened 2021-06-12
fix(provider): Correct bugs around inventory management

hydrogen18:
So we had two issues

1. Calling `.Sub()` on `ResourceUnits` was not idempotent. It mutated the receiver. This is wrong.
2. We had a race where we marked a deployment as allocated, so we consider it to be already in the resource counts from kube. But we don't fetch a new inventory data from kube, so this leads to reservations being made that should not be.

I wrote more tests around the provider's inventory code. All I really did here was

1. Correct a copy of a pointer to be a copy of a value
2. Don't process reservations after a deployment is marked `allocated` until we get a new update of the inventory from kubernetes
3. Update some tests that are failing as a result of this fix.

### DIFF
--- a/provider/cluster/inventory.go
+++ b/provider/cluster/inventory.go
@@ -346,6 +346,18 @@ func (is *inventoryService) run(reservations []*reservation) {
 
 	var fetchCount uint
 
+	var reserveChLocal <-chan inventoryRequest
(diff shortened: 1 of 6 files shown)


troian (approved):

boz (approved):
:rocket:

PR #1283, opened 2021-06-16
fix(cli): make owner flag optional

### LINKED ISSUE #1274: set the default owner as the address from the key specified with --from flag
The akash address used in the `--owner` flag can be derived from the key specified with the `--from` flag.

rsum1:
fixes #1274

### DIFF
--- a/x/market/client/cli/tx.go
+++ b/x/market/client/cli/tx.go
@@ -6,10 +6,12 @@ import (
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/flags"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/spf13/cobra"
+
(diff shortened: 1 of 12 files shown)


x/market/client/cli/tx.go
```diff
@@ -155,7 +156,7 @@ func cmdLeaseCreate(key string) *cobra.Command {
 				return err
 			}
 
-			id, err := BidIDFromFlagsWithoutCtx(cmd.Flags())
```
boz, line 158: I think we can get rid of these `*With(out)?Ctx` functions :confetti_ball:

boz (commented):
Looks great.  I think that there's some cleanup we can do (comment inline)

x/market/client/cli/tx.go
troian: 💯

hydrogen18 (approved):

PR #1293, opened 2021-06-29
Add provider lease-shell command

hydrogen18:
This adds a way to get a lease shell into an existing container that is already running. There are basically two ways to do this

1. Proxy everything through the provider and do auth there
2. Issue temporary authorizations via kubernetes RBAC so a customer can directly connect to their containers via kubectl or similar commands.

I chose route 1. The other option is to use ephemeral containers that share cgroups & namespaces with the exiting containers. If we want to do this in the future we can add that later without any real rework. The main advantage to doing that is the customer's container doesn't need to actually have a shell stored within it.
 
The way I implemented this was by reverse engineering the implementation of `kubectl exec` one step at a time, then splitting our implementation between the provider & the CLI. The provider proxies everything over a single websocket at this time. I've tested every scenario I can come up with and tried to make the error messages as helpful as possible to the end user. The biggest stumbling block is when users deploy containers which either don't exist (image can't be pulled) or the container crashes immediately on startup. In that case, they won't be able to get a shell into them since they never actually run. This isn't actually a step backwards however, it just indicates we need better visibility into deployment health or status.

The biggest drawback I think is the amount of code I had to rewrite. It's large enough that most of the implementation got split into its own files. I added some simple unit tests around utilities I created for use from within the code. Actually testing the implementation is quite difficult and would require a huge number of additional mocks to be generated at this time. I added another integration test that is larger identical to the others. It tests that it can remotely invoke a command, then tests the error cases that the client can commonly run into.

There are basically 4 components that live as part of this now

1. The provider  CLI command, which also does terminal management. Since this command runs an interactive terminal, it's important that the state be saved & restored in that case. I'm using the kubernetes code for this directly
2. The CLI "client", which handles dialing a websocket and de-multiplexing everything associated with this. All data comes back over the websocket as a binary message and has an identifier prepended to it. The client is responsible for directing this data to stdout, stderr, etc.
3. The provider router, which is responsible for accepting the websocket request. This has to the opposite of the client, and multiplex all the data onto the websocket. It also tries to validate all the parameters before reaching out to kubernetes.
4. The kubernetes client code, which actually talks to kubernetes. I don't really like the way the golang API for this works. It's very awkward and I had to instantiate many objects by hand in order to get it to work. You also have to use two clients to talk to kubernetes, one to build the request and another SPDY client to send it to kubernetes. If you build the request improperly, you get a silent failure. It was however faster to go through the trial and error process to figure this out rather than implement my own client library for talking to kubernetes.

The kubernetes client in part 4 is quite frustrating with respect to errors. Essentially all errors come back from kubernetes as untyped when something fails. It falls back to string matching in order to identify errors, but avoids ever just returning errors verbatim. There is no way to safely do this, since we don't know what information kubernetes could return.

Overall I think this is a pretty food first attempt. None of this requires blockchain updates, so we can iterate on it as fast as we would like to.


provider/cluster/kube/client_exec.go
```diff
@@ -0,0 +1,207 @@
...
+
+func (sp sortablePods) Less(i, j int) bool {
+	return strings.Compare(sp[i].Name, sp[j].Name) == -1
+}
+
+func (sp sortablePods) Swap(i, j int) {
+	sp[i], sp[j] = sp[j], sp[i]
+}
+
+func (c *client) Exec(ctx context.Context, leaseID mtypes.LeaseID, serviceName string, podIndex uint, cmd []string, stdin io.Reader,
```
troian, line 46: can we use container name instead of pod index here? won't need to fancy with sorting pod names

actually, when deployment object is being created from the SDL, container name is set to same value as service name.
one of the examples below from SJC1
```
  template:
    metadata:
      creationTimestamp: null
      labels:
        akash.network: 'true'
        akash.network/manifest-service: mongo
        akash.network/namespace: 0lgvkaccl4gmi7lef591mnpu3tshmhlg0en1hsqv011di
    spec:
      containers:
        - name: mongo
          image: 'mongo:latest'
          args:
            - '--storageEngine=wiredTiger'
          ports:
            - containerPort: 27017
              protocol: TCP
          env:
```

troian (changes requested):

provider/cluster/kube/client_exec.go
troian: nvm, i'm wrong

provider/gateway/rest/router.go
```diff
@@ -136,9 +142,234 @@ func newRouter(log log.Logger, addr sdk.Address, pclient provider.Client) *mux.R
...
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				for {
+					msgType, data, err := ws.ReadMessage()
+					if err != nil {
+						return
+					}
+
+					if msgType == websocket.BinaryMessage && len(data) > 1 {
```
boz, line 277: exit early to prevent nesting
```go
if msgType != websocket.BinaryMessage || len(data) <= 1 {
  continue
}
```

provider/gateway/rest/router.go
```diff
@@ -136,9 +142,234 @@ func newRouter(log log.Logger, addr sdk.Address, pclient provider.Client) *mux.R
...
+						msgID := data[0]
+						msg := data[1:]
+						if msgID == LeaseShellCodeStdin {
```
boz, line 280: :nail_care: `switch msgID {`

util/wsutil/wsutil.go
```diff
@@ -0,0 +1,40 @@
...
+
+type WsWriterWrapper struct {
+	connection wrappedConnection
+
+	id  byte
+	buf bytes.Buffer
+	l   sync.Locker
+}
+
+func NewWsWriterWrapper(conn wrappedConnection, id byte, l sync.Locker) WsWriterWrapper {
```
boz, line 21: Can you keep `WsWriterWrapper` private and return an [`io.Writer`](https://golang.org/pkg/io/#Writer) here?

provider/cluster/kube/client_exec.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil, cluster.ErrExecServiceNotRunning
+	}
+
+	// check that the requested pod is within the range
+	if podIndex >= uint(len(pods.Items)) {
+		return nil, fmt.Errorf("%w: valid range is [0, %d]", cluster.ErrExecPodIndexOutOfRange, len(pods.Items)-1)
+	}
+
+	// sort the pods, since we have no idea what order kubernetes returns them in
+	podsEff := sortablePods(pods.Items)
```
boz, line 114: :+1:

provider/gateway/rest/router.go
```diff
@@ -136,9 +142,234 @@ func newRouter(log log.Logger, addr sdk.Address, pclient provider.Client) *mux.R
...
+
+		podIndexStr := vars.Get("podIndex")
+		if len(podIndexStr) == 0 {
+			localLog.Error("missing parameter podIndex")
+			rw.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		var podIndex uint
+		_, err = fmt.Sscanf(podIndexStr, "%d", &podIndex)
+		if err != nil {
```
boz, line 235: can use [`strconv.ParseUint`](https://golang.org/pkg/strconv/#ParseUint) here.

Also, both of these would be a request error of some sort I think.

provider/gateway/rest/router.go
```diff
@@ -136,9 +142,234 @@ func newRouter(log log.Logger, addr sdk.Address, pclient provider.Client) *mux.R
...
+		if isTty {
+			terminalSizeUpdate = make(chan remotecommand.TerminalSize, 1)
+			tsq = channelToTerminalSizeQueue(terminalSizeUpdate)
+		}
+
+		if connectStdin {
+			stdinPipeIn, stdinPipeOut = io.Pipe()
+
+			wg.Add(1)
+			go func() {
```
boz, line 269: I think that this should be a function on its own for readability.

boz (commented):
Looks great, a couple comments inline.  I'll take another pass at it soon.

provider/cluster/client.go
```diff
@@ -19,7 +22,20 @@ import (
 	mtypes "github.com/ovrclk/akash/x/market/types"
 )
 
-var _ Client = (*nullClient)(nil)
+var (
+	_ Client = (*nullClient)(nil)
+	// Errors types returned by the Exec function on the client interface
+	ErrExec                        = errors.New("remote command execute error")
+	ErrExecNoServiceWithName       = fmt.Errorf("%w: no such service exists with that name", ErrExec)
```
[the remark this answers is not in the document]
hydrogen18: Doesn't that do the same thing as `fmt.Errorf` ?

hydrogen18:
@troian since im using a websocket do I need to setup the pong handler goroutine after the websocket is established?

troian:
yeah, better to have keepalive messages going through to maintain tcp session alive

hydrogen18:
I think I have addressed all the comments at this point in time

provider/cluster/client.go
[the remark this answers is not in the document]
boz: ~It creates a new `error` that embeds the previous `error` so you can have full access to the wrapped `error` it instead of just figuring it out by parsing strings.~

It looks like [they are the same now](https://golang.org/pkg/fmt/#Errorf)

provider/gateway/rest/router.go
```diff
@@ -136,9 +141,217 @@ func newRouter(log log.Logger, addr sdk.Address, pclient provider.Client) *mux.R
...
+
+		}
+
+		l := &sync.Mutex{}
+		stdout := wsutil.NewWsWriterWrapper(shellWs, LeaseShellCodeStdout, l)
+		stderr := wsutil.NewWsWriterWrapper(shellWs, LeaseShellCodeStderr, l)
+
+		subctx, subcancel := context.WithCancel(req.Context())
+		wg.Add(1)
+		go func() {
```
boz, line 279: standalone function

provider/gateway/rest/router.go
```diff
@@ -136,9 +141,217 @@ func newRouter(log log.Logger, addr sdk.Address, pclient provider.Client) *mux.R
...
+			defer wg.Done()
+			pingTicker := time.NewTicker(pingPeriod)
+			defer pingTicker.Stop()
+
+			for {
+				select {
+				case <-pingTicker.C:
+					if err = shellWs.WriteControl(websocket.PingMessage, nil, time.Now().Add(5*time.Second)); err != nil {
```
boz, line 287: `const foo = 5 * time.Second`

provider/gateway/rest/router_shell.go
```diff
@@ -0,0 +1,62 @@
...
+			return shellWs.SetReadDeadline(time.Now().Add(pingWait))
+		})
+
+		msgType, data, err := shellWs.ReadMessage()
+		if err != nil {
+			return
+		}
+
+		// Just ignore anything not a binary message or that is empty
+		if msgType != websocket.BinaryMessage || len(data) == 0 {
```
boz, line 27: :pray:

boz (commented):
quick run through, will do again soon.

hydrogen18:
Comments have been addressed

provider/cmd/shell.go
```diff
@@ -0,0 +1,190 @@
...
+	wg := &sync.WaitGroup{}
+	ctx, cancel := context.WithCancel(cmd.Context())
+
+	if tsq != nil {
+		terminalResizes = make(chan remotecommand.TerminalSize, 1)
+		go func() {
+			for {
+				// this blocks waiting for a resize event, the docs suggest
+				// that this isn't the case but there is not a code path that ever does that
+				// so this goroutine is just left running until the process exits
```
boz, line 138: wat :cry:

hydrogen18: I read through every code path I could find in the package that defines this. I couldn't find anything that closed the channel, which would unblock the `Next()` method.

colinster:
This pull request has been mentioned on **Akash Support**. There might be relevant details there:

https://forum.akash.network/t/sharing-my-first-deploy-experience-feedback-welcome/605/6

PR #1294, opened 2021-06-30
WIP feat: handle workloads with persistent storage

troian:
1. update SDL to handle multiple storage requests
2. use statefulSet when deployment needs persistent storage

**Parts**
- [x] SDL
- [x] Resource
- [x] Proto
- [ ] Inventory
- [x] Attributes
- [x] Bidengine
- [x] Provider create/update
- [x] Manifest

**tests**
- [x] SDL
- [x] Resource
- [ ] Manifest
- [ ] K8S Integration


provider/cluster/kube/client.go
```diff
@@ -183,9 +185,25 @@ func (c *client) Deploy(ctx context.Context, lid mtypes.LeaseID, group *manifest
...
+
+		persistent := false
+		for i := range service.Resources.Storage {
+			attrVal := service.Resources.Storage[i].Attributes.Find(StorageAttributePersistent)
+			if persistent, _ = attrVal.AsBool(); persistent {
+				break
+			}
+		}
+
+		if !persistent {
```
hydrogen18, line 197: Can this be arranged so it isn't inverted? Just use `if persistent` followed by `else`

provider/cluster/kube/statefulset_builder.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func newStatefulSetBuilder(log log.Logger, settings Settings, lid mtypes.LeaseID, group *manifest.Group, service *manifest.Service) *statefulSetBuilder {
+	return &statefulSetBuilder{
+		workloadBuilder: newWorkloadBuilder(log, settings, lid, group, service),
+	}
+}
+
+func (b *statefulSetBuilder) create() (*appsv1.StatefulSet, error) { // nolint:golint,unparam
+	replicas := int32(b.service.Count)
```
hydrogen18, line 24: Do we need to check this for being 0 or less? It seems weird kubernetes even allows an `int` type instead of `uint`

provider/cluster/kube/workload_builder.go
```diff
@@ -0,0 +1,184 @@
...
+
+	if mem := b.service.Resources.Memory; mem != nil {
+		requestedMem := clusterUtil.ComputeCommittedResources(b.settings.MemoryCommitLevel, mem.Quantity)
+		kcontainer.Resources.Requests[corev1.ResourceMemory] = resource.NewQuantity(int64(requestedMem.Value()), resource.DecimalSI).DeepCopy()
+		kcontainer.Resources.Limits[corev1.ResourceMemory] = resource.NewQuantity(int64(mem.Quantity.Value()), resource.DecimalSI).DeepCopy()
+	}
+
+	for _, storage := range b.service.Resources.Storage {
+		attr := storage.Attributes.Find(StorageAttributePersistent)
+		if persistent, _ := attr.AsBool(); persistent {
```
hydrogen18, line 77: This can probably be a helper method on the `workloadBuilder` object

provider/cluster/kube/workload_builder.go
```diff
@@ -0,0 +1,184 @@
...
+		kcontainer.Ports = append(kcontainer.Ports, corev1.ContainerPort{
+			ContainerPort: int32(expose.Port),
+		})
+	}
+
+	for i, storage := range b.service.Resources.Storage {
+		attr := storage.Attributes.Find(StorageAttributeReadOnly)
+		mount, valid := attr.AsString()
+		if !valid {
+			continue
```
hydrogen18, line 111: Is this an error?

provider/cluster/kube/workload_builder.go
```diff
@@ -0,0 +1,184 @@
...
+		if persistent, _ := attr.AsBool(); persistent {
+			continue
+		}
+
+		requestedStorage := clusterUtil.ComputeCommittedResources(b.settings.StorageCommitLevel, storage.Quantity)
+		kcontainer.Resources.Requests[corev1.ResourceEphemeralStorage] = resource.NewQuantity(int64(requestedStorage.Value()), resource.DecimalSI).DeepCopy()
+		kcontainer.Resources.Limits[corev1.ResourceEphemeralStorage] = resource.NewQuantity(int64(storage.Quantity.Value()), resource.DecimalSI).DeepCopy()
+	}
+
+	// TODO: this prevents over-subscription. skip for now.
```
hydrogen18, line 86: I think this comment is obsolete at this point. I have no idea what it means anymore

sdl/storage.go
```diff
@@ -3,18 +3,80 @@ package sdl
 import (
 	"sort"
 
+	"github.com/pkg/errors"
```
hydrogen18, line 6: Did you mean to import `errors` here?

provider/cluster/kube/ingress_builder.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/tendermint/tendermint/libs/log"
+	netv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/ovrclk/akash/manifest"
+	"github.com/ovrclk/akash/provider/cluster/util"
+	mtypes "github.com/ovrclk/akash/x/market/types"
+)
+
+// ingress
```
hydrogen18, line 18: We probably can kill this comment

provider/cluster/kube/workload_builder.go
```diff
@@ -0,0 +1,178 @@
...
+
+	return kcontainer
+}
+
+func (b *workloadBuilder) persistentVolumeClaims() []corev1.PersistentVolumeClaim {
+	var pvcs []corev1.PersistentVolumeClaim
+
+	for i, storage := range b.service.Resources.Storage {
+		attr := storage.Attributes.Find(StorageAttributePersistent)
+		if persistent, valid := attr.AsBool(); !valid || !persistent {
```
hydrogen18, line 126: Do we need to check valid here?

sdl/storage.go
```diff
@@ -3,18 +3,143 @@ package sdl
...
 	"gopkg.in/yaml.v3"
 
 	"github.com/ovrclk/akash/types"
 )
 
+const (
+	storageAttributePersistence = string("persistence")
```
hydrogen18, line 13: Why are these cast to strings?

sdl/storage.go
```diff
@@ -3,18 +3,143 @@ package sdl
...
+	//    ephemeral considered as one without mount point
+	// 2. ephemeral cannot have persistence
+	// 3. ephemeral cannot be readOnly (for now. configMaps are considered as ephemeral as well)
+	// todo make this as helper
+	ephemeral := false
+	for _, nd := range nodes {
+		attr := make(map[string]string)
+
+		for _, nd := range nd.Attributes {
+			attr[nd.Key] = nd.Value
```
hydrogen18, line 96: This would be easier to read as 

```
attrPersistence := false
attrMount := false

for k, v := range nd.Attributes {
  if k == storageAttributePersistence {
    attrPersistence = len(v.Value) ! = 0 // Convert to bool
} else if k == storageAttributeMount {
   attrMount = len(v.Value) != 0 // Convert to bool
}
```

Then you can just use the variables below.

sdl/storage.go
```diff
@@ -3,18 +3,143 @@ package sdl
...
+		} else {
+			ephemeral = true
+		}
+	}
+
+	// sort storage slice in the following order
+	// 1. smaller size
+	// 2. if sizes are equal then one without mount point goes up
+	// 3. if both mount points present use lexical order
+	sort.Slice(nodes, func(i, j int) bool {
```
hydrogen18, line 114: I generally try and write this as an independent function, as it can be tested on its own that way.

hydrogen18:
Does Kubernetes support some way of attaching block storage at a cluster level? I'm guessing that then any pod can make a volume claim to it. All the stuff I did in the past was ephemeral.

troian:
depends on CSI driver. some supporting raw block storage. https://kubernetes-csi.github.io/docs/drivers.html.
in pvc we can set block parameter instead of filesystem if needed

types/attribute.go
```diff
@@ -134,3 +178,75 @@ loop:
...
+// example
+// storage/1/persistent: true
+// storage/1/class: nvme
+// storage/2/persistent: false
+//
+// returns
+// - - persistent: true
+//     class: nvme
+// - - persistent: false
+func (attr Attributes) FilterGroupsByPrefix(prefix string) []Attributes {
```
troian, line 220: @hydrogen18 @boz comments on that?

hydrogen18: So we're adding a hierarchy to the attribute names now?

troian: some sort of. if provider supports 2 storage classes with different prices we need a way to specify that so bid process goes right

hydrogen18: Even if a provider had multiple storage classes, how would the provider know which kubernetes nodes to allocate a deployment to & make that decision? It would also need to take into account the inventory of that specific storage class when it starts bidding

troian: attributes are to ensure provider can bid. where to place on provider including node inspection, that what i'm working on right now

hydrogen18: If someone has different hardware configurations (ssd vs mechanical disk for example) why wouldn't they just create two kubernetes clusters and run two providers? It seems like it would be much simpler, versus trying to refactor the provider to deal with a heterogeneous kubernetes environment.

troian: if provider has EBS it can then support multiple storage classes within same kubernetes instance

provider/cluster/kube/client.go
troian: 💯

sdl/storage.go
troian: supposed to be other type. removed

sdl/storage.go
troian: 💯

provider/bidengine/pricing.go
```diff
@@ -90,9 +91,9 @@ func (fp scalePricing) CalculatePrice(_ context.Context, _ string, gspec *dtypes
 		memoryQuantity = memoryQuantity.Mul(groupCount)
 		memoryTotal = memoryTotal.Add(memoryQuantity)
 
-		storageQuantity := decimal.NewFromBigInt(group.Resources.Storage.Quantity.Val.BigInt(), 0)
-		storageQuantity = storageQuantity.Mul(groupCount)
-		storageTotal = storageTotal.Add(storageQuantity)
+		// storageQuantity := decimal.NewFromBigInt(group.Resources.Storage.Quantity.Val.BigInt(), 0)
+		// storageQuantity = storageQuantity.Mul(groupCount)
+		// storageTotal = storageTotal.Add(storageQuantity)
```
boz, line 96: [ ] todo

provider/cluster/kube/client.go
```diff
@@ -187,9 +190,25 @@ func (c *client) Deploy(ctx context.Context, lid mtypes.LeaseID, group *manifest
...
+				break
+			}
+		}
+
+		if persistent {
+			if err := applyStatefulSet(ctx, c.kc, newStatefulSetBuilder(c.log, c.settings, lid, group, service)); err != nil {
+				c.log.Error("applying statefulSet", "err", err, "lease", lid, "service", service.Name)
+				return err
+			}
+		} else {
```
boz, line 207: should we always `applyDeployment`?

provider/cluster/kube/ingress_builder.go
```diff
@@ -0,0 +1,93 @@
...
+		expose:          expose,
+		hosts:           make([]string, len(expose.Hosts), len(expose.Hosts)+1),
+	}
+
+	copy(builder.hosts, expose.Hosts)
+
+	if settings.DeploymentIngressStaticHosts {
+		uid := ingressHost(lid, service)
+		host := fmt.Sprintf("%s.%s", uid, settings.DeploymentIngressDomain)
+		builder.hosts = append(builder.hosts, host)
```
boz, line 36: :+1:

provider/cluster/kube/netpol_builder.go
```diff
@@ -0,0 +1,423 @@
...
+				},
+			},
+		}
+		result = append(result, &policy)
+	}
+
+	return result, nil
+	/**
+			{
+				ObjectMeta: metav1.ObjectMeta{
```
boz, line 192: let's just get rid of this.

types/attribute.go
```diff
@@ -3,14 +3,16 @@ package types
...
+	"strings"
 
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"gopkg.in/yaml.v3"
 )
 
 const (
 	moduleName                = "akash"
-	attributeNameRegexpString = `^[a-zA-Z][\w-]{1,30}[a-zA-Z0-9]$`
+	attributeNameRegexpString = `^(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+(?:[a-z](?:[a-z0-9]{0,61}[a-z0-9])?)?$`
```
boz, line 15: This is too complicated of a regex to be used in a transaction IMHO.

types/attribute.go
boz: :thinking:

boz (commented):
Looks great!

Would it be reasonable to put the builders in their own package (`provider/cluster/kube/builder`)?

Have to think about the match expressions a bit.  Seems reasonable so far tho.

troian:
work has been moved to #1348 1348

Issue #1281, opened 2021-06-16
accept hosts: allow migrating to different deployment


boz:
Hosts listed in the [`accept`](https://docs.akash.network/reference/sdl#services-expose) field are currently bound to the first deployment that uses them.  There can be only one deployment bound to an "accept" host at a time.

Currently, modifying a deployment's resources requires creating a new deployment.

To allow for zero-downtime upgrades, we should allow migrating an "accept" host to a new deployment that is owned by the same account.

The simplest workflow for this might be a manual one - add a provider command and corresponding gateway endpoint to migrate ownership of a host from one deployment to another.

```sh
akash provider hostname migrate foobar.com --dseq 10
```

The `accept` field of both `sdl` files should contain the host; it should be ignored if the host is owned by another deployment.

hydrogen18:
For the hostname migration, consider this scenario:

1. I create a deployment (dseq 100) and it uses a host "foobar.com"
2. Later I create a deployment (dseq 200) and it uses a host "foobar.com"
3. dseq 200 comes online and is working, so I use the migrate command to move the hostname to dseq 200. 
4. At this point dseq 200 is answering requests for "foobar.com"
5. The provider is restarted

How does the provider on startup determine which dseq should hold the reservation for "foobar.com" ? Just query kubernetes Ingress objects and see where it lives presently? What if it doesn't exist in kubernetes, are we ok with the earliest dseq getting the hostname?

boz:
> Just query kubernetes Ingress objects and see where it lives presently?

That seems reasonable.

> ok with the earliest dseq getting the hostname?

Yeah, sure.  If we happen to have the lease object then [`CreatedAt`](https://github.com/ovrclk/akash/blob/5cbf7dec4ca32340e54d6d1afbd698f34bf92bf7/x/market/types/lease.pb.go#L144)  (block height) would capture "earliest" lease a bit more accurately, but if it's not there then I don't think it's worth pulling just for this right now.

hydrogen18:
As part of this task I am currently learning more about how kubernetes supports operators. The theory being that the hostname management responsibility possibly can be moved outside of the provider process at this time.

The plan for this is as follows

1. Remove all codes that creates or updates ingress objects from the provider
2. Create a CRD which has contains for a hostname & the kubernetes deployment to host it with
3. The provider creates (or updates) CRD instances to associate a hostname with a specific deployment
[...]

PR #1299, opened 2021-07-02
WIP: Transfer ingress domain from one deployment to the next

hydrogen18:
~~This is my first pass at transferring hostnames deployments to the next.~~ Presently, a deployment effectively "owns" a hostname. Once created no other deployment can claim that hostname. This becomes an issue when a user hosts their website on the Akash platform, because they inevitably want to update deployment (to scale up capacity, or whatever) with a new one. Basically we have to somehow support this, to give customers the option of a zero-down time deploy.

The goal of this PR is to move ownership from a deployment level to an owner address level (corresponding to the wallet that originally laid claim to the hostname). Under this system, a domain may be used by any deployment owned by the same owner.

A few basic remarks

1. domains are still 1st come 1st serve. This doesn't solve the domain squatting problem
2. A deployment still gets a procedurally generated hostname as well that looks like random characters & numbers
3. The existing workflow for customers not using domain names is unchanged
4. The existing workflow for customers that want to just teardown their existing lease and suffer some down time is unchanged
5. This does solve the problem of a domain squatter trying to "intercept" a domain they've identified as in use on a provider and hold if for ransom with their own deployment

The way the new system works is

1. If a deployment is the 1st to lay claim to a hostname, the deployment gets the domain and owns
2. If a deployment is not the 1st to lay claim to a hostname, the deployment gets to use the domain name only if the owner addresses match. The existing deployment is evicted from the hostname **when the user requests it.** (see implementation below)

For case 1, this is the same way it works today.

For case 2, the provider considers the hostname to be immediately **owned by the account address**. However, the actual implementation creates all the necessary kubernetes resources, minus any hostnames already in use by the account address.  Hostnames are purged from the prior deployment and added to the new deployment's ingress in Kubernetes **only 
 when the user requests it**. This **allows for** important things: zero downtime deploys & a failed deploy just results in the old deployment continuing to serve data for that hostname.

With respect to the transfer from deployment to another, **this only happens when the user requests. This puts the onus on the customer to manage their hostnames &** gives the customer the flexibility to rollback to old deployments as well. **Creating a deployment that uses a hostname already in use always succeeds, but the hostname is only moved over whenever the user actually requests the provider do this.**

**An example of how this request is made using the command line:**

```
akash provider migrate-hostnames --dseq 1000 --from main --provider akash1uvj4vc3m7pgh5n8jezev08y96nhmgap73z7kqv akash.localhost
```

**In this case, the hostname `akash.localhost` is migrated to the deployment with sequence number 1000.**

This gives customers an easy to way to go rollback a botched update of their website. This capability is inline with what is provided by other cloud providers, although not as simple.

We need to give customers as much better way to see the status of their deployments as identified in my other PR, so I will incorporate as part of those changed another command which allows a customer to query a provider and see what hostnames are in use by the provider (for their address) and what the associated dseq is. At least for hostnames, this gives users a single place they can check the status and determine which dseq actually owns the hostname.

### DIFF
--- a/provider/cluster/hostname.go
+++ b/provider/cluster/hostname.go
@@ -221,12 +221,10 @@ type hostnameService struct {
 
 const HostnameSeparator = '.'
 
-func newHostnameService(ctx context.Context, cfg Config) *hostnameService {
-
+func newHostnameService(ctx context.Context, cfg Config, initialData map[string]dtypes.DeploymentID) *hostnameService {
 	blockedHostnames := make([]string, 0)
 	blockedDomains := make([]string, 0)
 	for _, name := range cfg.BlockedHostnames {
-
 		if len(name) != 0 && name[0] == HostnameSeparator {
 			blockedDomains = append(blockedDomains, name)
 			blockedHostnames = append(blockedHostnames, name[1:])
@@ -236,7 +234,7 @@ func newHostnameService(ctx context.Context, cfg Config) *hostnameService {
 	}
 
 	hs := &hostnameService{
-		inUse:            make(map[string]dtypes.DeploymentID),
+		inUse:            make(map[string]dtypes.DeploymentID, len(initialData)),
 		blockedHostnames: blockedHostnames,
 		blockedDomains:   blockedDomains,
 		requests:         make(chan reserveRequest),
@@ -244,6 +242,9 @@ func newHostnameService(ctx context.Context, cfg Config) *hostnameService {
 		releases:         make(chan dtypes.DeploymentID),
 		lc:               lifecycle.New(),
 		prepareRequest: make(chan prepareTransferRequest),
+	}
+	for k, v := range initialData {
+		hs.inUse[k] = v
 	}
 
 	go hs.lc.WatchContext(ctx)
--- a/provider/cluster/hostname.go
+++ b/provider/cluster/hostname.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"fmt"
 	lifecycle "github.com/boz/go-lifecycle"
-	cosmostypes "github.com/cosmos/cosmos-sdk/types"
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
 	dtypes "github.com/ovrclk/akash/x/deployment/types"
 	"github.com/pkg/errors"
 	"strings"
@@ -14,7 +14,7 @@ import (
 type HostnameServiceClient interface {
 	ReserveHostnames(hostnames []string, dID dtypes.DeploymentID) ReservationResult
 	ReleaseHostnames(dID dtypes.DeploymentID)
-	CanReserveHostnames(hostnames []string, ownerAddr cosmostypes.Address) <-chan error
+	CanReserveHostnames(hostnames []string, ownerAddr sdktypes.Address) <-chan error
 	PrepareHostnamesForTransfer(hostnames []string, dID dtypes.DeploymentID) <- chan error
 }
 
@@ -137,7 +137,7 @@ func reserveHostnamesImpl(store map[string]dtypes.DeploymentID, hostnames []stri
 	return
 }
 
-func (sh *SimpleHostnames) CanReserveHostnames(hostnames []string, ownerAddr cosmostypes.Address) <-chan error {
+func (sh *SimpleHostnames) CanReserveHostnames(hostnames []string, ownerAddr sdktypes.Address) <-chan error {
 	sh.lock.Lock()
 	defer sh.lock.Unlock()
 	ch := make(chan error, 1)
@@ -145,7 +145,7 @@ func (sh *SimpleHostnames) CanReserveHostnames(hostnames []string, ownerAddr cos
 	return ch
 }
 
-func canReserveHostnamesImpl(store map[string]dtypes.DeploymentID, hostnames []string, ownerAddr cosmostypes.Address, chErr chan<- error) {
+func canReserveHostnamesImpl(store map[string]dtypes.DeploymentID, hostnames []string, ownerAddr sdktypes.Address, chErr chan<- error) {
 	for _, hostname := range hostnames {
 		usedByDid, inUse := store[hostname]
 
@@ -196,7 +196,7 @@ type reserveRequest struct {
 type canReserveRequest struct {
 	hostnames []string
 	result    chan<- error
-	ownerAddr cosmostypes.Address
+	ownerAddr sdktypes.Address
 }
 
 type prepareTransferRequest struct {
@@ -363,7 +363,7 @@ func (hs *hostnameService) ReleaseHostnames(dID dtypes.DeploymentID) {
 	}
 }
 
-func (hs *hostnameService) CanReserveHostnames(hostnames []string, ownerAddr cosmostypes.Address) <-chan error {
+func (hs *hostnameService) CanReserveHostnames(hostnames []string, ownerAddr sdktypes.Address) <-chan error {
 	returnValue := make(chan error, 1) // Buffer of one so service does not block
 	lowercaseHostnames := make([]string, len(hostnames))
 	for i, hostname := range hostnames {
--- a/provider/cluster/hostname.go
+++ b/provider/cluster/hostname.go
@@ -73,6 +73,7 @@ func prepareHostnamesImpl(store map[string]dtypes.DeploymentID, hostnames []stri
 	for _, hostname := range toChange {
 		store[hostname] = dID
 	}
+	close(errCh)
 }
 
 func (sh *SimpleHostnames) ReserveHostnames(hostnames []string, dID dtypes.DeploymentID) ReservationResult {
@@ -242,6 +243,7 @@ func newHostnameService(ctx context.Context, cfg Config) *hostnameService {
 		canRequest:       make(chan canReserveRequest),
 		releases:         make(chan dtypes.DeploymentID),
 		lc:               lifecycle.New(),
+		prepareRequest: make(chan prepareTransferRequest),
 	}
 
 	go hs.lc.WatchContext(ctx)
--- a/provider/cluster/hostname.go
+++ b/provider/cluster/hostname.go
@@ -4,23 +4,17 @@ import (
 	"context"
 	"fmt"
 	lifecycle "github.com/boz/go-lifecycle"
+	cosmostypes "github.com/cosmos/cosmos-sdk/types"
 	dtypes "github.com/ovrclk/akash/x/deployment/types"
 	"github.com/pkg/errors"
 	"strings"
 	"sync"
 )
 
-type reserveRequest struct {
-	hostnames []string
-	result    chan<- error
-	doReserve bool
-	dID       dtypes.DeploymentID
-}
-
 type HostnameServiceClient interface {
-	ReserveHostnames(hostnames []string, did dtypes.DeploymentID) <-chan error
-	ReleaseHostnames(hostnames []string)
-	CanReserveHostnames(hostnames []string, did dtypes.DeploymentID) <-chan error
+	ReserveHostnames(hostnames []string, dID dtypes.DeploymentID) ReservationResult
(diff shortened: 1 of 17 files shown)


ngehani:
> This does two important things: zero downtime deploys & a failed deploy just results in the old deployment continuing to serve data for that hostname.

is this kinda like a blue/green deployment scenario? Only when a new/updated deployment is running, the old one goes away because the domain name is not given to the new one.

hydrogen18:
> > This does two important things: zero downtime deploys & a failed deploy just results in the old deployment continuing to serve data for that hostname.
> 
> is this kinda like a blue/green deployment scenario? Only when a new/updated deployment is running, the old one goes away because the domain name is not given to the new one.

Yes, that is exactly what I have in mind.

provider/cluster/hostname.go
```diff
@@ -4,79 +4,174 @@ import (
...
+	select {
+	case err := <-rr.ChErr:
+		if err != nil {
+			return nil, err
+		}
+	case <-wait:
+		return nil, errors.New("bob")
+	}
+
+	return <-rr.ChReplacedHostnames, nil
```
boz, line 45: this can/should be in the `select` so that you don't have to worry about closing or sending `nil` to `ChErr`.

provider/cluster/hostname.go
```diff
@@ -4,79 +4,174 @@ import (
 	"context"
 	"fmt"
 	lifecycle "github.com/boz/go-lifecycle"
+	cosmostypes "github.com/cosmos/cosmos-sdk/types"
```
boz, line 7: use `sdk` like is done everywhere else.

provider/cluster/kube/client.go
```diff
@@ -136,6 +139,93 @@ func openKubeConfig(cfgPath string, log log.Logger) (*rest.Config, error) {
 	return rest.InClusterConfig()
 }
 
+func (c *client) ClearHostname(ctx context.Context, ownerAddress cosmostypes.Address, dseq uint64, hostname string) error {
+	labelSelector := &strings.Builder{}
+	_, _ = fmt.Fprintf(labelSelector, "%s=%s", akashLeaseOwnerLabelName, ownerAddress.String())
+	_, _ = fmt.Fprintf(labelSelector, ",%s=%d", akashLeaseDSeqLabelName, dseq)
```
boz, line 145: Construct one from a map then serialize it to a string.  See [here](https://github.com/ovrclk/akash/blob/9f13194fa52ab1c2d0abec149aba284aa87735d1/provider/cluster/kube/builder.go#L536-L540) for constructing one.

boz (commented):
Some minor comments inline.  Not done reviewing.

Bigger picture, I'd prefer to have a manual switchover like is described in #1281 than an automatic one.  We don't know if the new deployment is functional, etc...

I don't want to encode these kinds of auto-migrate behaviors at this level.

hydrogen18:
@boz  I think i've addressed most of your comments (or they just went away with the refactor). I've moved everything over to be manual & user initiated for now. The description in the PR has been updated, changes & additions are in bold.

provider/cluster/hostname.go
```diff
@@ -4,92 +4,227 @@ import (
...
+}
+
+func (rr ReservationResult) Wait(wait <-chan struct{}) ([]string, error) {
+	select {
+	case err := <-rr.ChErr:
+		return nil, err
+	case v := <-rr.ChWithheldHostnames:
+		return v, nil
+	case <-wait:
+		return nil, errors.New("bob")
```
boz, line 38: ahem

provider/cluster/hostname.go
```diff
@@ -4,92 +4,227 @@ import (
...
+		store[hostname] = dID
+	}
+	close(errCh)
+}
+
+func (sh *SimpleHostnames) ReserveHostnames(hostnames []string, dID dtypes.DeploymentID) ReservationResult {
+	sh.lock.Lock()
+	defer sh.lock.Unlock()
+	errCh := make(chan error, 1)
+	resultCh := make(chan []string, 1)
```
boz, line 83: It doesn't look like any threads are used here.  Why do we need channels?

provider/cluster/hostname.go
```diff
@@ -4,92 +4,227 @@ import (
...
-	ReleaseHostnames(hostnames []string)
-	CanReserveHostnames(hostnames []string, did dtypes.DeploymentID) <-chan error
+	ReserveHostnames(hostnames []string, dID dtypes.DeploymentID) ReservationResult
+	ReleaseHostnames(dID dtypes.DeploymentID)
+	CanReserveHostnames(hostnames []string, ownerAddr sdktypes.Address) <-chan error
+	PrepareHostnamesForTransfer(hostnames []string, dID dtypes.DeploymentID) <- chan error
 }
 
 type SimpleHostnames struct {
 	Hostnames map[string]dtypes.DeploymentID
```
boz, line 22: can both of these be private?

provider/cluster/hostname.go
```diff
@@ -126,25 +266,47 @@ loop:
...
+		deploymentID: dID,
+		chErr:     chErr,
+	}
+	select {
+		case hs.prepareRequest <- v:
+	case <-hs.lc.ShuttingDown():
+		chErr <- ErrNotRunning
+	}
+
+	return chErr
```
boz, line 297: It is generally cleaner to contain asynchronous things within the implementation here, like is done with everything else.  Why are we returning a channel here?

provider/cluster/manager.go
```diff
@@ -108,47 +112,73 @@ func (dm *deploymentManager) teardown() error {
...
-			runch = dm.startDeploy()
 
 		case shutdownErr = <-dm.lc.ShutdownRequest():
 			break loop
 
+		case forceResponse := <-dm.forcech:
+			if dm.state == dsDeployComplete {
+				newch := dm.handleUpdate()
+				if newch != nil {
+					runch = newch
```
boz, line 168: Put some documentation in here about how `handleUpdate()` will only return non-`nil` in cases that `runch` is already `nil`.

Also,

```go
if newch := dm.handleUpdate(); newch != nil {
  runch = newch
}
```

boz (commented):
looks good.

did one pass, need to continue later.

Some comments inline.  I'd like to see less channel use proliferation if possible.

provider/cluster/hostname.go
hydrogen18: This type is used in tests to avoid the need to stop / start another goroutine in each test

provider/cluster/hostname.go
hydrogen18: This type is used in tests to avoid the need to stop / start another goroutine in each test, so it uses the same code path as the normal one as well.

boz: can you put these in a file name that ends in `_test.go`?

hydrogen18: its used in other packages like in the file `provider/manifest/manager_test.go`. Is it still import-able if the definition is in `_test.go` file?

boz: I believe so, but I could be wrong.

hydrogen18: if I move this type into a file ending in `_test.go` then try and run that test I get the following error

```
~/ovrclk/akash/provider/manifest$ go test -v ''
# github.com/ovrclk/akash/provider/manifest [github.com/ovrclk/akash/provider/manifest.test]
./manager_test.go:84:16: undefined: "github.com/ovrclk/akash/provider/cluster".SimpleHostnames
```

provider/cluster/service_transfer_hostnames.go
```diff
@@ -0,0 +1,286 @@
...
+	var ctx context.Context
+	ctx, xferMgr.cancel = context.WithCancel(context.Background())
+	go func() {
+		select {
+		case <- s.lc.ShuttingDown():
+			xferMgr.cancel()
+		case <- ctx.Done():
+			return
+		}
+	}()
```
boz, line 89: can you cancel all `xferHostnameManagers`s when shutting down in the main loop instead of doing this?

provider/cluster/service_transfer_hostnames.go
```diff
@@ -0,0 +1,286 @@
...
+	log log.Logger
+
+	client Client
+}
+
+var errDestintationDeploymentDoesNotExist = errors.New("destination deployment does not exist")
+var errTransferRunning = errors.New("hostname transfer already running")
+
+func (s *service) transferHostnames(req transferHostnamesRequest) error {
+	ownerAddr, err := req.destination.GetOwnerAddress()
```
boz, line 39: do this in the client thread [here](https://github.com/ovrclk/akash/pull/1299/files#diff-a55c682314e02a83146fa4a7a6c67dc87dda87d97559f5edbfdb3cae4ad14efeR219)

boz (commented):
It's really difficult to understand what is going on with the `transferHostnamesManager` workflow.

Can this be simplified at all?

Having the workers send and depend on signals from one another looks treacherous.

provider/cluster/service_transfer_hostnames.go
hydrogen18: That should work.

provider/cluster/hostname.go
hydrogen18: I will see about refactoring this

hydrogen18:
`transferHostnames` currently works on the presumption that a deployment owns an ingress in Kubernetes. I can add a timeout to the goroutines context if progress isn't happening to abort the things going on. That doesn't really solve much however.

You can't have more than one ingress with any given hostname. Each path (we only create one) goes to a single kubernetes ingress backend. A backend uses a kubernetes service selector. If the concept of a Ingress controller is instead owned by a deployment owner, we could update the service selector to have match clause that targets whatever pod we want. This alone doesn't solve much, as you still need to coordinate the removal & addition of the match clauses between multiple deployments, avoiding having any one deployment overwrite the others changes. The only thing this would do is possibly eliminate the brief period of time in which the ingress controller is currently returning 404 for a hostname when one in transferred over.

It is plausible to go around and group up the individual kubernetes operations (removing an ingress with a hostname, then creating another with the same hostname) and apply them all at once from a single goroutine. This brings in its own set of problems, because then the deployment managers aren't really responsible for managing the deployments anymore. You'd have to block off deployment managers from making any changes during that time period. This moves the problem from one place to two.

It is possible to define a kubernetes ingress with multiple paths sharing the same hostname. This is a feature of the NGINX ingress controller: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#canary . This doesn't solve much as the transfer & coordination problem still exists.

The only solution I can think of that comes close to simplifying things is

1. Move hostname to a concept owned by a deployment owner, not the deployment.
2. Create all ingresses in a namespace unique to a deployment owner, not the deployment.
3. Any time a hostname should target a given deployment, update the service selector  on the backend.
4. Use a goroutine per deployment owner to make changes in step 3, avoiding any chance of changes getting clobbered
5. Any time a deployment starts with a hostname _that is not in use_ automatically make the update in step 3 for the user
6. Any time a deployment starts with ah hostname _that is in use_  do nothing.
7. Any time the user sends a request to "migrate" a hostname just perform step 3 
8. As deployments are closed for an owner, check to see if a service selector in an ingress now targets nothing. If it is, delete the associated hostname.

The SDL now specifies hostnames in the same manner, but listing a hostname only indicates that a deployment could **possibly** serve that hostname. Not that it is a requirement.

This approaches requires that somehow we 'upgrade' existing deployments to this strategy at the time that we update the provider. I haven't actually done this yet, but it should be possible. We'd need to plan that out and test it extensively.

Note Step 2 requires a change to our allowlist in the network security of each pod, but that isn't very difficult. 

In Step 7, when the user requests a migrate we don't care about the deployments status. It's up to the end user to make sure they are switching to a deployment that is functional.

hydrogen18:
All feedback seems pertinent at this time, but I am continuing to investigate refactoring this into a operator - based scheme for controlling the provisioning of hostnames & ingress in Kubernetes. I will update this PR again with another comment when the results of that are complete.

PR #1322, opened 2021-07-14
fix(provider): correct deployment failure on multiple service exposure

hydrogen18:
After looking at what happened with @tombeynon 's deployment, the provider tried to tell kubernetes to use the same port twice on a service (which is basically just an IP address inside Kubernetes). I think whenever I refactored some of the manifest code, I made an error and had it create multiple definitions for the same service. Due to the way we deploy into kubernetes, each service can always see any other service for the same deployment group running in kubernetes. This works because of how our network security policies are applied to the deployments in the namespace. 

The `to` field is mostly an annotation in other words. If we ever get a cross-provider VPN working then it would probably actually be something we can use to determine what needs to be cross-connected. 

I just changed the code not to create multiple definitions for the same port and to validate that the requested service is actually accessible.

### DIFF
--- a/manifest/types.go
+++ b/manifest/types.go
@@ -73,7 +73,6 @@ type ServiceExpose struct {
 	Port         uint16 // Port on the container
 	ExternalPort uint16 // Port on the service definition
 	Proto        ServiceProtocol
-	Service      string
 	Global       bool
 	Hosts        []string
 }
--- a/pkg/apis/akash.network/v1/types.go
+++ b/pkg/apis/akash.network/v1/types.go
@@ -270,7 +270,6 @@ func (mse ManifestServiceExpose) toAkash() (manifest.ServiceExpose, error) {
 		Port:         mse.Port,
 		ExternalPort: mse.ExternalPort,
 		Proto:        proto,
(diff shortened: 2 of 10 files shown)


hydrogen18:
I've got to double check the code on this, but it looks like that we're probably over counting the number of endpoints when we build the data for the deployment groups that goes on chain.

boz (commented):
I think we'd need to migrate the existing stored manifests to remove these fields.

How would it affect current deployments, services, ingresses?  Would it affect their names?

I'd prefer to not lose information.  I'm not sure why it's an issue to keep this.

hydrogen18:
> I think we'd need to migrate the existing stored manifests to remove these fields.
> 
> How would it affect current deployments, services, ingresses? Would it affect their names?
> 
> I'd prefer to not lose information. I'm not sure why it's an issue to keep this.

What would be preserving? I'm not entirely sure what we'd be losing. You can't create a functional deployment with the manifest that Tom supplied presently. Therefore, none exist.

hydrogen18:
Rolling this into my current work on another branch

PR #1323, opened 2021-07-15
Ingress configuration options

hydrogen18:
This gives users the options in their SDL to configure

1. read timeout
2. send timeout
3. body size limits
4. number of retries (for when replicas > 1)
5. retry cases
6. retry timeout

These become annotations that control our kubernetes NGINX ingress controller. There are some limits on these values, for example I have it so body size can't be more than 50 MB right now.

If this isn't specified in the SDL the defaults are the same as what is in effect today on everyone's ingress controller.

### DIFF
--- a/provider/cluster/kube/builder.go
+++ b/provider/cluster/kube/builder.go
@@ -944,11 +944,56 @@ func ingressHost(lid mtypes.LeaseID, svc *manifest.Service) string {
 	return strings.ToLower(base32.HexEncoding.WithPadding(base32.NoPadding).EncodeToString(uid))
 }
 
+func (b *ingressBuilder) annotations() map[string]string {
+	// For kubernetes/ingress-nginx
+	// https://github.com/kubernetes/ingress-nginx
+	const root = "nginx.ingress.kubernetes.io"
+
+	httpOptions := b.expose.HTTPOptions
+
+	result := map[string]string{
+
+		fmt.Sprintf("%s/proxy-send-timeout", root):        fmt.Sprintf("%dms", httpOptions.SendTimeout),
+		fmt.Sprintf("%s/proxy-read-timeout", root):        fmt.Sprintf("%dms", httpOptions.ReadTimeout),
+		fmt.Sprintf("%s/proxy-next-upstream-tries", root): strconv.Itoa(int(httpOptions.NextTries)),
+		fmt.Sprintf("%s/proxy-body-size", root):           strconv.Itoa(int(httpOptions.MaxBodySize)),
+	}
+
+	nextTimeoutKey := fmt.Sprintf("%s/proxy-next-upstream-timeout", root)
+	if httpOptions.NextTimeout > 0 {
+		result[nextTimeoutKey] = fmt.Sprintf("%dms", httpOptions.NextTimeout)
+	} else {
+		result[nextTimeoutKey] = "0" // Magic value for disable
+	}
+
+	builder := strings.Builder{}
+
+	for i, v := range httpOptions.NextCases {
+		first := string(v[0])
+		isHTTPCode := strings.ContainsAny(first, "12345")
+
+		if isHTTPCode {
+			builder.WriteString("http_")
+		}
+		builder.WriteString(v)
+
+		if i != len(httpOptions.NextCases)-1 {
+			// The actual separator is the space character for kubernetes/ingress-nginx
+			builder.WriteRune(' ')
+		}
+	}
+
+	result[fmt.Sprintf("%s/next-upstream", root)] = builder.String()
+
+	return result
+}
+
 func (b *ingressBuilder) create() (*netv1.Ingress, error) { // nolint:golint,unparam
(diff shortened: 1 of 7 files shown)


dmikey (approved):
I would be interested to see perf testing against the ingress controller with different body loads. My overall question is how large and how many simultaneous bodies can be uploaded before degraded perf across the controller. I'm assuming this controller is one per cluster, and a provider is responsible for that cluster configuration and performance. How do we measure that for quality of service?

hydrogen18:
The NGINX controller is an instance set at present, so it is indeed one per cluster. We can move to a daemonset in the future if the need arises to get more performance.

At present I think our actual bottleneck is the internal networking of the cluster, we probably should spend more time to make sure we're getting the most out of the infrastructure we have with our networking setup.

sdl/v2.go
```diff
@@ -24,11 +52,92 @@ type v2ExposeTo struct {
...
+		nextCases = defaultNextCases
+	} else {
+		for _, nextCase := range nextCases {
+			switch nextCase {
+			case nextCaseOff:
+				if len(nextCases) != 1 {
+					return manifest.ServiceExposeHTTPOptions{}, errCannotSpecifyOffAndOtherCases
+				}
+			case nextCaseError:
+			case nextCaseTimeout:
```
troian, line 119: missing `fallthrough` ?

troian (changes requested):
some tests are failing as well

```
--- FAIL: TestV2HTTPOptionsParse (0.00s)
    v2_test.go:337: 
        	Error Trace:	v2_test.go:337
        	Error:      	Received unexpected error:
        	            	open _testdata/simple_httpoptions.yaml: no such file or directory
        	Test:       	TestV2HTTPOptionsParse
```

sdl/v2.go
hydrogen18: It should take no action in the event that this case is hit, which is the correct thing

0x0I (approved):

provider/cluster/kube/builder.go
```diff
@@ -944,11 +944,56 @@ func ingressHost(lid mtypes.LeaseID, svc *manifest.Service) string {
...
+	// For kubernetes/ingress-nginx
+	// https://github.com/kubernetes/ingress-nginx
+	const root = "nginx.ingress.kubernetes.io"
+
+	httpOptions := b.expose.HTTPOptions
+
+	result := map[string]string{
+
+		fmt.Sprintf("%s/proxy-send-timeout", root):        fmt.Sprintf("%dms", httpOptions.SendTimeout),
+		fmt.Sprintf("%s/proxy-read-timeout", root):        fmt.Sprintf("%dms", httpOptions.ReadTimeout),
```
boz, line 957: should we just have them input it as a string `"100ms"`?

I wonder what will happen if we have invalid values here.

sdl/v2.go
```diff
@@ -24,11 +52,92 @@ type v2ExposeTo struct {
...
+		ho.ReadTimeout != 0 ||
+		ho.SendTimeout != 0 ||
+		ho.NextTries != 0 ||
+		ho.NextTimeout != 0
+}
+
+func (ho v2HTTPOptions) asManifest() (manifest.ServiceExposeHTTPOptions, error) {
+	maxBodySize := ho.MaxBodySize
+	if 0 == maxBodySize {
+		maxBodySize = defaultMaxBodySize
```
boz, line 84: these defaults would ideally be configurable

boz (commented):
Looks great!

Will this be compatible with current, saved manifests?

Should we put a limit on some of these values?

boz (commented):
I see the limits now.

This is super cool

sdl/v2.go
hydrogen18: On chain or per provider ?

provider/cluster/kube/builder.go
hydrogen18: The input type is `uint32` so they can't really input anything invalid. If they input zero we use the default. A positive number is a valid value, up to the limit.

hydrogen18:
> Looks great!
> 
> Will this be compatible with current, saved manifests?
> 
> Should we put a limit on some of these values?

It is compatible with saved manifest because it is additive only. Omitting this section results in values that are the same as the defaults today

sdl/v2.go
boz: I think per provider.

forget it, this is fine.

troian:
lgtm. one more little think. there are non-gofmt file(s) https://github.com/ovrclk/akash/pull/1323/checks?check_run_id=3085774438

jay-dee7:
@hydrogen18 @boz @dmikey guys this PR looks great, is there any blocker for this PR to get merged?

hydrogen18:
@jay-dee7 This refactor is going to get merged into some other work I am doing to overhaul how we handle Kubernetes ingress objects and the assignment of hostnames to a users deployment on Akash

Issue #1243, opened 2021-05-04
Provider inventory for endpoints is wrong


hydrogen18:
For some reason the provider never really updates its count of available endpoints. The total # available always decreases. Restarting the provider gets the real number back

![Screenshot_2021-05-04_10-23-50](https://user-images.githubusercontent.com/130230/117049117-635f4200-acd9-11eb-8d0f-c1aeea2494da.png)

hydrogen18:
@arijitAD  this is related to #1317 . Endpoints are part of what we consider inventory.

Issue #1317, opened 2021-07-09
Available inventory still seems to get corrupted over time


hydrogen18:
The provider still gets stuck believing that it doesn't have enough capacity for new leases, even when some kubernetes nodes are basically idle. 

There must be at least one code path that somehow messes up the computation and causes the inventory code to believe that it does not have any available capacity. Restarting the provider seems to eventually resolve the issue.

arijitAD:
Fixed by #1337.

PR #1337, opened 2021-07-27
Restore inventory on cancel reservation.

arijitAD:
## Description

Fixes https://github.com/ovrclk/akash/issues/1243 and https://github.com/ovrclk/akash/issues/1317

## Motivation and Context

## How Has This Been Tested?

```
Test_OrderCloseBeforeReserveReturn
```

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/provider/bidengine/order.go
+++ b/provider/bidengine/order.go
@@ -3,9 +3,6 @@ package bidengine
 import (
 	"context"
 	"fmt"
-	atypes "github.com/ovrclk/akash/x/audit/types"
-	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promauto"
 	"regexp"
 	"time"
 
@@ -18,8 +15,11 @@ import (
 	"github.com/ovrclk/akash/pubsub"
 	metricsutils "github.com/ovrclk/akash/util/metrics"
 	"github.com/ovrclk/akash/util/runner"
+	atypes "github.com/ovrclk/akash/x/audit/types"
 	dtypes "github.com/ovrclk/akash/x/deployment/types"
 	mtypes "github.com/ovrclk/akash/x/market/types"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/tendermint/tendermint/libs/log"
 )
 
@@ -360,6 +360,13 @@ loop:
 
 	// cancel reservation
 	if !won {
+		if clusterch != nil {
+			result := <-clusterch
+			clusterch = nil
+			if result.Error() == nil {
+				reservation = result.Value().(ctypes.Reservation)
+			}
+		}
 		if reservation != nil {
 			o.log.Debug("unreserving reservation")
 			if err := o.cluster.Unreserve(reservation.OrderID()); err != nil {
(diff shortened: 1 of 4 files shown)


provider/bidengine/order.go
```diff
@@ -360,6 +360,13 @@ loop:
 
 	// cancel reservation
 	if !won {
+		if clusterch != nil {
+			result := <-clusterch
+			clusterch = nil
+			if result.Error() == nil {
+				reservation = result.Value().(ctypes.Reservation)
+			}
+		}
```
boz, line 369: good find.  should this be done after `cancel()`?

provider/cluster/inventory.go
```diff
@@ -450,6 +448,10 @@ loop:
 				is.log.Info("removing reservation", "order", res.OrderID())
 
 				reservations = append(reservations[:idx], reservations[idx+1:]...)
+				// reclaim availableExternalPorts if unreserving allocated resources
+				if res.allocated {
+					is.availableExternalPorts += reservationCountEndpoints(res)
```
boz, line 453: looks reasonable, but I defer to @hydrogen18 on this

boz (commented):
Great finds, @arijitAD

provider/bidengine/order.go
hydrogen18: The code was arranged like that at one point and caused all sorts of weird errors. We'd cancel the context then wait on the results and log a bunch of errors (instead of things being cleaned up)

provider/bidengine/order.go
```diff
@@ -360,6 +360,13 @@ loop:
 
 	// cancel reservation
 	if !won {
+		if clusterch != nil {
+			result := <-clusterch
```
hydrogen18, line 364: This makes sense but we also need to avoid blocking forever here. The simplest solution may just be to use `time.After` with a long timeout. @boz thoughts?

provider/cluster/inventory.go
hydrogen18: Makes sense to me, I just missed the case. When the reservation is has `allocated = true` the code does `is.availableExternalPorts -= externalPortCount` so this restores the previous value when the reservation is released _if it has been allocated_.

It looks like @arijitAD added a test for this so :+1: :100:

hydrogen18 (approved):

provider/bidengine/order.go
arijitAD: Can there be a case where this channel will block infinitely? I looked into `func (is *inventoryService) reserve` code and this always seems to return data to the channel.

boz: We try to be really careful about this stuff.  Since we're not cancelling the context I worry a bit, but if it's a clear case then okay, forget it.

Alternatively we can send in context that has a bound duration to begin with.

hydrogen18: @boz if the inventory service is shutting down it won't answer any more requests right? So wouldn't this goroutine get stuck "forever" when stopping the provider?

boz: It *should* respond with an error when it's shutting down.  Let me check.

boz: Does this cover the case you're talking about @hydrogen18? https://github.com/ovrclk/akash/blob/5df45f7b009e04b95dbdf430bf5f149b9cab14c7/provider/cluster/inventory.go#L155-L157

PR #1341, opened 2021-07-28
Match provider config on order bid.

arijitAD:
## Description

Fixes https://github.com/ovrclk/akash/issues/715

## Motivation and Context

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/provider/bidengine/order.go
+++ b/provider/bidengine/order.go
@@ -411,7 +411,13 @@ func (o *order) shouldBid(group *dtypes.Group) (bool, error) {
 
 	// does provider have required attributes?
 	if !group.GroupSpec.MatchAttributes(o.session.Provider().Attributes) {
-		o.log.Debug("unable to fulfill: incompatible attributes")
+		o.log.Debug("unable to fulfill: incompatible provider attributes")
+		return false, nil
+	}
+
+	// does order have required attributes?
+	if !o.cfg.Attributes.SubsetOf(group.GroupSpec.Requirements.Attributes) {
+		o.log.Debug("unable to fulfill: incompatible order attributes")
 		return false, nil
 	}
 
--- a/provider/cmd/run.go
+++ b/provider/cmd/run.go
@@ -7,12 +7,14 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
-	mparams "github.com/ovrclk/akash/x/market/types"
-	"github.com/shopspring/decimal"
 	"io"
 	"net/http"
 	"os"
 	"time"
+
+	mparams "github.com/ovrclk/akash/x/market/types"
+	config2 "github.com/ovrclk/akash/x/provider/config"
+	"github.com/shopspring/decimal"
 
 	"github.com/pkg/errors"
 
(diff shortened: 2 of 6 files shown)


provider/bidengine/order.go
```diff
@@ -404,7 +404,13 @@ func (o *order) shouldBid(group *dtypes.Group) (bool, error) {
 
 	// does provider have required attributes?
 	if !group.GroupSpec.MatchAttributes(o.session.Provider().Attributes) {
-		o.log.Debug("unable to fulfill: incompatible attributes")
+		o.log.Debug("unable to fulfill: incompatible provider attributes")
+		return false, nil
+	}
+
+	// does order have required attributes?
+	if !o.cfg.Attributes.SubsetOf(group.GroupSpec.Requirements.Attributes) {
```
dmikey, line 412: What's the difference between the session and the cfg ?

provider/bidengine/order.go
```diff
@@ -404,7 +404,13 @@ func (o *order) shouldBid(group *dtypes.Group) (bool, error) {
 
 	// does provider have required attributes?
 	if !group.GroupSpec.MatchAttributes(o.session.Provider().Attributes) {
-		o.log.Debug("unable to fulfill: incompatible attributes")
+		o.log.Debug("unable to fulfill: incompatible provider attributes")
+		return false, nil
```
boz, line 408: woah, good catch.

provider/bidengine/order.go
boz, line 412: Does `MatchAttributes` just call `SubsetOf`?  Let's use the exact same matching method if not.

boz (approved):
Looks great!  A couple comments inline.

provider/bidengine/order.go
hydrogen18: I'm unsure why would need this additional check, if there is a bug its probably in `MatchAttributes` and should be fixed there

arijitAD: `session` gives the attributes stored on-chain when the provider was created with `$akash tx provider create`. If an order specifies some required attributes, then the session's provider attributes must have those attributes, otherwise it can't bid on the order.

On the other hand, `cfg` gives the attributes that the order must have for the provider to bid on it. These are specified using `$akash provider run --provider-config <path-to-conf-file>`.

arijitAD: Both `MatchAttributes` and `SubsetOf` internally just call `AttributesSubsetOf`. So, it seems fine to keep it as it is.

hydrogen18: I don't actually see this `--provider-config` flag, was this added recently?

arijitAD: Yes. It's added in this PR.

provider/cmd/run.go
```diff
@@ -277,6 +280,11 @@ func RunCmd() *cobra.Command {
 		return nil
 	}
 
+	cmd.Flags().String(FlagProviderConfig, "", "provider configuration file path")
```
hydrogen18, line 283: This flag should probably be optional. Otherwise, existing providers will need to create a file that lists no attributes in order to continue to get the same behavior today.

arijitAD: Done

hydrogen18 (approved):

PR #1343, opened 2021-07-29
Upgrade cosmos sdk to latest release.

arijitAD:
## Motivation and Context

## Types of changes

## Checklist:


hydrogen18:
I think we may have to backport all the bug fixes I made into the "latest" version of cosmos SDK again

arijitAD:
# Akash: cosmos lib 0.43.0-rc2 upgrade
* [x] Use new ibc repo: https://github.com/cosmos/ibc-go/blob/v1.0.0-rc3/docs/migrations/ibc-migration-043.md
* [x] cosmos-sdk v0.43 changelog: https://github.com/cosmos/cosmos-sdk/blob/release/v0.43.x/CHANGELOG.md
* [x] Some easy renaming changes (`Marshaler` -> `Codec` ones).
* [x] `sdkutil/broadcast.go`: Private func `PrepareFactory` had to be copied from sdk out to Akash code :(
* [x] `BaseApp.SetAppVersion()` -> `BaseApp.SetVersion()`. See: https://github.com/cosmos/cosmos-sdk/pull/8897
* [x] `AppModule` now has a new `ConsensusVersion()` func that needs to be implemented by every akash module. Returning 1 by default for akash modules.
* [x] Simulations for every module require a new arg `cdc *codec.ProtoCodec` in `simtypes.NewOperationMsg()`. Setting that to `nil` as all the messages are `LegacyMsg` which don't seem require the codec based on the implementation code.
* [x] `ibcclient.NewClientUpdateProposalHandler()` -> `ibcclient.NewClientProposalHandler`. See: https://github.com/cosmos/cosmos-sdk/pull/8673
* [ ] Need to correctly set `x/audit/keeper/key.go:keyAddrPrefixLen` which depended on the `sdk.AddrLen` which was fixed 20 bytes previously, but can be variable length now. See: https://github.com/cosmos/cosmos-sdk/pull/8363
* [ ] Need to handle upgrades properly (in-place, chain, module upgrades). See: https://docs.cosmos.network/v0.43/migrations/chain-upgrade-guide-043.html#

hydrogen18:
For tendermint, the only changes I made are the following 

```
diff --git a/rpc/client/http/http.go b/rpc/client/http/http.go
index 10fada8a6b..a825d9adec 100644
--- a/rpc/client/http/http.go
+++ b/rpc/client/http/http.go
@@ -737,15 +737,7 @@ func (w *WSEvents) eventListener() {
 
                        w.mtx.RLock()
                        if out, ok := w.subscriptions[result.Query]; ok {
-                               if cap(out) == 0 {
-                                       out <- *result
-                               } else {
-                                       select {
-                                       case out <- *result:
-                                       default:
-                                               w.Logger.Error("wanted to publish ResultEvent, but out channel is full", "result", result, "query", result.Query)
-                                       }
-                               }
+                               out <- *result
                        }
                        w.mtx.RUnlock()
                case <-w.Quit():
diff --git a/rpc/jsonrpc/client/ws_client.go b/rpc/jsonrpc/client/ws_client.go
index 1c7ade6578..d0cfd33a36 100644
--- a/rpc/jsonrpc/client/ws_client.go
+++ b/rpc/jsonrpc/client/ws_client.go
@@ -20,9 +20,9 @@ import (
 
 const (
        defaultMaxReconnectAttempts = 25
-       defaultWriteWait            = 0
-       defaultReadWait             = 0
-       defaultPingPeriod           = 0
+       defaultWriteWait            = 15 * time.Second
+       defaultReadWait             = 31 * time.Second
+       defaultPingPeriod           = 15 * time.Second
 )
 
 // WSClient is a JSON-RPC client, which uses WebSocket for communication with
```

The changes have not been made in the tendermint version you are targeting and needs to be backported by forking that tag and applying them in our repo.

hydrogen18:
For cosmos SDK, we made the following changes in our tag v0.41.4-akash-4

```
diff --git a/x/auth/legacy/v040/migrate.go b/x/auth/legacy/v040/migrate.go
index 363ec7ba82..f819a1a33d 100644
--- a/x/auth/legacy/v040/migrate.go
+++ b/x/auth/legacy/v040/migrate.go
@@ -2,6 +2,7 @@ package v040
 
 import (
        codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+       multisigtypes "github.com/cosmos/cosmos-sdk/crypto/keys/multisig"
        sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
        v039auth "github.com/cosmos/cosmos-sdk/x/auth/legacy/v039"
        v040auth "github.com/cosmos/cosmos-sdk/x/auth/types"
@@ -11,9 +12,16 @@ import (
 // convertBaseAccount converts a 0.39 BaseAccount to a 0.40 BaseAccount.
 func convertBaseAccount(old *v039auth.BaseAccount) *v040auth.BaseAccount {
        var any *codectypes.Any
-       // If the old genesis had a pubkey, we pack it inside an Any. Or else, we
-       // just leave it nil.
-       if old.PubKey != nil {
+
+       _, ok := old.PubKey.(*multisigtypes.LegacyAminoPubKey)
+
+       // If pubkey is multisig type, then leave it as nil for now
+       // Ref: https://github.com/cosmos/cosmos-sdk/issues/8776#issuecomment-790552126
+       // Else if the old genesis had a pubkey, we pack it inside an Any.
+       // Or else, we just leave it nil.
+       if ok {
+               // TODO migrate multisig public_keys
+       } else if old.PubKey != nil {
                var err error
                any, err = codectypes.NewAnyWithValue(old.PubKey)
                if err != nil {
```

This change needs to be backported

```
diff --git a/x/bank/keeper/keeper.go b/x/bank/keeper/keeper.go
index 1a7d435841..6a1b7fc8e7 100644
--- a/x/bank/keeper/keeper.go
+++ b/x/bank/keeper/keeper.go
@@ -236,7 +236,8 @@ func (k BaseKeeper) SetDenomMetaData(ctx sdk.Context, denomMetaData types.Metada
 }
 
 // SendCoinsFromModuleToAccount transfers coins from a ModuleAccount to an AccAddress.
-// It will panic if the module account does not exist.
+// It will panic if the module account does not exist. An error is returned if
+// the recipient address is black-listed or if sending the tokens fails.
 func (k BaseKeeper) SendCoinsFromModuleToAccount(
        ctx sdk.Context, senderModule string, recipientAddr sdk.AccAddress, amt sdk.Coins,
 ) error {
@@ -246,6 +247,10 @@ func (k BaseKeeper) SendCoinsFromModuleToAccount(
                panic(sdkerrors.Wrapf(sdkerrors.ErrUnknownAddress, "module account %s does not exist", senderModule))
        }
 
+       if k.BlockedAddr(recipientAddr) {
+               return sdkerrors.Wrapf(sdkerrors.ErrUnauthorized, "%s is not allowed to receive funds", recipientAddr)
+       }
+
        return k.SendCoins(ctx, senderAddr, recipientAddr, amt)
 }
 
@@ -387,6 +392,7 @@ func (k BaseKeeper) trackDelegation(ctx sdk.Context, addr sdk.AccAddress, blockT
        if ok {
                // TODO: return error on account.TrackDelegation
                vacc.TrackDelegation(blockTime, balance, amt)
+               k.ak.SetAccount(ctx, acc)
        }
 
        return nil
@@ -402,6 +408,7 @@ func (k BaseKeeper) trackUndelegation(ctx sdk.Context, addr sdk.AccAddress, amt
        if ok {
                // TODO: return error on account.TrackUndelegation
                vacc.TrackUndelegation(amt)
+               k.ak.SetAccount(ctx, acc)
        }
 
        return nil
```

This change has been partially applied in the new version. The changes to the functions `trackDelegation` & `trackUndelegation` need to be applied.

```
diff --git a/x/evidence/types/evidence.go b/x/evidence/types/evidence.go
index 95705cc5b9..fca6126ec3 100644
--- a/x/evidence/types/evidence.go
+++ b/x/evidence/types/evidence.go
@@ -88,7 +88,8 @@ func (e Equivocation) GetTotalPower() int64 { return 0 }
 // FromABCIEvidence converts a Tendermint concrete Evidence type to
 // SDK Evidence using Equivocation as the concrete type.
 func FromABCIEvidence(e abci.Evidence) exported.Evidence {
-       consAddr, err := sdk.Bech32ifyAddressBytes(sdk.Bech32PrefixConsAddr, e.Validator.Address)
+       bech32PrefixConsAddr := sdk.GetConfig().GetBech32ConsensusAddrPrefix()
+       consAddr, err := sdk.Bech32ifyAddressBytes(bech32PrefixConsAddr, e.Validator.Address)
        if err != nil {
                panic(err)
        }
```

This change has been applied and does not need to be backported

boz:
* `x/auth/legacy/v040/migrate.go`: we might as well migrate it, but it's not used anymore (we're past that migration)
* `x/bank/keeper/keeper.go`: IIRC they've done some further refactoring, and this was fixed in another way (cosmos hub was upgraded for this security fix).

arijitAD:
```
const (
	keyAddrPrefixLen = 1 + 20 // TODO: discuss and find a proper fix for this
)
```
https://github.com/ovrclk/akash/pull/1343/files#diff-d2cc399939acd1af527a3fed7af83beab3993932f4f6e12cd44c307e6cba014eR13

We need to correctly set `x/cert/keeper/key.go:keyAddrPrefixLen` which depended on the `sdk.AddrLen` which was fixed to 20 bytes previously, but can be variable length now (20 or 32).

We need to write migration for address if we are storing the address in a module similar to this code. https://github.com/cosmos/cosmos-sdk/blob/0ab5726a7aa5c2d66916b14becb696e6e19a1efb/x/bank/legacy/v043/store.go#L51

I suspect `cert` module might be storing the address. https://github.com/ovrclk/akash/blob/04e7a7667dd94b5a15fa039e4f7df5c9ad93be4f/x/cert/keeper/key.go#L42 
Can someone verify? Also, if there are other modules using address?

Ref:
* https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-028-public-key-addresses.md
* https://github.com/cosmos/cosmos-sdk/issues/8041
* https://github.com/cosmos/cosmos-sdk/pull/8363

troian:
@arijitAD yes, `cert` modules uses `address` and we need migration there.
it seems we need to add extra byte after prefix to identify address length (unless there is a univocal way identifying address length by first 20bytes) and migration should set it as 20bytes key for all existing certificates.

troian:
@arijitAD does new address format have prefix defining address length in it?

arijitAD:
> @arijitAD does new address format have prefix defining address length in it?

From this migration, it seems that 1-byte prefix is allocated for length. https://github.com/cosmos/cosmos-sdk/blob/0ab5726a7aa5c2d66916b14becb696e6e19a1efb/x/bank/legacy/v043/store.go#L55

troian:
i see, then we have to set it same way by adding extra byte for address type

boz:
> @arijitAD yes, `cert` modules uses `address` and we need migration there.
> it seems we need to add extra byte after prefix to identify address length (unless there is a univocal way identifying address length by first 20bytes) and migration should set it as 20bytes key for all existing certificates.

Let's use type instead of length for the prefix if possible.

Also, if we're going to go and rewrite our indexes, we should use the binary/packed address instead of the base 16 ascii representation representation.

troian:
> > @arijitAD yes, `cert` modules uses `address` and we need migration there.
> > it seems we need to add extra byte after prefix to identify address length (unless there is a univocal way identifying address length by first 20bytes) and migration should set it as 20bytes key for all existing certificates.
> 
> Let's use type instead of length for the prefix if possible.
> 
> Also, if we're going to go and rewrite our indexes, we should use the binary/packed address instead of the base 16 ascii representation representation.

yah, sorry, i meant to say extra byte as address type but was also curious if new address has any sort of meta information in it to determine it's length. it's not a case anyways and extra byte for type is only option

x/cert/keeper/keeper.go
```diff
@@ -59,18 +59,13 @@ func (k keeper) CreateCertificate(ctx sdk.Context, owner sdk.Address, crt []byte
 		return types.ErrCertificateExists
 	}
 
-	iter := sdk.KVStorePrefixIterator(store, certificatePrefix(owner))
-	defer func() {
-		_ = iter.Close()
-	}()
```
arijitAD, line 65: The `iter` was declared but wasn't used. Seemed unnecessary, so removed.

troian: 🙏

sdkutil/address.go
```diff
@@ -0,0 +1,10 @@
+package sdkutil
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+// GetAccAddressFromBech32 creates an AccAddress from a Bech32 string.
+// It internally calls `sdk.AccAddressFromBech32` and ignores the error.
+func GetAccAddressFromBech32(address string) sdk.AccAddress {
+	addr, _ := sdk.AccAddressFromBech32(address)
```
troian, line 8: should we set here panic if address is invalid?

script/protoc-swagger-gen.sh
```diff
@@ -21,7 +21,7 @@ done
 # combine swagger files
 # uses nodejs package `swagger-combine`.
 # all the individual swagger files need to be configured in `config.json` for merging
-swagger-combine ./client/docs/config.json \
+.cache/node_modules/.bin/swagger-combine ./client/docs/config.json \
```
troian, line 24: it should pick swagger-combine from PATH as we manually set it. doesn't it work?

x/cert/keeper/key.go
```diff
@@ -5,21 +5,20 @@ import (
...
 	"github.com/ovrclk/akash/x/cert/types"
 )
 
 const (
-	keyAddrPrefixLen = 1 + sdk.AddrLen
-)
-
-var (
-	prefixCertificateID = []byte{0x01}
+	keyAddrPrefixLen = 1 /*len(PrefixCertificateID)*/ + 1 /*owner_address_len (1 byte)*/
```
troian, line 14: something like that to make it more readable?
```
keyAddrPrefixLen = 2 // 1 byte for PrefixCertificateID) followed by owner_address_len byte
```

sdkutil/address.go
arijitAD: I had done that initially, but, that causes `TestCreateBidNonExistingOrder` to fail here: https://github.com/ovrclk/akash/blob/cosmos-upgrade/x/market/handler/handler_test.go#L134 because it passes an empty string for order owner and provider address.

I guess, we can set those strings to a valid address that doesn't exist in the store to make the test pass.

script/protoc-swagger-gen.sh
arijitAD: simply running `make codegen` wasn't working locally for me without this change.

boz (commented):
Looks great @arijitAD.

For the prefixes, I would really like to use functions that return new objects than to share mutable state if it is possible.

I believe that this will affect querying - possibly searching or desearializing indexes and data.  Take a look at [these](https://github.com/ovrclk/akash/blob/be9e181635076e0454ea1ce46266c1f16e1f8605/x/deployment/keeper/grpc_query.go) queriers in the modules to be sure.

x/cert/keeper/key.go
arijitAD: It is `const`. The diff shows `var` on the LHS, it is already `const` on the RHS.

util/legacy/v013/helpers.go
```diff
@@ -0,0 +1,37 @@
...
+
+	for ; oldStoreIter.Valid(); oldStoreIter.Next() {
+		bech32Addr := string(oldStoreIter.Key()[:V012Bech32AddrLen])
+		addr, err := types.AccAddressFromBech32(bech32Addr)
+		if err != nil {
+			panic(err)
+		}
+
+		endBz := oldStoreIter.Key()[V012Bech32AddrLen:]
+		newStoreKey := append(append(prefixBz, address.MustLengthPrefix(addr)...), endBz...)
```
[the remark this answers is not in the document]
arijitAD: Yeah! they call these variable-length addresses. So, using length is a generic way to encode and be future proof if some new address of length say 40 gets introduced. The max length they have set for an address is 255, so well within a byte.

x/audit/types/key.go
```diff
@@ -10,3 +10,7 @@ const (
 	// RouterKey is the message route for provider
 	RouterKey = ModuleName
 )
+
+var (
+	PrefixProviderID = []byte{0x01}
```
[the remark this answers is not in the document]
arijitAD: Done

sdkutil/address.go
arijitAD: Done

app/app.go
```diff
@@ -454,10 +452,46 @@ func NewApp(
 	return app
 }
 
+func (app *AkashApp) registerUpgradeHandlers() {
+	app.keeper.upgrade.SetUpgradeHandler("v0.43", func(ctx sdk.Context, plan upgradetypes.Plan,
```
arijitAD, line 456: Please suggest a suitable name for the upgrade handler. Currently, it is named `v0.43` which is based on the cosmos version we are upgrading to. Should it be something like: `v0.13` as that will be the version of the next akash  release?
Or should it be something like: `akashnet-2-upgrade-1` as it was previously?

app/app.go
```diff
@@ -350,9 +350,15 @@ func NewApp(
 		}, app.akashAppModules()...)...,
 	)
 
+	// During begin block slashing happens after distr.BeginBlocker so that
+	// there is nothing left over in the validator fee pool, so as to keep the
+	// CanWithdrawInvariant invariant.
+	// NOTE: staking module is required if HistoricalEntries param > 0
+	// NOTE: capability module's beginblocker must come before any modules using capabilities (e.g. IBC)
```
hydrogen18, line 357: Did you find this information somewhere in the Cosmos docs or just figure it out from reading the code?

arijitAD: Took it from [cosmos sdk code](https://github.com/cosmos/cosmos-sdk/blob/1dba6735739e9b4556267339f0b67eaec9c609ef/simapp/app.go#L325-L329)

sdkutil/address.go
```diff
@@ -0,0 +1,13 @@
+package sdkutil
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+// MustAccAddressFromBech32 creates an AccAddress from a Bech32 string.
+// It internally calls `sdk.AccAddressFromBech32` and ignores the error.
```
hydrogen18, line 6: I think this comment is wrong? It panics if there is an error.

sdkutil/broadcast.go
```diff
@@ -86,6 +87,34 @@ func BroadcastTX(ctx client.Context, flags *pflag.FlagSet, msgs ...sdk.Msg) erro
 
 }
 
+// PrepareFactory has been copied from cosmos-sdk to make it public.
```
hydrogen18, line 90: This is kind of weird? Why is it needed to copy a function from cosmos SDK into our code?

x/cert/keeper/key.go
```diff
@@ -31,18 +30,23 @@ func certificateKey(id types.CertID) []byte {
...
+	if _, err := buf.Write(address.MustLengthPrefix(id.Bytes())); err != nil {
 		panic(err)
 	}
 
 	return buf.Bytes()
 }
 
 func certificateSerialFromKey(key []byte) big.Int {
-	if len(key) < keyAddrPrefixLen+1 {
+	if len(key) < keyAddrPrefixLen {
```
hydrogen18, line 42: I don't think we need this comparison at this time, its redundant with the one below it.

app/app.go
troian: can we do it using this pattern `akash_v0.13.0_cosmos_v0.43.0` as `-` is valid symbol in semantic versioning indicating it is prerelease

arijitAD: Done

sdkutil/address.go
arijitAD: Thanks, corrected.

sdkutil/broadcast.go
arijitAD: That function was previously public, but `cosmos v0.43.0` made it private:
https://github.com/cosmos/cosmos-sdk/blame/release/v0.43.x/CHANGELOG.md#L114

This is being used by `BroadcastTX()` which is also a function originally defined in `cosmos-sdk` but slightly modified for a use-case of ours. So, there is no other option than to copy even the `PrepareFactory` function to our code.

x/cert/keeper/key.go
arijitAD: Its not redundant, both of them check different indexes before accessing those indexes.
This check exists because we access the index `keyAddrPrefixLen-1` in the next line. If we don't do this check here, then it will anyways panic there with `array index out of range`.

The other check happening below it is also similar, just checking that the index exists before accessing it. If that is not done, then even the next line after that check would panic with `index out of range`.

So, either we should remove both the checks as a panic will anyways be reported (just the panic message will be different), or should keep them both (to have the panic message be as per our choice).

I just kept the convention of having the panic message be of our choice.

x/deployment/legacy/v013/store.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// MigrateStore performs in-place store migrations from v0.12 to v0.13. The
+// migration includes:
+//
+// - Change addresses to be length-prefixed
+func MigrateStore(ctx sdk.Context, storeKey sdk.StoreKey) error {
+	store := ctx.KVStore(storeKey)
+	v013.MigratePrefixBech32AddrBytes(store, types.DeploymentPrefix())
+	v013.MigratePrefixBech32AddrBytes(store, types.GroupPrefix())
```
boz, line 16: Do the addresses need to be migrated from ascii/hex to binary here?

boz (commented):
Looks great!  I might have missed it, but I didn't see address migration for `x/deployment`?

x/deployment/legacy/v013/store.go
arijitAD: That is what is being done by `v013.MigratePrefixBech32AddrBytes()`. It migrating addresses from the bech32 string format to length prefixed binary format.

boz (approved):
:rocket:

app/app.go
hydrogen18: Can you put a link to that in the source code? Someone else may want to look at that as well.

PR #1348, opened 2021-08-17
feat: persistent storage

troian:
1. update SDL to handle multiple storage requests
2. use statefulSet when deployment needs persistent storage

**Parts**
- [x] SDL
- [x] Resource
- [x] Proto
- [x] Inventory
- [x] Attributes
- [x] Bidengine
- [x] Provider create/update
- [x] Manifest

**tests**
- [x] SDL
- [x] Resource
- [ ] Manifest
- [ ] K8S Integration


pkg/apis/akash.network/v1/types.go
```diff
@@ -40,7 +40,30 @@ type ManifestSpec struct {
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type StorageClassState struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	// +optional
+	metav1.ObjectMeta `json:"metadata"`
+
+	Spec   StorageClassStateSpec
```
hydrogen18, line 54: JSON tags on this line & the next?

pkg/apis/akash.network/v1/types.go
```diff
@@ -40,7 +40,30 @@ type ManifestSpec struct {
...
+	Status StorageClassStateStatus
+}
+
+type StorageClassStateStatus struct {
+	State   string `json:"state,omitempty"`
+	Message string `json:"message,omitempty"`
+}
+
+type StorageClassStateSpec struct {
+	Capacity  uint64 `json:"capacity"`
```
hydrogen18, line 64: OmitEmpty on these json tags?

provider/bidengine/pricing.go
```diff
@@ -90,9 +91,9 @@ func (fp scalePricing) CalculatePrice(_ context.Context, _ string, gspec *dtypes
 		memoryQuantity = memoryQuantity.Mul(groupCount)
 		memoryTotal = memoryTotal.Add(memoryQuantity)
 
-		storageQuantity := decimal.NewFromBigInt(group.Resources.Storage.Quantity.Val.BigInt(), 0)
-		storageQuantity = storageQuantity.Mul(groupCount)
-		storageTotal = storageTotal.Add(storageQuantity)
+		// storageQuantity := decimal.NewFromBigInt(group.Resources.Storage.Quantity.Val.BigInt(), 0)
```
hydrogen18, line 94: remove commented out code?

provider/bidengine/provider_attributes.go
```diff
@@ -88,6 +141,23 @@ loop:
...
+			case pass.currattr <- pass.attr:
+			default:
+			}
+		case attr := <-pass.newattr:
+			pass.attr = attr
+			pass.pushCurrAttributes()
+		case <-pass.errfetchattr:
+			// if attributes fetch fails give it retry withing reasonable timeout
+			time.AfterFunc(attrFetchRetryPeriod, func() {
+				pass.fetchAttributes()
```
hydrogen18, line 159: I'm trying to figure out what the utility of this `.fetchAttributes` method is? It looks like the same goroutine that receives from `pass.fetchattr` also calls `pass.fetchAttributes`. Couldn't you just use a boolean flag to indicate something needed to be done?

provider/bidengine/provider_attributes.go
```diff
@@ -285,3 +377,75 @@ func (pass *providerAttrSignatureService) GetAuditorAttributeSignatures(auditor
...
+
+func (pass *providerAttrSignatureService) tryFetchAttributes(ctx context.Context) {
+	select {
+	case pass.fetchInProgress <- struct{}{}:
+		go func() {
+			var err error
+			defer func() {
+				<-pass.fetchInProgress
+				if err != nil {
+					pass.errfetchattr <- err
```
hydrogen18, line 414: it seems like `errFetchAttr` should be  camel case as well for consistency.

provider/cluster/kube/builder/netpol.go
```diff
@@ -0,0 +1,430 @@
...
+				PolicyTypes: []netv1.PolicyType{
+					netv1.PolicyTypeIngress,
+				},
+			},
+		}
+		result = append(result, &policy)
+	}
+
+	return result, nil
+	/**
```
hydrogen18, line 197: We can get rid of this

provider/cluster/kube/builder/statefulset.go
```diff
@@ -0,0 +1,80 @@
...
+				Spec: corev1.PodSpec{
+					RuntimeClassName: effectiveRuntimeClassName,
+					SecurityContext: &corev1.PodSecurityContext{
+						RunAsNonRoot: &falseValue,
+					},
+					AutomountServiceAccountToken: &falseValue,
+					Containers:                   []corev1.Container{b.container()},
+				},
+			},
+			VolumeClaimTemplates: b.persistentVolumeClaims(),
```
hydrogen18, line 63: Shouldn't the `persistentVolumeClaims()` function be defined on this type since it is only used here?

provider/cluster/kube/builder/storageclassstate.go
```diff
@@ -0,0 +1,32 @@
+package builder
```
hydrogen18, line 1: Is this unused?

provider/cluster/kube/builder/workload.go
```diff
@@ -0,0 +1,180 @@
...
+		},
+		ImagePullPolicy: corev1.PullIfNotPresent,
+		SecurityContext: &corev1.SecurityContext{
+			RunAsNonRoot:             &falseValue,
+			Privileged:               &falseValue,
+			AllowPrivilegeEscalation: &falseValue,
+		},
+	}
+
+	if cpu := b.service.Resources.CPU; cpu != nil {
```
hydrogen18, line 70: I know this is basically identical to what is on master, but do we actually have a case where this can be `nil`? Same question applies to memory below this. Thoughts @boz  ?

provider/cluster/kube/builder/workload.go
```diff
@@ -0,0 +1,180 @@
...
+		kcontainer.Resources.Requests[corev1.ResourceMemory] = resource.NewQuantity(int64(requestedMem.Value()), resource.DecimalSI).DeepCopy()
+		kcontainer.Resources.Limits[corev1.ResourceMemory] = resource.NewQuantity(int64(mem.Quantity.Value()), resource.DecimalSI).DeepCopy()
+	}
+
+	for _, storage := range b.service.Resources.Storage {
+		attr := storage.Attributes.Find(StorageAttributeMount)
+		if mount, _ := attr.AsString(); mount != "" {
+			kcontainer.VolumeMounts = append(kcontainer.VolumeMounts, corev1.VolumeMount{
+				// matches VolumeName in persistentVolumeClaims below
+				Name: fmt.Sprintf("%s-%s", b.service.Name, storage),
```
hydrogen18, line 87: What does this name look like? It's calling `%s` on an object, which I know works but im trying to figure out what is actually created here

provider/cluster/kube/client.go
```diff
@@ -710,13 +747,97 @@ func (rp resourcePair) available() resource.Quantity {
...
+	storageclasses            map[string]bool
+}
+
+type NodeStorage struct {
+}
+
+type ActiveClusterStorage struct {
+}
+
+func (c *client) countKubeCall(err error, name string) {
```
hydrogen18, line 768: does this need to be part of the client ?

provider/cluster/kube/client.go
```diff
@@ -756,15 +875,27 @@ func (c *client) activeNodes(ctx context.Context) (map[string]nodeResources, err
...
 				allocatable: memory,
 			},
-			storage: resourcePair{
+			ephemeralStorage: resourcePair{
 				allocatable: storage,
 			},
+			volumesAttached: resourcePair{
+				allocated: *resource.NewQuantity(int64(len(knode.Status.VolumesAttached)), resource.DecimalSI),
+			},
+			storageclasses: make(map[string]bool),
```
hydrogen18, line 884: Camel case `storageClasses` ?

pkg/apis/akash.network/v1/types.go
troian: idea they are not supposed to be empty

provider/cluster/kube/client.go
```diff
@@ -187,26 +199,42 @@ func (c *client) Deploy(ctx context.Context, lid mtypes.LeaseID, group *manifest
...
 		}
 
 		if len(service.Expose) == 0 {
 			c.log.Debug("no services", "lease", lid, "service", service.Name)
 			continue
 		}
 
-		serviceBuilderLocal := newServiceBuilder(c.log, c.settings, lid, group, service, false)
-		if serviceBuilderLocal.any() {
+		serviceBuilderLocal := builder.NewService(c.log, c.settings, lid, group, service, false)
```
hydrogen18, line 228: The name of this function sort of conflicts with some of our other code. Normally `NewService` creates an instance of something with an associated goroutine somewhere in our codebase. This `NewService` is actually just a helper for defining kubernetes stuff it appears. I'd prefer it if we could change the name to something less ambiguous.

provider/bidengine/provider_attributes.go
```diff
@@ -88,6 +141,23 @@ loop:
...
+			pass.processAttrReq(ctx, req)
+		case <-pass.pushattr:
+			select {
+			case pass.currattr <- pass.attr:
+			default:
+			}
+		case attr := <-pass.newattr:
+			pass.attr = attr
+			pass.pushCurrAttributes()
+		case <-pass.errfetchattr:
```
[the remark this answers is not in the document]
troian: 1. `fetchAttributes` won't write into `pass.fetchattr` channel if there is no capacity

provider/bidengine/pricing.go
troian: done

provider/cluster/kube/client.go
troian: it's being sitting there for a while. it should be middleware to k8s rest client

provider/cluster/client.go
```diff
@@ -69,12 +72,34 @@ func ErrorIsOkToSendToClient(err error) bool {
...
+	inv.nodes = append(inv.nodes, newNode(id, allocateable, available))
+}
+
+func (inv *inventory) Nodes() []ctypes.Node {
+	return inv.nodes
+}
+
+func (inv *inventory) AddStorageClass(class string) {
+
+}
```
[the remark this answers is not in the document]
troian: removed

provider/cluster/kube/client.go
troian: 💯

provider/bidengine/pricing.go
```diff
@@ -90,9 +138,29 @@ func (fp scalePricing) CalculatePrice(_ context.Context, _ string, gspec *dtypes
...
+			storageClass := ""
+			attr := storage.Attributes.Find(sdl.StorageAttributePersistent)
+			if isPersistent, _ := attr.AsBool(); isPersistent {
+				attr = storage.Attributes.Find(sdl.StorageAttributeClass)
+				if class, set := attr.AsString(); set {
+					storageClass = class
+				}
+			}
+
+			total, exists := storageTotal[storageClass]
```
hydrogen18, line 154: so if `ifPersistent` is `false`, then `storageClass` is going to be the empty string. So obviously `exists` is false. Which means the code below that checks `if !exists` is going to return an error. Does this mean all non-persistent storage classes cause a failure in this code?

provider/bidengine/pricing.go
```diff
@@ -90,9 +138,29 @@ func (fp scalePricing) CalculatePrice(_ context.Context, _ string, gspec *dtypes
...
+
+			if !exists {
+				return sdk.Coin{}, errors.Wrapf(errNoPriceScaleForStorageClass, storageClass)
+			}
+
+			total = total.Add(storageQuantity)
```
hydrogen18, line 160: Shouldn't this go in `storageTotal` ?

provider/bidengine/pricing.go
```diff
@@ -246,11 +322,11 @@ func MakeShellScriptPricing(path string, processLimit uint, runtimeLimit time.Du
...
 
 type dataForScriptElement struct {
-	Memory           uint64 `json:"memory"`
-	CPU              uint64 `json:"cpu"`
-	Storage          uint64 `json:"storage"`
-	Count            uint32 `json:"count"`
-	EndpointQuantity int    `json:"endpoint_quantity"`
+	Memory           uint64            `json:"memory"`
+	CPU              uint64            `json:"cpu"`
+	Storage          map[string]uint64 `json:"storage"`
```
hydrogen18, line 327: This needs to be communicated as a breaking change in the release notes. Otherwise it should be fine.

provider/cluster/kube/builder/workload.go
troian: There was a missing field.
It is now `Name: fmt.Sprintf("%s-%s", b.service.Name, storage.Name),`

provider/bidengine/pricing.go
troian: fixed

provider/cluster/client.go
```diff
@@ -69,12 +72,34 @@ func ErrorIsOkToSendToClient(err error) bool {
 
 type node struct {
 	id                    string
-	availableResources    atypes.ResourceUnits
-	allocateableResources atypes.ResourceUnits
+	availableResources    ctypes.ResourceUnits
+	allocateableResources ctypes.ResourceUnits
+}
+
+var _ ctypes.Node = (*node)(nil)
```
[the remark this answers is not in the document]
troian: code removed

provider/cluster/client.go
```diff
@@ -37,7 +37,10 @@ var (
 	errNotImplemented = errors.New("not implemented")
 )
 
+var _ Client = (*nullClient)(nil)
```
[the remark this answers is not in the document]
troian: from readability prospective it's better to separate logic blocks. or did you mean?
```go
var (
    _ Client = (*nullClient)(nil)
)
```

proto/akash/base/v1beta1/resource.proto
```diff
@@ -24,22 +25,31 @@ message Memory {
...
 message Storage {
   option (gogoproto.equal) = true;
-  ResourceValue quantity   = 1
-      [(gogoproto.nullable) = false, (gogoproto.jsontag) = "size", (gogoproto.moretags) = "yaml:\"size\""];
-  repeated Attribute attributes = 2 [
+  string name = 1 [
+    (gogoproto.jsontag) = "name",
+    (gogoproto.moretags) = "yaml:\"name\""
+  ];
+  ResourceValue quantity = 2 [
```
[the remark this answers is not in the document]
troian: i think it won't be backwards compatible anyways as new field introduces whole new api version. correct me if i'm wrong

provider/cluster/kube/builder/workload.go
troian: yeah, good point. even with gpu only workloads cpu is required

boz:
What's going on with this one?

PR #1349, opened 2021-08-18
Convert hostname management to an operator

hydrogen18:
This accomplishes the follow things

1. Converts hostname mangement to an operator approach, a separate process that monitors CRDs
2. Fixes a bug with multiple service expose entries in SDL, **without changing the manifest format** 
3. Allows migrating hostnames between deployments with the same owner
4. Allows configuration of NGINX ingress options like maximum body size
5. Removes the Cluster Client configuration from the object (see below)
6. Changed our kustomize scripts not to authorize the entire akash-services namespace as admin
7. Added a kustomize script for the hostname operator

**Implementation notes**

1. I added a new CRD that maps a hostname to a service in a deployment
2. I used an approach of enumerating all the CRD entries, then starting a watch on them. The Kubernetes informer code around this isn't very useful as there is no easy way to listen for events
3. The CRD changes to the manifest entry are additive, and it appears kubernetes is just fine with this

You can run the kube example to see how this works presently, there is another target `make hostname-operator` that runs the code to manage the ingress resources

**Removal of cluster client from the kube client object**

The kube cluster client object took a Settings object when it was created. Inside the hostname operator, one of my goals is to have as little configuration as possible. The Settings object was only used by the `Deploy()` and `LeaseStatus()` functions. I originally wanted to split up the client into a bunch of different objects but didn't find any clean way to do that. What I did instead is required the `context` passed into those functions to have the `Settings` as a value inside of it. This is a little strange, but breaks the requirement that the creator of the client have the configuration to pass in. It also avoids some import cycles which break compilation.

**TODO**

1. ~~Add a kustomize entry to run the operator inside kubernetes with RBAC to manage the ingress resources~~
2. ~~Update the single example to run the operator in the local cluster~~
3. Update the readme's in `_run` to reflect this change
4. ~~Add tests~~
5. Determine if the kube operator can be scraped to by prometheus to get metrics


provider/cluster/kube/client_ingress.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	switch {
+	case err == nil:
+		_, err = c.kc.NetworkingV1().Ingresses(ns).Update(ctx, obj, metav1.UpdateOptions{})
+		metricsutils.IncCounterVecWithLabelValues(kubeCallsCounter, "networking-ingresses-update", err)
+	case kubeErrors.IsNotFound(err):
+		_, err = c.kc.NetworkingV1().Ingresses(ns).Create(ctx, obj, metav1.CreateOptions{})
+		metricsutils.IncCounterVecWithLabelValues(kubeCallsCounter, "networking-ingresses-create", err)
+	}
```
boz, line 90: I know we use this pattern elsewhere, it just struck me that there's a diabolical case:

1. two workers upserting an ingress.
2. both fail on `Update()`, continue on to `Create()`.
3. One `Create()` succeeds, one `Create()` fails*

* Or, possible database conflict/corruption at `3.`?

provider/cluster/manager.go
```diff
@@ -245,7 +248,35 @@ func (dm *deploymentManager) stopMonitor() {
...
+			LeaseID: dm.lease,
+			Group:   &groupCopy,
+			Status:  event.ClusterDeploymentUpdated,
+		}
+		err = dm.bus.Publish(ev)
+		if err != nil {
+			dm.log.Error("failed publishing event", "err", err)
+		}
+
+		close(chErr)
```
boz, line 277: would prefer `defer close(chErr)` as the first statement to make it easier to read.

provider/cluster/manager.go
```diff
@@ -254,16 +285,72 @@ func (dm *deploymentManager) startTeardown() <-chan error {
...
+				}
+			}
+		}
+	}
+
+	for host, serviceExpose := range hosts {
+		err = dm.client.DeclareHostname(ctx, dm.lease, host, hostToServiceName[host], uint32(serviceExpose.ExternalPort))
+		if err != nil {
+			// TODO - counter
+			return withheldHostnames, err
```
boz, line 348: should it unreserve undeclared hostnames?

provider/cluster/service.go
```diff
@@ -72,21 +76,39 @@ func NewService(ctx context.Context, session session.Session, bus pubsub.Bus, cl
...
 
-	hostnames := newHostnameService(ctx, cfg)
+	allHostnames, err := client.AllHostnames(ctx)
+	if err != nil {
+		sub.Close()
+		return nil, err
+	}
+
+	// Note: one side effect of this code is to add reservations for auto generated hostnames
+	// This is not normally done, but also doesn't cause any problems
```
boz, line 86: can they migrate those names?

boz (approved):
Really well done.

I ran out of steam towards the bottom, so I will try to give it another pass tomorrow, but it's a high-quality changeset and anything might come up can be dealt with later.

provider/cluster/service.go
hydrogen18: From what I remember if you try and deploy and SDL with a hostname in the autogenerated TLD it gets rejected entirely. I'll double check. It shouldn't be allowed.

provider/cluster/manager.go
hydrogen18: I'm not sure what you're asking? How would this reserve something that isn't declared?

provider/cluster/manager.go
hydrogen18: I'll go back over this to see what I am doing with this code

provider/cluster/kube/client_ingress.go
hydrogen18: I'll have to look see if two goroutines can be calling this at the same time.

provider/cluster/manager.go
boz: I might have read this wrong.  I thought it was:

1. reserving all hostnames [further above](https://github.com/ovrclk/akash/pull/1349/files/13529347b830af3c213901d93c94a3ecbdaf9069#diff-d0fa560a89bd5c31e83e29b07f984194896f8f4604cfceb60d02a3e9c10c30f6R293)
2. "declaring" the hostnames (I assumed creating the ingress plumbing configuration) here.

So if it's reserving then declaring, but declaring doesn't complete, we might want to unreserve the undeclared hostnames.  I was probably reading it wrong.

provider/cluster/kube/client_ingress.go
boz: It's okay.  We should probably make a generic optimistic locking/retry utility for everywhere that we do this.

provider/cluster/manager.go
boz: It's just a preferred pattern (there are more instances of this in this PR).

It makes reading the function a lot easier if you know that no matter what, this channel is going to be closed when it returns.

Also we aren't closing the channel when we send an error back here, which may be by-design inconsistency or not, but it adds more mental overhead when reading through it.

provider/cluster/manager.go
hydrogen18: What you're describing is correct.

If this fails, the deployment fails. At that time the goroutine in `run()` same file has a `defer` that calls `dm.hostnameService.ReleaseHostnames(dm.lease)` . So it should drop any hostnames reserved at that point.

provider/cluster/manager.go
```diff
@@ -254,16 +288,85 @@ func (dm *deploymentManager) startTeardown() <-chan error {
...
-func (dm *deploymentManager) doDeploy() error {
+func (dm *deploymentManager) doDeploy() ([]string, error) {
+	var err error
+
+	allHostnames := util.AllHostnamesOfManifestGroup(*dm.mgroup)
+	// Either reserve the hostnames, or confirm that they already are held
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	// Weird hack to tie this context to the lifecycle of the parent service, so this doesn't
```
hydrogen18, line 299: @boz @troian Is there a cleaner or saner way to handle this?

hydrogen18:
Splitting into two PRs to update each branch

PR #1352, opened 2021-08-23
feat(inflation): Implement Akash custom inflation function according to the whitepaper.

arijitAD:
## Motivation and Context

## Types of changes

## Checklist:


app/app.go
```diff
@@ -452,6 +467,30 @@ func NewApp(
...
+	inflation := sdk.NewDec(int64(i))
+
+	// min, max inflation based on a defined range parameter 'r'
+	// inflation range = [I(t) - I(t) * R, I(t) + I(t) * R]
+	r, err := sdk.NewDecFromStr("0.05") // let's say we allow 5% variance
+	if err != nil {
+		panic(err)
+	}
+	minInflation := minter.Inflation.Sub(minter.Inflation.Mul(r))
+	maxInflation := minter.Inflation.Add(minter.Inflation.Mul(r))
```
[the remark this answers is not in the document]
boz: I guess the issue I have is: does using `idealInflation + inflationRateChange` overly restrict the possible range?  If so then the entire range of acceptable inflation rates isn't available.

I think that we should use `idealInflation` to get `[max,min]`, and calculate `currentInflation = minter.Inflation + inflationRateChange`.  This way it will continue moving until an acceptable bonded ratio is found (in theory).

We also need to make a minimum `minInflation` - it can be `0` for now.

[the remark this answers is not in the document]
arijitAD: > I think that we should use `idealInflation` to get `[max,min]`, and calculate `currentInflation = minter.Inflation + inflationRateChange`. This way it will continue moving until an acceptable bonded ratio is found (in theory).
> 
> We also need to make a minimum `minInflation` - it can be `0` for now.

Done

app/app.go
```diff
@@ -463,6 +479,45 @@ func NewApp(
 	return app
 }
 
+func inflationCalculator(ctx sdk.Context, minter minttypes.Minter, params minttypes.Params, bondedRatio sdk.Dec) sdk.Dec {
+	i0 := 100.0
+	t := ctx.BlockTime().Sub(genesisTime).Seconds() / (60 * 60 * 8766) // years passed
```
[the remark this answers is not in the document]
boz: But `t` will be a severe step function at each year passing, instead of a more incremental daily increase.

go.mod
```diff
@@ -52,7 +52,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/ovrclk/cosmos-sdk v0.43.0-patches
+	github.com/cosmos/cosmos-sdk => github.com/ovrclk/cosmos-sdk v0.42.0-alpha1.0.20210909100325-11e54c8fc931
```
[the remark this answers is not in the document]
boz: Hmm, let's carry the patch over into `v0.44.4-patches` for now.

app/app.go
```diff
@@ -463,6 +481,52 @@ func NewApp(
...
+	var deploymentParams deploymenttypes.Params
+	deploymentParamSubspace.GetParamSet(ctx, &deploymentParams)
+	tHalf := float64(deploymentParams.InflationDecayFactor)
+
+	// Number of hours in an year = 8766 (averaging the leap year hours)
+	// Number of minutes in an hour = 60
+	// Number of seconds in a minute = 60
+	// => 60 * 60 * 8766 = Number of seconds per year
+	t := ctx.BlockTime().Sub(genesisTime).Seconds() / (60 * 60 * 8766) // years passed
+	i := 100.0 * math.Pow(2, -t/tHalf)                                 // 100 = initial inflation (i0)
```
boz, line 494: Initial inflation was not 100%.  Let's make this a parameter as well.

app/app.go
```diff
@@ -463,6 +481,52 @@ func NewApp(
...
+		Sub(bondedRatio.Quo(params.GoalBonded)).
+		Mul(params.InflationRateChange)
+	inflationRateChange := inflationRateChangePerYear.Quo(sdk.NewDec(int64(params.BlocksPerYear)))
+
+	// note inflationRateChange may be negative
+	currentInflation := minter.Inflation.Add(inflationRateChange)
+
+	// min, max currentInflation based on a defined range parameter 'r'
+	// currentInflation range = [I(t) - I(t) * R, I(t) + I(t) * R]
+	r, err := sdk.NewDecFromStr("0.05") // let's say we allow 5% variance
```
boz, line 508: Let's make the variance a parameter too.

boz (changes requested):
Looking great.

Let's not pollute `x/deployment` - make a new module for it (`x/inflation`), and put the actual inflation calculation in there as well.

I commented inline re a couple more parameters that we should make driven by governance.

go.mod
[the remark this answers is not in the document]
arijitAD: Done

app/app.go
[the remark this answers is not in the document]
arijitAD: No, `t` is a float value. It isn't an int, so it won't be a step function. Float value will increase continuously, unlike int.
i.e., `t` can represent `0.5` years. It won't go to 1 year directly from 0 years.

app/app.go
arijitAD: Done

app/app.go
arijitAD: Done

app/app.go
```diff
@@ -173,6 +178,12 @@ func NewApp(
 	logger log.Logger, db dbm.DB, tio io.Writer, loadLatest bool, invCheckPeriod uint, skipUpgradeHeights map[int64]bool,
 	homePath string, appOpts servertypes.AppOptions, options ...func(*bam.BaseApp),
 ) *AkashApp {
+	// find out the genesis time, to be used later in inflation calculation
+	if genDoc, err := tmtypes.GenesisDocFromFile(filepath.Join(homePath, "config/genesis.json")); err != nil {
+		panic(err)
```
arijitAD, line 183: This is panicking in the test environment, but not locally. Will have to debug it further.

arijitAD: Fixed

arijitAD:
```
app/export.go:232:7: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
	r := rand.New(rand.NewSource(seed))
```

This lint failure can't be fixed. The function from cosmos-sdk that we need to use, accepts only `math/rand` and not `crypto/rand`. So, we have no choice. Also, it is used only in tests. Doesn't affect security in any sensible way.

boz:
> ```
> app/export.go:232:7: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
> 	r := rand.New(rand.NewSource(seed))
> ```
> 
> This lint failure can't be fixed. The function from cosmos-sdk that we need to use, accepts only `math/rand` and not `crypto/rand`. So, we have no choice. Also, it is used only in tests. Doesn't affect security in any sensible way.

you can disable the check for that line with `// nolint: gosec` like [here](https://github.com/ovrclk/akash/blob/9dbc53f3b4d51025534228817bdbe791f3b0a1a9/provider/gateway/utils/utils.go#L21)

boz:
Please rebase against master.

arijitAD:
> you can disable the check for that line with `// nolint: gosec` like [here](https://github.com/ovrclk/akash/blob/9dbc53f3b4d51025534228817bdbe791f3b0a1a9/provider/gateway/utils/utils.go#L21)

Thanks, Done.

arijitAD:
> Please rebase against master.

Done.

app/export.go
```diff
@@ -224,3 +227,14 @@ func Setup(isCheckTx bool) *AkashApp {
 
 	return app
 }
+
+func OptsWithGenesisTime(seed int64) servertypes.AppOptions {
+	r := rand.New(rand.NewSource(seed)) // nolint: gosec
+	genTime := simulation.RandTimestamp(r)
+
+	appOpts := viper.New()
```
hydrogen18, line 235: What is the rationale behind creating our own viper instance here?

hydrogen18:
This seems to have genesis time as a configurable option? Wouldn't this create all sorts of issues if one validator was using a different time?

x/inflation/types/v1beta2/inflation_calculator.go
```diff
@@ -0,0 +1,68 @@
...
+	initialInflation, err := strconv.ParseFloat(inflationParams.InitialInflation, 32)
+	if err != nil {
+		panic(err)
+	}
+
+	// Number of hours in an year = 8766 (averaging the leap year hours)
+	// Number of minutes in an hour = 60
+	// Number of seconds in a minute = 60
+	// => 60 * 60 * 8766 = Number of seconds per year
+	t := ctx.BlockTime().Sub(GenesisTime).Seconds() / (60 * 60 * 8766) // years passed
```
hydrogen18, line 33: is `ctx.BlockTime()` the same across each node @boz  ? I haven't spent much time thinking about it. I mostly try and think in block height.

x/inflation/types/v1beta2/inflation_calculator.go
```diff
@@ -0,0 +1,68 @@
...
+	i := initialInflation * math.Pow(2, -t/tHalf)
```
hydrogen18, line 34: `i` is a bad choice of a variable name here. It is generally used for iterators, etc.

x/inflation/types/v1beta2/inflation_calculator.go
```diff
@@ -0,0 +1,68 @@
...
+	idealInflation := sdk.NewDec(int64(i))
```
hydrogen18, line 35: This converts an int to a decimal number, which implies some loss of precision. Whatever the intended operation is here should be documented in a comment.

app/app.go
```diff
@@ -173,6 +179,19 @@ func NewApp(
...
+		genTime, ok := v.(time.Time)
+		if !ok {
+			panic("expected GenesisTime to be a Time value")
+		}
+		inflationtypes.GenesisTime = genTime
+	} else if genDoc, err := tmtypes.GenesisDocFromFile(filepath.Join(homePath, "config/genesis.json")); err != nil {
+		panic(err)
+	} else {
+		inflationtypes.GenesisTime = genDoc.GenesisTime
+	}
```
boz, line 194: Put this in a helper function that returns the genesis time.

Please don't use global variables for state.  Find another way to pass the parameters into the inflation calculation.

x/inflation/alias.go
```diff
@@ -0,0 +1,23 @@
+package inflation
+
```
boz, line 2: I've never really liked this alias pattern - it seems unnecessary, especially when there are `var`s.  It's okay to keep for now but we should do a pass to clean these up.

x/inflation/types/v1beta2/inflation_calculator.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+var (
+	// these are set during NewApp and used in InflationCalculator
+
+	GenesisTime            time.Time
+	InflationParamSubspace paramstypes.Subspace
+)
+
+func InflationCalculator(ctx sdk.Context, minter minttypes.Minter, params minttypes.Params, bondedRatio sdk.Dec) sdk.Dec {
```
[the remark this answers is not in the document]
boz: agreed.  in particular, rounding issues will likely cause different nodes to produce different values.

x/inflation/types/v1beta2/inflation_calculator.go
```diff
@@ -0,0 +1,68 @@
...
+	var inflationParams Params
+	InflationParamSubspace.GetParamSet(ctx, &inflationParams)
+	tHalf := float64(inflationParams.InflationDecayFactor)
+	initialInflation, err := strconv.ParseFloat(inflationParams.InitialInflation, 32)
```
boz, line 24: Params should be represented by the appropriate type here.  See how `x/deploy` does it [here](https://github.com/ovrclk/akash/blob/9dbc53f3b4d51025534228817bdbe791f3b0a1a9/x/deployment/types/v1beta2/params.go).

I agree with @hydrogen18  - we should use a decimal or rational here.

x/inflation/types/v1beta2/inflation_calculator.go
boz: Yes, I believe it's equivalent.  I can double check.

I still think that the above calculation should not use integer division - it will be `0` for a whole year, then `1` for a year, then `2`, etc..., causing unnecessary and unwanted discontinuities on the anniversary of genesis.

x/inflation/types/v1beta2/inflation_calculator.go
boz: I don't think that there is any loss of precision from int to decimal.

I do think that we should use a consistent type from the outset of this whole thing - rational and decimal for arbitrary precision, then truncate as necessary at the end (or continuously via some large fixed precision).

x/inflation/types/v1beta2/params.go
```diff
@@ -0,0 +1,95 @@
...
+	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+	"github.com/pkg/errors"
+)
+
+var _ paramtypes.ParamSet = (*Params)(nil)
+
+const (
+	DefaultInflationDecayFactor uint32 = 2 // years
+	DefaultInitialInflation     string = "100.0"
+	DefaultVarince              string = "0.05"
```
boz, line 15: Make these `sdk.Decimal` or even integers that are used to populate a decimal later.  Maybe look at how some other parameters in cosmos-sdk are implemented.

boz (changes requested):

app/export.go
arijitAD: `NewApp` needs an instance of `servertypes.AppOptions` interface.
Viper implements this interface, so using it for testing.

x/inflation/types/v1beta2/inflation_calculator.go
hydrogen18: My statement was accurate. It converts a float, to an int, to a decimal. We should try and document the conversion from float to decimal or avoid the whole thing.

x/inflation/types/v1beta2/params.go
arijitAD: Done

x/inflation/types/v1beta2/inflation_calculator.go
arijitAD: Done

app/app.go
arijitAD: Done

x/inflation/types/v1beta2/inflation_calculator.go
arijitAD: Done

x/inflation/types/v1beta2/inflation_calculator.go
arijitAD: Removed.

x/inflation/types/v1beta2/inflation_calculator.go
arijitAD: It is not using integer division.

`ctx.BlockTime().Sub(GenesisTime).Seconds()` returns a `float64`, so it is indeed a continuous function, and not a step function. i.e., it can have values like 0.1 -> 0.2 -> 0.3 -> and so on...

x/inflation/types/v1beta2/inflation_calculator.go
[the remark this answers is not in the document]
arijitAD: Done

proto/akash/inflation/v1beta2/genesis.proto
```diff
@@ -0,0 +1,16 @@
...
+package akash.inflation.v1beta2;
+
+import "gogoproto/gogo.proto";
+import "akash/inflation/v1beta2/params.proto";
+
+option go_package = "github.com/ovrclk/akash/x/inflation/types/v1beta2";
+
+// GenesisState stores slice of genesis deployment instance
+message GenesisState {
+  Params params = 2 [
```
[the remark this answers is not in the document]
arijitAD: Thanks, fixed.

x/inflation/types/v1beta2/params.go
```diff
@@ -0,0 +1,96 @@
...
+
+var (
+	DefaultInitialInflation = sdk.NewDec(100)
+	DefaultVarince          = sdk.MustNewDecFromStr("0.05")
+
+	MaxInitialInflation = sdk.NewDec(100)
+	MinInitialInflation = sdk.ZeroDec()
+
+	MaxVariance = sdk.NewDec(1)
+	MinVariance = sdk.ZeroDec()
```
boz, line 28: Make these functions so that they're constant-like.

x/inflation/types/v1beta2/inflation_calculator.go
```diff
@@ -0,0 +1,60 @@
...
+		tHalf := float64(inflationParams.InflationDecayFactor)
+
+		// Number of hours in an year = 8766 (averaging the leap year hours)
+		// Number of minutes in an hour = 60
+		// Number of seconds in a minute = 60
+		// => Number of seconds per year = 60 * 60 * 8766 = 31557600
+		t := ctx.BlockTime().Sub(genesisTime).Seconds() / 31557600 // years passed (can be a fraction, eg: 0.5)
+		idealInflation := inflationParams.InitialInflation.Mul(sdk.MustNewDecFromStr(
+			strconv.FormatFloat(math.Pow(2, -t/tHalf), 'f', -1, 64),
+		))
```
boz, line 29: Use `sdk.Dec` or standard `big.*`.

boz (changes requested):

arijitAD:
> This seems to have genesis time as a configurable option? Wouldn't this create all sorts of issues if one validator was using a different time?

The option is used to to supply the genesis time, only to make unit tests pass. Actual validators would have the genesis time in the config file, and that should be same for all validators, I think.

x/inflation/types/v1beta2/inflation_calculator.go
arijitAD: There's no `Pow()` equivalent on `sdk.Dec` or `big.*` that accepts a fraction. That's why using this way of formatting the `Pow()` float value to a string and parsing a `sdk.Dec` from that.

boz: I see.  How annoying.

How about using [this](https://github.com/robpike/ivy/blob/master/value/power.go#L24) from [this](https://github.com/golang/go/issues/14102), or something similar?

hydrogen18:
> > This seems to have genesis time as a configurable option? Wouldn't this create all sorts of issues if one validator was using a different time?
> 
> The option is used to to supply the genesis time, only to make unit tests pass. Actual validators would have the genesis time in the config file, and that should be same for all validators, I think.

Can't the tests just assume some random arbitrary genesis time then? Just pick a time that makes sense, i.e. sometime in 2018 or something?

x/inflation/types/v1beta2/inflation_calculator.go
hydrogen18: I think `ctx.BlockTime().Sub(GenesisTime)` returns a `time.Duration` type. So can we just call `.Milliseconds()` to get an "exact" numerical value here? Millisecond level precision should be more than accurate enough for calculating interest I would think

boz:
@arijitAD, [this](https://github.com/cockroachdb/apd) looks promising.

boz:
There's also this, which looks like it's more actively maintained: https://github.com/ericlagergren/decimal

x/inflation/types/v1beta2/params.go
arijitAD: Done

x/inflation/types/v1beta2/inflation_calculator.go
arijitAD: Done

arijitAD:
> Can't the tests just assume some random arbitrary genesis time then? Just pick a time that makes sense, i.e. sometime in 2018 or something?

No, having a random genesis time doesn't work with tests. They start failing in that case. Seems, there are some parts in tests which depend on the genesis time being correctly set.

x/inflation/types/v1beta2/inflation_calculator.go
```diff
@@ -0,0 +1,83 @@
...
+					Mul(
+						decimal.New(-1, 0),
+						decimal.WithPrecision(sdk.Precision).
+							Quo(
+								yearsPassed,
+								decimal.New(int64(inflationParams.InflationDecayFactor), 0),
+							),
+					),
+			)
+		idealInflation := inflationParams.InitialInflation.Mul(sdk.MustNewDecFromStr(pow.String()))
```
boz, line 52: Use [`Int()`](https://pkg.go.dev/github.com/ericlagergren/decimal?utm_source=godoc#Big.Int) or [`Float()`](https://pkg.go.dev/github.com/ericlagergren/decimal?utm_source=godoc#Big.Float).

Multiply then divide by precision if needed.

boz (changes requested):

x/inflation/types/v1beta2/inflation_calculator.go
arijitAD: `inflationParams.InitialInflation` is a `sdk.Dec`. `sdk.Dec` has 18 precision digits.
`pow` is a `decimal.Big`. I am using the same precision of 18 digits for `decimal.Big` everywhere.

So, converting `pow` to string and parsing it into `sdk.Dec` won't lose any precision.

Could you please elaborate why and where should we use `Int()/Float()`? Also, why and where we need to `multiply and divide by precision`?

boz: We don't need that much precision, especially for the final result.

If the result is less than one and we use 6 units of precision then you can make an sdk.Decimal by with something like:

`sdk.Decimal(math.Int(val * 10^6)) / 10^6`

proto/akash/inflation/v1beta2/params.proto
```diff
@@ -0,0 +1,35 @@
...
+  // It is a decimal value in the range [0.0, 100.0].
+  string initial_inflation = 2 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable)   = false,
+    (gogoproto.customname) = "InitialInflation",
+    (gogoproto.jsontag)    = "initial_inflation",
+    (gogoproto.moretags)   = "yaml:\"initial_inflation\""
+  ];
+
+  // Variance defines the fraction by which inflation can vary from its previous value in a block.
```
boz, line 26: Isn't it how much it can vary from the ideal inflation curve?

x/inflation/types/v1beta2/params.go
```diff
@@ -0,0 +1,94 @@
...
+	v, ok := i.(uint32)
+	if !ok || v < 1 {
+		return errors.Wrapf(ErrInvalidParam, "%T", i)
+	}
+
+	return nil
+}
+
+func validateInitialInflation(i interface{}) error {
+	v, ok := i.(sdk.Dec)
```
boz, line 73: Will this work if the param type is a string? https://github.com/ovrclk/akash/pull/1352/files#diff-bbf326520f4fa8aa2e6f40542d4dfb257fe6bc3233f0d5b382b7021416619d3bR18

hydrogen18:
> > Can't the tests just assume some random arbitrary genesis time then? Just pick a time that makes sense, i.e. sometime in 2018 or something?
> 
> No, having a random genesis time doesn't work with tests. They start failing in that case. Seems, there are some parts in tests which depend on the genesis time being correctly set as per the test seed.

If that is the case then can `OptsWithGenesisTime` be moved into a `_test.go` file and in the regular codepath we continue to use `simapp.EmptyAppOptions{}` ?

x/inflation/types/v1beta2/params.go
arijitAD: That is a string only in proto definition. In Go code, it maps to `sdk.Dec` because of this configured option in the proto definition:
```
(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
```
This is how other modules in `cosmos-sdk` are mapping their parameters to `sdk.Dec`.

x/inflation/types/v1beta2/inflation_calculator.go
arijitAD: Done

proto/akash/inflation/v1beta2/params.proto
arijitAD: Yes, updated comment.

arijitAD:
> If that is the case then can `OptsWithGenesisTime` be moved into a `_test.go` file and in the regular codepath we continue to use `simapp.EmptyAppOptions{}` ?

No, it doesn't seem feasible. There are existing test utility functions in non `_test.go` files, which have to use `OptsWithGenesisTime`.

boz (approved):
Looks great!

We'll have to adjust some of the defaults to make sure it matches mainnet, but this is good to go for now.

hydrogen18:
The e2e tests are failing with 

```
E[2021-11-22|13:45:28.281] applying statefulSet                         module=provider-cluster-kube err="StatefulSet.apps \"web\" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden" lease=akash1v7u8mq7npa6g7pcncnqgshc5dw5cf5m40gww26/100/1/1/akash1ra7cz7jsljstqk3230426949s79j079h8ptjg7 service=web
I[2021-11-22|13:45:28.284] declaring hostname               
```

@troian is this something added with storage?

troian:
@hydrogen18 correct, i’ll give it a check why it updated multiple times

PR #1358, opened 2021-08-30
feat[DO NOT REVIEW]: storage codegen

troian:
codegen results for #1348


troian:
codegen is back to #1348

PR #1362, opened 2021-09-01
feat(provider): add script migration from v0.12.x hostname management…

hydrogen18:
This script should migrate providers from the old system to the new system. Please see the markdown file for explanation of how this is to be used.

### DIFF
--- a/script/provider_migrate_to_hostname_operator.py
+++ b/script/provider_migrate_to_hostname_operator.py
@@ -1,0 +1,244 @@
+import time
+import re
+import subprocess
+import sys
+import os
+import tempfile
+import json
+import logging
+import logging.config
+import pickle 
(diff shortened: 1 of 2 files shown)


boz (approved):
I think it looks okay.  FWIW I think it can all be done with just `kubectl` and `jq`.

Before we run this I'd like to take a raw backup using just `kubectl`, something like:

```sh
kubectl get ing -A -l akash.network -o json > ing-backup.json
```

script/provider_migrate_to_hostname_operator.py
```diff
@@ -0,0 +1,245 @@
+import time
+import re
+import kubernetes
```
hydrogen18, line 3: should be removed

PR #1373, opened 2021-09-15
Use command for services from SDL

hydrogen18:
It looks like we just were not setting command in the SDL at all. This fixes that


dmikey (approved):

troian:
remember to use conventional commit when merging

PR #1375, opened 2021-09-17
reduce default bid amount

dmikey:
Prop 12 has passed, the lowest effective bid accepted is 5,000,000 uakt. This lowers the default amount to match.

### DIFF
--- a/x/market/types/params.go
+++ b/x/market/types/params.go
@@ -9,7 +9,7 @@ import (
 var _ paramtypes.ParamSet = (*Params)(nil)
 
 var (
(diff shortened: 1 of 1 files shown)


boz (approved):
:rocket:

troian (approved):

hydrogen18 (approved):

hydrogen18:
@dmikey Long term I want to just query the parameter off the chain to get the right value. This should be fine for now however. 

@boz we should roll this into the v0.14 update since it is basically a no-code change right?

boz (approved):
:rocket:

boz:
Is this already in `master` and `mainnet/main`?

PR #1385, opened 2021-09-24
feat: Bid pricing based on USD scale

arijitAD:
This PR adds a script that uses the current USD price of an AKT to calculate the bid price.

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/provider/bidengine/pricing_test.go
+++ b/provider/bidengine/pricing_test.go
@@ -7,6 +7,7 @@ import (
 	io "io"
 	"math"
 	"math/big"
+	"net/http"
 	"os"
 	"os/exec"
 	"path"
@@ -15,12 +16,13 @@ import (
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/shopspring/decimal"
+	"github.com/stretchr/testify/require"
+
 	"github.com/ovrclk/akash/testutil"
 	atypes "github.com/ovrclk/akash/types"
 	"github.com/ovrclk/akash/types/unit"
 	dtypes "github.com/ovrclk/akash/x/deployment/types"
-	"github.com/shopspring/decimal"
-	"github.com/stretchr/testify/require"
 )
 
 func Test_ScalePricingRejectsAllZero(t *testing.T) {
@@ -547,6 +549,23 @@ func Test_ScriptPricingWritesJsonToStdin(t *testing.T) {
 }
 
 func Test_ScriptPricingFromScript(t *testing.T) {
+	const (
+		mockApiResponse = `{"akash-network":{"usd":3.57}}`
+		addr            = "localhost:8080"
+		expectedPrice   = 67843138
+	)
+
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(w, mockApiResponse)
+	})
+
+	go func() {
+		t.Fatal(http.ListenAndServe(addr, nil))
+	}()
+
+	err := os.Setenv("API_URL", addr)
+	require.NoError(t, err)
+
 	scriptPath, err := filepath.Abs("testdata/usd_pricing_oracle.sh")
 	require.NoError(t, err)
 	pricing, err := MakeShellScriptPricing(scriptPath, 1, 30000*time.Millisecond)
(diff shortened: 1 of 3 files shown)


provider/bidengine/testdata/usd_pricing_oracle.sh
```diff
@@ -0,0 +1,78 @@
...
+fi
+
+# finally, calculate the total cost in USD of all resources and validate it
+total_cost_usd=$(bc -l <<<"${cpu_cost_usd}+${memory_cost_usd}+${storage_cost_usd}+${endpoint_cost_usd}")
+if [ 1 -eq "$(bc <<<"${total_cost_usd}<0")" ] || [ 0 -eq "$(bc <<<"${total_cost_usd}<=${MAX_INT64}")" ]; then
+  exit 1
+fi
+
+# call the API and find out the current USD price per AKT
+API_RESPONSE=$(curl -s "https://api.coingecko.com/api/v3/simple/price?ids=akash-network&vs_currencies=usd")
```
hydrogen18, line 63: Maybe the URL you are using should be an env. var. that is up at the top of the document? So it is easier to configure in the future.

provider/bidengine/testdata/usd_pricing_oracle.sh
```diff
@@ -0,0 +1,78 @@
...
+curl_exit_status=$?
+if [ $curl_exit_status != 0 ]; then
+  exit $curl_exit_status
+fi
+usd_per_akt=$(jq '."akash-network"."usd"' <<<"$API_RESPONSE")
```
hydrogen18, line 68: Since this is just bidding do we want to fail here if the answer for this number is 0?

provider/bidengine/pricing_test.go
```diff
@@ -545,6 +546,20 @@ func Test_ScriptPricingWritesJsonToStdin(t *testing.T) {
 	}
 }
 
+func Test_ScriptPricingFromScript(t *testing.T) {
+	scriptPath, err := filepath.Abs("testdata/usd_pricing_oracle.sh")
```
hydrogen18, line 550: Wouldn't this actually wind up making the unit test dependent on coin gecko (or whatever pricing oracle)? I don't think that is a good idea.

provider/bidengine/testdata/usd_pricing_oracle.sh
dmikey: Think we should have a better backup then that, what if the oracle drops completely?

hydrogen18:
@boz since this is just a script should be put this in a separate repo?

provider/bidengine/testdata/usd_pricing_oracle.sh
hydrogen18: Up on line 65, if curl fails the script fails. No bid happens.

boz:
> @boz since this is just a script should be put this in a separate repo?

I think it's okay to add to `script/` to use as a reference.

provider/bidengine/testdata/usd_pricing_oracle.sh
Rini-Tikki: Done

arijitAD: Done. Added env var.

provider/bidengine/testdata/usd_pricing_oracle.sh
arijitAD: Done

provider/bidengine/pricing_test.go
arijitAD: Done. Added a mock HTTP server.

hydrogen18 (approved):

provider/bidengine/pricing_test.go
```diff
@@ -545,6 +548,40 @@ func Test_ScriptPricingWritesJsonToStdin(t *testing.T) {
...
+		expectedPrice   = 67843138
+	)
+
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		_, err := fmt.Fprint(w, mockAPIResponse)
+		require.NoError(t, err)
+	})
+
+	go func() {
+		t.Error(http.ListenAndServe(addr, nil))
```
boz, line 564: It is best to use port `0` and get an ephemeral port in tests.

We should also shut down the server before returning.

arijitAD: Done. Added `httptest` server which starts with port `0`

boz (approved):
:rocket:

troian (approved):

PR #1386, opened 2021-09-24
fix(provider): Count ports for a service only in reservations

hydrogen18:
Right now the current code counts any service endpoint as consuming a port. This isn't really true for HTTP services, they get multiplexed all via the ingress controller.

This code just counts the endpoints that wind up being a NodePort in Kubernetes are consuming a port.


boz (approved):
:rocket:

boz:
is this in `master` and `mainnet/main`?

PR #1390, opened 2021-09-29
add lease closed on

hydrogen18:
This adds a new field to the lease message. It's called closed on. It records the block height that the lease is closed on. For an open lease it will be 0.

For leases that are already closed on the blockchain, they will forever have 0 in this spot. Trying to search all the transactions on the blockchain and fill that in doesn't seem worth it.


client/docs/swagger-ui/swagger.yaml
```diff
@@ -29728,6 +29740,9 @@ definitions:
                 created_at:
                   type: string
                   format: int64
+                closed_on:
+                  type: string
+                  format: int64
```
boz, line 29745: Is this set in `x/escrow`?

boz (approved):
Looks great.  RE `swagger.yaml` - it looks like fields are added to other objects, but the indenting is confusing to me.

client/docs/swagger-ui/swagger.yaml
hydrogen18: x/market since it is part of lease.

boz: It just looks like it's part of the escrow object:

2021-09-29-110333_632x358_scrot

This was the confusing bit to me.

hydrogen18: OK....that is weird. I have no idea. It's on the lease message itself. Let me look at what is getting generated for the swagger docs.

hydrogen18:
I'm reworking this based off the current master

PR #1394, opened 2021-10-01
feat: Allow fractional uakt for settlement

arijitAD:
## Description

This PR rebases #1201 on latest master branch.

Wherever possible, `v1beta1` protos have been kept. Only where necessary, `v1beta2` protos have been introduced and the corresponding `v1beta1` protos moved to `x/<module>/legacy/types/v1beta1` for the purpose of upgrade.

## Motivation and Context

## Types of changes

## Checklist:


app/app.go
```diff
@@ -498,6 +502,17 @@ func (app *AkashApp) registerUpgradeHandlers() {
...
+	app.keeper.upgrade.SetUpgradeHandler("akash_v0.15.0", func(ctx sdk.Context,
+		plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+		migrations.MigrateFractionalUakt(ctx,
+			app.keeper.deployment.(dkeeper.Keeper),
+			app.keeper.escrow.(ekeeper.Keeper),
+			app.keeper.market.(mkeeper.Keeper))
+
+		return app.mm.RunMigrations(ctx, app.configurator, fromVM)
+	})
+
```
arijitAD, line 515: I still need to move the upgrade logic for the corresponding modules to the right places. Instead of it lying in the `app/migrations` package, it should go to `legacy` package of each module.

arijitAD: This is done.

util/legacy/proto_compatibility_test.go
```diff
@@ -0,0 +1,331 @@
...
+	mtypesv1beta1 "github.com/ovrclk/akash/x/market/legacy/types/v1beta1"
+	mtypes "github.com/ovrclk/akash/x/market/types"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	cdc = app.MakeEncodingConfig().Marshaler.(codec.BinaryCodec)
+)
+
+func TestDeployment_DeploymentProto_IsCompatible(t *testing.T) {
```
arijitAD, line 23: This test indicates that v1beta1 `Deployment` proto serialization is compatible with v1beta2 `Deployment` proto.
It is because nothing changed inside the `Deployment` proto itself, instead only a import changed in the proto file, which is being used in some other type, not in the `Deployment` msg.

So, this indicates that we don't need the upgrade for `Deployment` proto.
I have included this upgrade ATM, but I think it should be removed as it is unnecessary.

app/app.go
```diff
@@ -464,7 +464,7 @@ func NewApp(
 }
 
 func (app *AkashApp) registerUpgradeHandlers() {
-	app.keeper.upgrade.SetUpgradeHandler("akash_v0.13.0_cosmos_v0.43.0", func(ctx sdk.Context,
+	app.keeper.upgrade.SetUpgradeHandler("akash_v0.15.0_cosmos_v0.44.1", func(ctx sdk.Context,
```
arijitAD, line 467: If this PR will be included in the `v0.15.0` release, then this change is fine. Also, we can actually combine multiple migrations for a module into one migration, and keep the `ConsensusVersion` for `deployment`,`escrow` & `market` at 2 (this PR changes that to 3).

But, if this PR is going in `v0.16` or later, then this change isn't valid. We will need to keep the existing upgrade handler as is, and register a new upgrade handler for this PR.

_run/common.mk
```diff
@@ -67,6 +67,7 @@ node-init-genesis: init-dirs
 	$(AKASH) init node0
 	cp "$(GENESIS_PATH)" "$(GENESIS_PATH).orig"
 	cat "$(GENESIS_PATH).orig" | \
+        jq -M '.app_state.gov.voting_params.voting_period = "15s"' | \
```
boz, line 70: :nail_care: consistent indenting

boz (commented):
:rocket:

_run/common.mk
hydrogen18: you can probably thank me for that

PR #1398, opened 2021-10-04
manifest watchdog stop fix

hydrogen18:
Fixes

1. only allow a manifest watchdog to be stopped once
2. log manifest errors

### DIFF
--- a/provider/manifest/watchdog.go
+++ b/provider/manifest/watchdog.go
@@ -11,24 +11,28 @@ import (
 )
 
 type watchdog struct {
-	leaseID types.LeaseID
-	timeout time.Duration
-	lc      lifecycle.Lifecycle
-	sess    session.Session
-	ctx     context.Context
-	log     log.Logger
+	leaseID  types.LeaseID
+	timeout  time.Duration
+	lc       lifecycle.Lifecycle
+	sess     session.Session
+	ctx      context.Context
+	log      log.Logger
+	stopLock chan struct{}
 }
 
 func newWatchdog(sess session.Session, parent <-chan struct{}, done chan<- dtypes.DeploymentID, leaseID types.LeaseID, timeout time.Duration) *watchdog {
 	ctx, cancel := context.WithCancel(context.Background())
 	result := &watchdog{
-		leaseID: leaseID,
-		timeout: timeout,
-		lc:      lifecycle.New(),
-		sess:    sess,
-		ctx:     ctx,
-		log:     sess.Log().With("leaseID", leaseID),
+		leaseID:  leaseID,
+		timeout:  timeout,
+		lc:       lifecycle.New(),
+		sess:     sess,
+		ctx:      ctx,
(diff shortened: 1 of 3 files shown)


provider/manifest/watchdog.go
```diff
@@ -46,7 +50,11 @@ func newWatchdog(sess session.Session, parent <-chan struct{}, done chan<- dtype
 }
 
 func (wd *watchdog) stop() {
-	wd.lc.ShutdownAsync(nil)
+	select {
+	case <-wd.stopLock:
+		wd.lc.ShutdownAsync(nil)
```
boz, line 55: I think this is a good band-aid, but a band-aid nonetheless.

If a second `stop()` is blocking, we should try to figure out why.

hydrogen18: Why is it a band aid? I thought the lifecycle component was never meant to be called twice for shutdown?

boz: `ShutdownAsync` can be called twice if it is not being called from the same thread as the main loop and `ShutdownInitiated()` is called quickly after the first one like is being done here: https://github.com/ovrclk/akash/blob/6184be05aa61bc2059b23dc2a72147de9180ba68/provider/manifest/watchdog.go#L59-L60

boz: I see - if [this](https://github.com/ovrclk/akash/blob/6184be05aa61bc2059b23dc2a72147de9180ba68/provider/manifest/watchdog.go#L64-L67) is blocking then even the first call will block.

hydrogen18: Yes, that is one case. I could change it so that the context used is tied into the lifecycle object and closed when the lifecycle object starts closing. I'm not entirely sure if that is desirable, it might actually mean the bid gets left open however.

boz: I think using `runner.Run()` to execute the tx asynchronously while still handling shutdowns is good enough.  wait for the runner to finish before returning if it has already started.

hydrogen18: OK, let me take a pass at that.

hydrogen18:
@boz I changed the code to call the shutdown initiated method initially then wait for an async. task to complete for the bid close.

provider/manifest/watchdog.go
```diff
@@ -52,20 +53,28 @@ func (wd *watchdog) stop() {
...
-		BidID: types.MakeBidID(wd.leaseID.OrderID(), wd.sess.Provider().Address()),
-	})
-	if err != nil {
-		wd.log.Error("failed closing bid", "err", err)
+	wd.lc.ShutdownInitiated(err)
+	if runch != nil {
+		result := <-runch
+		if err := result.Error(); err != nil {
+			wd.log.Error("failed closing bid", "err", err)
+		}
```
boz, line 78: Nice fix, I don't think I've seen it used this way before.

boz (approved):

PR #1399, opened 2021-10-05
manifest manager fixes

hydrogen18:
Looking at the code I found a number of potential problems. 

**Changes**

1. Attach a 20 second timeout in the request router to each manifest submit
2. Ignore all data in lease events not for this provider
3. Don't process the data in lease events, just use them to clear a local cache
4. handle timeouts in the manifest manager of each individual request, by ignoring timed out requests
5. fetch deployment leases with the appropriate query methods when needed, cache for up to 5 seconds. Don't allow requests to run for more than 1 minute
6. Timeouts in change 5 are configurable via CLI switches

### DIFF
--- a/provider/gateway/rest/router.go
+++ b/provider/gateway/rest/router.go
@@ -55,6 +55,7 @@ const (
 	// errors from private use staring
 	websocketInternalServerErrorCode = 4000
 	websocketLeaseNotFound           = 4001
+	manifestSubmitTimeout            = 120 * time.Second
 )
 
 type wsStreamConfig struct {
@@ -396,7 +397,7 @@ func createManifestHandler(log log.Logger, mclient pmanifest.Client) http.Handle
 			return
 		}
 
-		subctx, cancel := context.WithTimeout(req.Context(), 20*time.Second)
+		subctx, cancel := context.WithTimeout(req.Context(), manifestSubmitTimeout)
 		defer cancel()
 		if err := mclient.Submit(subctx, requestDeploymentID(req), mani); err != nil {
 			if errors.Is(err, manifestValidation.ErrInvalidManifest) {
--- a/provider/manifest/manager.go
+++ b/provider/manifest/manager.go
@@ -351,13 +351,15 @@ func (m *manager) emitReceivedEvents() {
 
 	latestManifest := m.manifests[len(m.manifests)-1]
 	m.log.Debug("publishing manifest received", "num-leases", len(m.localLeases))
+	copyOfData := new(dtypes.QueryDeploymentResponse)
+	*copyOfData = m.data
 	for _, lease := range m.localLeases {
 		m.log.Debug("publishing manifest received for lease", "lease_id", lease.LeaseID)
 		if err := m.bus.Publish(event.ManifestReceived{
 			LeaseID:    lease.LeaseID,
 			Group:      lease.Group,
 			Manifest:   latestManifest,
-			Deployment: &m.data,
+			Deployment: copyOfData,
 		}); err != nil {
 			m.log.Error("publishing event", "err", err, "lease", lease.LeaseID)
 		}


provider/gateway/rest/router.go
```diff
@@ -396,7 +396,9 @@ func createManifestHandler(_ log.Logger, mclient pmanifest.Client) http.HandlerF
 			return
 		}
 
-		if err := mclient.Submit(req.Context(), requestDeploymentID(req), mani); err != nil {
+		subctx, cancel := context.WithTimeout(req.Context(), 20*time.Second)
```
boz, line 399: constant up top please.  these kinds of things would ideally be configurable with environment variables.

provider/manifest/manager.go
```diff
@@ -274,40 +319,118 @@ func (m *manager) fillAllRequests(response error) {
...
+	response, err := m.session.Client().Query().Leases(subctx, &mtypes.QueryLeasesRequest{
+		Filters: mtypes.LeaseFilters{
+			Owner:    m.daddr.Owner,
+			DSeq:     m.daddr.DSeq,
+			GSeq:     0,
+			OSeq:     0,
+			Provider: m.session.Provider().GetOwner(),
+			State:    mtypes.LeaseActive.String(),
+		},
+		Pagination: nil,
```
hydrogen18, line 351: Do I actually need to paginate this request?

provider/manifest/manager.go
```diff
@@ -221,6 +237,30 @@ loop:
...
+func (m *manager) handleTimeout(req manifestRequest) {
+	x := -1
+	for i, reqEntry := range m.pendingRequests {
+		if reqEntry == req {
+			x = i
+			break
+		}
+	}
+
+	if x != -1 {
```
boz, line 249:
```go
if x == -1 {
  return
}
```

provider/manifest/manager.go
```diff
@@ -221,6 +237,30 @@ loop:
 
 }
 
+func (m *manager) handleTimeout(req manifestRequest) {
+	x := -1
+	for i, reqEntry := range m.pendingRequests {
+		if reqEntry == req {
+			x = i
+			break
+		}
```
boz, line 246: You can remove it here.

```go
if reqEntry != req {
  continue
}
reqEntry.ch <- reqEntry.ctx.Err()
m.pendingRequests = append(m.pendingRequests[:i],m.pendingRequests[i+1:]...)
return

File: provider/manifest/manager.go [hunk 5]
```diff
@@ -221,6 +237,30 @@ loop:
...
+	}
+
+	if x != -1 {
+		removed := m.requests[x]
```
boz, line 250: `removed := m.pendingRequests[x]`?

boz (commented):
Some stylistic changes requested, especially around hard-coded durations.

This all seems unnecessarily complicated (pre-existing cruft), I'll have to take a longer look at it, but we can give this a try.

provider/manifest/manager.go
hydrogen18: hmm, that is interesting. I need to see why the test is passing now

provider/gateway/rest/router.go
hydrogen18: > Some stylistic changes requested, especially around hard-coded durations.
> 
> This all seems unnecessarily complicated (pre-existing cruft), I'll have to take a longer look at it, but we can give this a try.

I think I see a simpler way of handling this, without the goroutine explosion I created. I'll give it a try while I incorporate your other comments.

hydrogen18:
I think I see a simpler way of handling this, without the goroutine explosion I created. I'll give it a try while I incorporate Adam's comments.

boz (approved):
I think some of the complexity was because on provider restart, some of the data does not need to be re-queried.

If this handles provider restarts well, then great, let's give it a go!

hydrogen18:
Two conditions lead to the data being fetched

1. a new manifest
2. new lease information via an event

When the provider starts up, the managers for existing deployments are started. But until a new manifest shows up or a lease is created the data should not be fetched. When those happen, we actually need to have the data. So it shouldn't create a 'storm' of RPC requests or anything like that if a provider is restarted.

PR #1400, opened 2021-10-06
fix(docs): add missing command

arno01:


script/provider_migrate_to_hostname_operator.md
```diff
@@ -82,5 +83,5 @@ The last line specifies the image tag and should correspond to whatever version
 To install the operator into kubernetes perform the following from the `/_docs` directory.
 
 ```
-kubectl kustomize ./kustomize/akash-hostname-operator | kubectl apply -f -
+kubectl kustomize _docs/kustomize/akash-hostname-operator | kubectl apply -f -
```
hydrogen18, line 86: If you look up on the lines above states you are in the `/_docs` directory. So if we are going to make this change then we need to remove that line.

PR #1415, opened 2021-10-14
Add a lease secondary key by provider

hydrogen18:
This uses a secondary key to store a reference to a lease, but by provider instead of by owner.


hydrogen18:
I think we have, I should be able to change that easily since we are creating new keys

PR #1422, opened 2021-10-21
fix kube dns in deployments

hydrogen18:
The way I made DNS work from within a pod was kind of hacky. This is a better method suggested by Andy.

### DIFF
--- a/_run/kube/Makefile
+++ b/_run/kube/Makefile
@@ -2,7 +2,7 @@ include ../common.mk
 include ../common-commands.mk
 include ../common-kind.mk
 
-SDL_PATH = grafana.yaml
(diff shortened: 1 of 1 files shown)


provider/cluster/kube/builder/builder.go
```diff
@@ -45,7 +45,8 @@ const (
 
 var (
 	dnsPort     = intstr.FromInt(53)
-	dnsProtocol = corev1.Protocol("UDP")
+	udpProtocol = corev1.Protocol("UDP")
+	tcpProtocol = corev1.Protocol("UDP")
```
arno01, line 49: @hydrogen18 should not be the "TCP" for tcpProtocol?

PR #1436, opened 2021-10-28
fix(ci): semver.sh - suppress error output

boz:
Suppress unnecessary output, see [here](https://github.com/ovrclk/akash/runs/4029548529?check_suite_focus=true#step:10:10) for example.


boz:
We should fix this along with #1437.

PR #1441, opened 2021-11-05
fix(provider): improve network policy for kube-dns access

arno01:


boz (approved):
:rocket:

hydrogen18:
@boz you good with rolling this out into mainnet? It will only affect new deployments, I don't have any migration code for existing ones.

boz:
> @boz you good with rolling this out into mainnet? It will only affect new deployments, I don't have any migration code for existing ones.

Yeah, let's do it.

PR #1447, opened 2021-11-10
hostname operator  - ignore problematic namespaces

hydrogen18:
There are sometimes CRDs from zombie or otherwise left over deployments that cause an error. Right now the hostname operator spins on that error.

The goal of this PR is to allow the hostname operator to move past that error by eventually just ignoring the problematic CRDs. They still should be removed but for now this is probably a best case workaround

**Changes**

1. Detect any errors correlating to a resource in kubernetes not found and track the count of them
2. Once the threshold is reached (3 by default) ignore subsequent events from those leases, they are probably a zombie
3. Ignored leases are tracked in a list that is pruned if it gets too large
4. Make all values that make sense to be configurable via flags, but should probably be set environmental variables
5. Add a read only HTTP interface
6. Add an endpoint for getting the state of the managed hostnames
7. Add an endpoint for getting the state of the ignored leases

The only errors that are tracked are those corresponding to a resource not being found, this is usually a missing namespace in production. General errors such as network connectivity, etc. aren't considered permanent and could result in all CRDs being ignored which is bad.

The ignored leases list is trimmed periodically if it is too large. By default the limit is over 100k entries, so hopefully we don't have many zombies running around. The pruning process tries to get rid of old entries first but always make sure to trim it down below the configured value. Otherwise an out of memory condition could happen.

The HTTP listener is setup to just have two endpoints for exposing data that is already there. We have no real way presently other than using the logs to figure out what the operator is doing. This should help with that. It also gives the namespace in k8s that would be associated with any potential entry, to aid in debugging.

The HTTP responses are not created dynamically. To avoid the use of any locks, the operator just periodically re-renders the responses if the data has changed. This avoids having to introduce any locks around the data and is probably more performant than rendering on each request anyways. The default interval for the speed at which the data is updated is 5 seconds. The `Last-Modified` header on the response indicates whenever the data was actually created, just to be clear to anyone looking at it.

I updated the deployment stuff in `_docs`  for the hostname operator to make sure the Ingress object exposes the new endpoint as intended.

### DIFF
--- a/provider/cmd/hostname_operator_test.go
+++ b/provider/cmd/hostname_operator_test.go
@@ -121,13 +121,42 @@ func makeHostnameOperatorScaffold(t *testing.T) *hostnameOperatorScaffold {
 
 	op := &hostnameOperator{
 		hostnames: make(map[string]managedHostname),
+		ignoreList: make(map[mtypes.LeaseID]ignoreListEntry),
 		client:    client,
 		log:       l,
+		ignoreListData: newPreparedResult(),
+		hostnamesData: newPreparedResult(),
 	}
 
 	scaffold.op = op
 
 	return scaffold
+}
+
+func TestHostnameOperatorPrune(t *testing.T) {
+	s := makeHostnameOperatorScaffold(t)
+	require.NotNil(t, s)
+	defer s.cancel()
+
+	s.op.prune() // does nothing, should be fine to call
+	s.client.On("GetManifestGroup", mock.Anything, mock.Anything).Return(false, crd.ManifestGroup{}, nil)
+	const testIterationCount = 200000
+	for i := 0 ; i != testIterationCount; i++ {
+		ev := testHostnameResourceEv{
+			leaseID:      testutil.LeaseID(t),
+			hostname:     "foobar.com",
+			eventType:    cluster.ProviderResourceAdd,
+			serviceName:  "the-ervice",
+			externalPort: 1234,
+		}
+		err := s.op.applyEvent(s.ctx, ev)
+		require.Error(t, err)
+	}
+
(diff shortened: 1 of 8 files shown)


provider/cmd/hostname_operator_test.go
```diff
@@ -156,14 +193,19 @@ func TestHostnameOperatorApplyDelete(t *testing.T) {
...
+
 }
 
 func TestHostnameOperatorApplyDeleteFails(t *testing.T) {
 	s := makeHostnameOperatorScaffold(t)
 	require.NotNil(t, s)
 	defer s.cancel()
 
-	const hostname = "qux.io"
+	const hostname = "lsn.io"
```
boz, line 208: Let's use [`.test` TLD](https://datatracker.ietf.org/doc/html/rfc2606#section-2).  There was some drama/hilarity recently with a test suite that was spamming a site that I can't find right now.

boz (commented):
this looks really clean, but I can't give it a thorough review right now.

we may need to make a utility to remove problematic hostnames in the meantime (#1451)

boz (approved):
Merging this now so we can get ewr1 running again.

It's not clear to me why we need to track problematic namespaces instead of just ignoring the errors and moving on.

The status API is cool.

I guess I should apply the new `kubectl kustomize` output?

hydrogen18:
> Merging this now so we can get ewr1 running again.
> 
> It's not clear to me why we need to track problematic namespaces instead of just ignoring the errors and moving on.
> 
> The status API is cool.
> 
> I guess I should apply the new `kubectl kustomize` output?

Yes. After the new output is applied you can send a request to the ingress controller with the default hostname (or kustomize it to whatever you want) to reach the status API

PR #1448, opened 2021-11-11
feat(node): add sequence number to endpoint type

hydrogen18:
This adds sequence number to the endpoint type in `v1beta2`. This is normally 0 and is a positive number whenever the IP address type is specified (in the future). This allows a provider with the group spec. from the block chain to count up the number of required IP address leases and make a bid appropriately.

There was already a migration for the Groups type which uses this endpoint type (indirectly) so I just documented in the migration we expect all prior data to have 0 present. I don't think this data is used anywhere else as we don't duplicate the deployment information in the blockchain data store.


troian:
do we need schema upgrade for this change?

boz (commented):
It might be difficult to figure out the correct endpoint mappings for a deployment when a deployment is updated.

* I have two services, one of them uses shared, the other leased.  I decide I want both of them to use the leased endpoint.  Should we not allow this?
* I have three leased endpoints, four services.  I change one service from one endpoint to another.
* I change the name of my endpoints - how is sequence determined?

Using a name instead of a sequence like the group name does _might_ help keep things manageable in some cases.

hydrogen18:
I don't think we need to use sequence number to actually deploy the application do we? The SDL specifies what actually happens when the deployment is deployed or updated.

hydrogen18:
> do we need schema upgrade for this change?

I think this already does that: https://github.com/ovrclk/akash/blob/master/x/deployment/keeper/migrations.go#L21

boz:
> I don't think we need to use sequence number to actually deploy the application do we? The SDL specifies what actually happens when the deployment is deployed or updated.

yeah, agreed.  I was assuming that the sequence numbers are derived from some ordering that is applied to the named endpoints in an sdl.  Maybe I need to read the description more closely.

hydrogen18:
My thought was each IP endpoint would get a sequence number starting at 1 and work its way up. For bidding, the provider just counts up the number of unique IPs in use once all groups are taken into consideration

boz:
ok just making it a list in the sdl and not a map should do that.   just have to be careful of people changing the contents or ordering of it.

hydrogen18:
I'm not sure I understand the difference. Sequence number has no informational value the way I look at it. What could the provider possibly infer from two different sequence numbers?

boz:
> I'm not sure I understand the difference. Sequence number has no informational value the way I look at it. What could the provider possibly infer from two different sequence numbers?

I'm talking about the SDL.  How do you determine the sequence number from the SDL?

hydrogen18:
You could use pretty much any scheme the way I look at it. As long as a single endpoint maps to a single sequence number it would be fine. You could just sort the map by endpoint name, then start at 1 and increment until you've assigned them all.

boz:
> You could use pretty much any scheme the way I look at it. As long as a single endpoint maps to a single sequence number it would be fine. You could just sort the map by endpoint name, then start at 1 and increment until you've assigned them all.

That is what I'm saying might take some care.  What if they change the name int the SDL, etc.

boz (approved):
:ship:

troian (approved):

PR #1455, opened 2021-11-15
feat(hostname operator): ignore problematic namespaces

hydrogen18:
Same change as the other changes, just keeping master to up to date


troian (approved):

boz (approved):
:rocket:

PR #1456, opened 2021-11-18
Fix provider behavior around existing bids

hydrogen18:
There were a few issues brought up here that I fixed

1. If the provider finds a bid on startup, apply the standard timeout to it before closing
2. If the provider finds a bid on startup, make sure it is in the bid open state
3. If the provider finds a bid on startup that is very old, close it immediately
4. If the provider order goroutine gets a bid closed event for that order then terminate. This means someone did it with the CLI most likely
5. Don't close the bid on bid created. The bid is already in the `BidLost` state due to the way the market module is implemented. The network basically closes it for us, saving us fees.

**Code style stuff**

1. Remove `order.bidPlaced`. I don't know why this was a member variable. It is just a local variable now.
2. Add a helper for computing bid timeout
3. Add a helper for determining if bid timeout is enabled
4. Add created at block height to provider session, indicating when the session object was created at
5. I estimated block time at 5 seconds, since I don't want to fetch it from the order goroutine
6. Remove our pervasive use of `context.Background()` in the provider commands. I don't know how this ever got there but `cmd.Context()` is the correct context to be using

### DIFF
--- a/provider/bidengine/order.go
+++ b/provider/bidengine/order.go
@@ -26,7 +26,6 @@ import (
 // order manages bidding and general lifecycle handling of an order.
 type order struct {
 	orderID   mtypes.OrderID
-	bidPlaced bool
 	cfg       Config
 
 	session                    session.Session
@@ -93,7 +92,6 @@ func newOrderInternal(svc *service, oid mtypes.OrderID, cfg Config, pass Provide
 	order := &order{
 		cfg:                        cfg,
 		orderID:                    oid,
-		bidPlaced:                  false,
 		session:                    session,
 		cluster:                    svc.cluster,
 		bus:                        svc.bus,
@@ -121,6 +119,18 @@ func newOrderInternal(svc *service, oid mtypes.OrderID, cfg Config, pass Provide
 
 var matchBidNotFound = regexp.MustCompile("^.+bid not found.+$")
 
+func (o *order) bidTimeoutEnabled() bool {
+	return o.cfg.BidTimeout > time.Duration(0)
+}
+
+func (o *order) getBidTimeout() <- chan time.Time{
+	if o.bidTimeoutEnabled() {
+		return time.After(o.cfg.BidTimeout)
+	}
+
+	return nil
+}
+
 func (o *order) run(checkForExistingBid bool) {
 	defer o.lc.ShutdownCompleted()
 	ctx, cancel := context.WithCancel(context.Background())
@@ -140,6 +150,7 @@ func (o *order) run(checkForExistingBid bool) {
 		reservation ctypes.Reservation
 
 		won bool
+		bidPlaced bool
 		msg *mtypes.MsgCreateBid
 	)
 
@@ -163,6 +174,7 @@ func (o *order) run(checkForExistingBid bool) {
 		storedGroupCh = groupch
 		groupch = nil
 	}
+
 loop:
 	for {
 		select {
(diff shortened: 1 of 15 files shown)


hydrogen18:
@arno01 This should address some of your concerns around provider startup with bids that already exist.

arno01:
I'm running Akash from this branch now, there is another issue -- whenever I'm closing the deployment, the `cluster.inventory.active` is not decrementing.

I.e. I had 3 active deployments, have created 2 more, ran for few minutes, have closed them (one after another as well) - still seeing 5 active deployments:

```
$ pstats.sh https://akash-provider.nixaid.com:8443
host                                    num  num-p  cpu  cpu-a  cpu-p  mem      mem-a    mem-p  fs       fs-a     fs-p
https://akash-provider.nixaid.com:8443  5    0      1.6  24.85  0      1.44531  67.7642  0      1.44531  1577.34  0
```

though, only 3 are active:

```
$ akash query market lease list --provider akash1nxq8gmsw2vlz3m68qvyvcf3kh6q269ajvqw6y0 --page 1 --limit 100 -o json --state active | jq -r '.leases[] | [ (.lease | (.lease_id | .provider, .owner, .dseq, .gseq, .oseq), (.price.amount|tonumber), .state), (.escrow_payment | .state, (.balance.amount, .withdrawn.amount)) ] | @csv'
"akash1nxq8gmsw2vlz3m68qvyvcf3kh6q269ajvqw6y0","akash1ae0zhju86vcpdhau82p9g40kpwshxfawqtqxu4","3416784",1,1,1,"active","open","0","170499"
"akash1nxq8gmsw2vlz3m68qvyvcf3kh6q269ajvqw6y0","akash1q5amxhp27afplheuy6fyuq0y3fs5zwxnjagk5s","3436369",1,1,2,"active","open","0","301726"
"akash1nxq8gmsw2vlz3m68qvyvcf3kh6q269ajvqw6y0","akash1wl9c7jawvptt9xsx3ac58jt5lnwdxayk905fad","3483003",1,1,1,"active","open","0","104252"
```

Workaround is to restart akash service to clear those numbers up.

hydrogen18:
So after the restart, the reported number is correct?

hydrogen18:
@arno01 it should be fixed now if you pull the branch down again, then recompile

@boz this branch should fix issues with the provider not wanting to stop when signaled to

provider/bidengine/order.go
```diff
@@ -182,8 +194,31 @@ loop:
...
+					// just use the height from when the session is created
+					createdAtBlock := bid.GetCreatedAt()
+					blockAge := createdAtBlock - o.session.CreatedAtBlockHeight()
+					const minTimePerBlock = 5 * time.Second
+					atLeastThisOld := time.Duration(blockAge) * minTimePerBlock
+					if atLeastThisOld > o.cfg.BidTimeout {
+						o.session.Log().Info("found expired bid", "block-height", createdAtBlock)
+						break loop
+					}
+				}
```
boz, line 219: put this in a method or function `isStaleBid()`

boz (changes requested):
Looks good. Was `o.bidPlaced` for handling startup?

arno01:
> @arno01 it should be fixed now if you pull the branch down again, then recompile

@hydrogen18 thank you very much! It's now working flawlessly! <3

provider/bidengine/order.go
hydrogen18: This is done now, it wraps up nicely into a helper

hydrogen18:
> Looks good. Was `o.bidPlaced` for handling startup?

I'm not really sure. We only access it now from a single code function so it makes sense to move it out.

The only thing I thought about was the fact that a provider may start dozens (or hundreds) of these goroutines on startup which all ask the RPC node for the bid info. However, querying for an exact bid should be a very lightweight request, it goes directly to a key in the data store and then loads the result. So I don't know that it is worth it to add any delay to the bid request at this time.

boz (approved):
:rocket:

colinlowenberg:
This pull request has been mentioned on **Akash Community**. There might be relevant details there:

https://forum.akash.network/t/hostname-error-status-code-was-not-success-500-hostname-not-allowed-host-name/3278/2

PR #1465, opened 2021-11-27
fix(sdkutil): get txhash from serialized tx rather than response

### LINKED ISSUE #1432: port `--broadcast-mode block` fix
Port the `--broadcast-mode block` fix to our [broadcast wrapper](https://github.com/ovrclk/akash/blob/mainnet/main/sdkutil/broadcast.go) to `mainnet/main` and `master`.

troian:
Pass command context to the BroadcastTX allows tests to
properly handle context.Cancel

fixes #1432

### DIFF
--- a/sdkutil/broadcast.go
+++ b/sdkutil/broadcast.go
@@ -3,6 +3,7 @@ package sdkutil
 import (
 	"bufio"
 	"context"
+	"encoding/hex"
 	"fmt"
 	"os"
 	"strings"
@@ -19,7 +20,7 @@ import (
 )
 
 const (
-	broadcastBlockRetryTimeout = 30 * time.Second
+	BroadcastBlockRetryTimeout = 300 * time.Second
 	broadcastBlockRetryPeriod  = time.Second
 
 	// sadface.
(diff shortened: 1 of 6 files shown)


boz (commented):
Can we re-use some code between `client/broadcaster/client.go` and `sdkutil/broadcaster.go`?  I don't want to have this brittle hack in two places.

For `mainnet/main` we only need `sdkutil/broadcaster.go` right now.

hydrogen18:
The polling frequency of 1 second seems like it might be too frequent but otherwise this seems fine

hydrogen18 (approved):

boz:
> The polling frequency of 1 second seems like it might be too frequent but otherwise this seems fine

It's for an interactive session (the cli), and it typically finishes within ~5-10 seconds

sdkutil/broadcast.go
```diff
@@ -119,32 +120,28 @@ func PrepareFactory(clientCtx client.Context, txf tx.Factory) (tx.Factory, error
...
-	case err.Error() != timeoutErrorMessage:
-		// other error, return
+	cres, err := cctx.BroadcastTxCommit(txb)
+	if err == nil {
+		// good job
+		return cres, nil
+	} else if !strings.HasSuffix(err.Error(), timeoutErrorMessage) {
 		return cres, err
-	default:
-		// timeout error, continue on to retry
```
boz, line 143: would be good to keep the comments.

boz (approved):

hydrogen18 (approved):

PR #1468, opened 2021-12-02
Add version endpoint

arijitAD:
## Motivation and Context

## Types of changes

## Checklist:


provider/cluster/kube/client.go
```diff
@@ -786,3 +787,7 @@ func (c *client) deploymentsForLease(ctx context.Context, lid mtypes.LeaseID) (m
 
 	return serviceStatus, nil
 }
+
+func (c *client) KubeVersion() (*version.Info, error) {
+	return c.kc.(*kubernetes.Clientset).ServerVersion()
```
hydrogen18, line 792: It seems like this method exists on an interface on its own. It's more robust to just cast to that interface type. Also we should check for the cast failing and just return an error.

https://github.com/kubernetes/client-go/blob/master/discovery/discovery_client.go#L120

boz: Yes, always use methods where available.

Take the time to do things right @arijitAD

boz (changes requested):

provider/cluster/kube/client.go
arijitAD: Fixed

provider/gateway/rest/router_test.go
```diff
@@ -307,6 +308,44 @@ func TestRouteDoesNotExist(t *testing.T) {
 	})
 }
 
+func TestRouteVersionOK(t *testing.T) {
+	runRouterTest(t, false, func(test *routerTest) {
+		status := versionInfo{
+			Akash: "",
```
hydrogen18, line 314: Please use a non-empty string here for the test. Anything is fine, just `"akashtest"` or whatever should be OK

hydrogen18:
Just the one comment about the test that should be improved, other than that this looks good to me.

hydrogen18 (approved):

provider/gateway/rest/router.go
```diff
@@ -391,6 +400,25 @@ func leaseShellHandler(log log.Logger, mclient pmanifest.Client, cclient cluster
 	}
 }
 
+type versionInfo struct {
+	Akash string            `json:"akash"`
```
boz, line 404: should we include commit hash, build date, go version, platform, etc...?

provider/gateway/rest/router_test.go
arijitAD: Done

provider/gateway/rest/router.go
arijitAD: Added all the useful information that is available via `akash version` command.

boz (approved):
:rocket:

PR #1469, opened 2021-12-02
fix(sdl): check volume counts to not exceed limits

troian:

### DIFF
--- a/provider/bidengine/order.go
+++ b/provider/bidengine/order.go
@@ -494,6 +494,12 @@ func (o *order) shouldBid(group *dtypes.Group) (bool, error) {
 		return false, nil
 	}
 
+	for _, resources := range group.GroupSpec.GetResources() {
+		if len(resources.Resources.Storage) > o.cfg.MaxGroupVolumes {
+			o.log.Info(fmt.Sprintf("unable to fulfill: group volumes count exceeds (%d > %d)", len(resources.Resources.Storage), o.cfg.MaxGroupVolumes))
+			return false, nil
+		}
+	}
 	signatureRequirements := group.GroupSpec.Requirements.SignedBy
 	if signatureRequirements.Size() != 0 {
 		// Check that the signature requirements are met for each attribute
--- a/provider/bidengine/config.go
+++ b/provider/bidengine/config.go
@@ -13,4 +13,5 @@ type Config struct {
 	Deposit         sdk.Coin
 	BidTimeout      time.Duration
 	Attributes      types.Attributes
+	MaxGroupVolumes int
 }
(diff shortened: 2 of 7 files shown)


x/deployment/types/v1beta2/resource_list_validation.go
```diff
@@ -129,6 +129,10 @@ func validateStorage(u types.Volumes) ([]sdk.Int, error) {
 		return nil, errors.Errorf("error: invalid unit storage, cannot be nil")
 	}
 
+	if len(u) > validationConfig.MaxGroupVolumes {
```
boz, line 132: Won't changing this require a chain upgrade?

boz (changes requested):
I think that we should enforce this in the provider, with the max being configurable.

This shouldn't make SDL *parsing* throw an error, tho we can print a warning by validating the manifest or deployment object.

x/deployment/types/v1beta2/resource_list_validation.go
troian: should not as we don't have deployments on both mainnet/testnet with storage.

boz: but once we do, it will, right?

troian: correct

boz: Let's not do that.  I don't want to get into another situation like we were in with the resource limits.

troian: actually not, it does not change message structure but limits amount of entries

boz: what if we want to change the limit

troian: well, yeah, it's gonna be nasty as software upgrade to the network will be required.
i'll make it as limit on the provider side, and print warning as you suggested above

hydrogen18: Are we just going to put a soft limit in on the provider where it doesn't bid on anything with less than a configured amount of volumes?

troian: yeah

boz (approved):
:ship:

provider/bidengine/order.go
```diff
@@ -434,6 +434,12 @@ func (o *order) shouldBid(group *dtypes.Group) (bool, error) {
 		return false, nil
 	}
 
+	for _, resources := range group.GroupSpec.GetResources() {
+		if len(resources.Resources.Storage) > o.cfg.MaxGroupVolumes {
+			o.log.Debug(fmt.Sprintf("unable to fulfill: group volumes count exceeds (%d > %d)", len(resources.Resources.Storage), o.cfg.MaxGroupVolumes))
```
hydrogen18, line 439: Seems like this could just be `Info` log level? It shouldn't happen too often.

hydrogen18 (approved):

provider/bidengine/order.go
troian: indeed

PR #1472, opened 2021-12-05
docs(storage): setup and usage of persistent storage

troian:


_docs/persistentstorage.md
```diff
@@ -0,0 +1,305 @@
...
+In `profile` the storage has been upgraded to support multiple entries. Both old and new notations are supported. Both examples below are showing valid use-cases of the sdl
+
+```yaml
+storage:
+  size: 512Mi
+```
+
+```yaml
+storage:
+  - size: 512Mi
```
hydrogen18, line 72: Above example has `size: 512Mi` whereas this has `- size: 512Mi` .Is this supposed to be a list ? Or a dictionary?

_docs/persistentstorage.md
```diff
@@ -0,0 +1,305 @@
...
+  - name: data
+    size: 1Gi
+```
+
+##### Storage name (alias)
+
+Each entry has new field `name` used by [services](#services) to reference various storage specific parameters. It can be omitted for single value usecase and default value is set to `default`
```
hydrogen18, line 79: Can we drop this and just make the naming mandatory? We are trying to avoid as much "magic" behavior in the SDL as possible.

_docs/persistentstorage.md
```diff
@@ -0,0 +1,305 @@
...
+storage:
+  - size: 512Mi
+    # name is set to "default"
+  - size: 1Gi
+    # should have name set
+```
+
+##### Storage attributes
+
+1. `persistent` - either volume requires persistence or not. Default value is set to `false`
```
hydrogen18, line 111: What are the implications to the provider for setting this or not?

_docs/persistentstorage.md
hydrogen18, line 79: Do all replicas of a service get the same mounted volume?

_docs/persistentstorage.md
```diff
@@ -0,0 +1,305 @@
...
+    attributes:
+      persistent: true
+      class: beta2
+```
+
+#### Services
+
+If deployment references profile with multiple volumes then service **must** include params section to configure mount points.
+
+all examples in this section below have been shortened and have only information required to show usage of storage
```
hydrogen18, line 130: It seems like we should just include complete examples of each behavior here

_docs/persistentstorage.md
troian: both are valid examples since we want to have backward compatibility with previous SDL

_docs/persistentstorage.md
troian: this is set by the tenant. by default the volume is ephemeral

_docs/persistentstorage.md
```diff
@@ -0,0 +1,305 @@
...
+      count: 1
+```
+
+## CEPH/ROOK
+
+### Teardown
+
+#### Zapping devices
+
+Disks on nodes used by Rook for osds must be reset to a usable (aka blank) state with the [following methods](https://github.com/rook/rook/blob/master/Documentation/ceph-teardown.md#zapping-devices)
```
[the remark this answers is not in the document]
troian: the entire section of provider is WIP

PR #1499, opened 2022-01-21
WIP: IP address marketplace via metal LB

hydrogen18:
This is the current state of my IP address market place work. There are likely some issues with this & some things that still need to be implemented, but at this time please take a look at it.

**High Level Changes**

1. Add Metal LB to our KIND test cluster
2. Expose the Metal LB metrics interface
3. Add Metal LB address pool to our KIND test cluster, using `24.0.0.0/8` network
4. Add an IP address operator
5. Add kustomize stuff for the ip address operator
6. Deploy the IP address operator in the `_run/single` example by default
7. Update SDL to allow specifying IP address endpoints
8. Update the bidding engine to take into account the IP address leases when computing a bid

**Other changes**

1. Deduplicate as much common logic between the hostname operator & IP address operator as possible
2. Move the operators into their own directory, independent of  `provider/cmd`
3. Add separate interfaces for the IP address operator stuff
4. Refactor the cluster/lease-status command to not act as a single point of everything. Stuff can be brought together inside the HTTP gateway code if a deployment has a bunch of different aspects to be reported on (for example, leased IPs)
5. Update inventory code to run a teardown if a deploy fails. This should act as a stopgap if we're adding resources and something goes wrong 1/2 way through.

**Unimplemented things**

1. Restricting the IP operator to a specific IP pool. Right now anything in metal LB is fair game
2. Add the ability to migrate IPs between deployments on the same provider
3. Add the ability to standup a deployment without the IPs currently available, since they could be migrated over in the future from existing deployments

Example new SDL:

```
---
version: "2.0"

services:
  web:
    image: quay.io/ovrclk/demo-app
    expose:
      - port: 80
        as: 80
        accept:
          - hello.localhost
        to:
          - global: true
            ip: "foobar"
  bew:
    image: quay.io/ovrclk/demo-app
    expose:
      - port: 80
        as: 81
        to:
          - global: true
            ip: "foobar"

profiles:
  compute:
    web:
      resources:
        cpu:
          units: 0.1
        memory:
          size: 16Mi
        storage:
          size: 128Mi
  placement:
    westcoast:
      attributes:
        region: us-west
      pricing:
        web: 
          denom: uakt
          amount: 1000

deployment:
  web:
    westcoast:
      profile: web
      count: 1
  bew:
    westcoast:
      profile: web
      count: 1
endpoints:
  foobar:
    kind: "ip"
```


figurestudios:
Just a little suggestion: Would it be possible to collaborate with Sentinel on this marketplace?

hydrogen18:
@figurestudios Thanks for mentioning it. We've talked to a few users about running Sentinel on Akash and this is one of the features we've identified

_run/kube/deployment.yaml
```diff
@@ -52,4 +53,7 @@ deployment:
   bew:
     westcoast:
       profile: web
-      count: 1
\ No newline at end of file
+      count: 1
+endpoints:
+  foobario:
```
boz, line 58: _would_ `foobar.io` be a legal key?

_run/metallb-service.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: v1
```
boz, line 1: is this in `_run/` because the `_docs/kustomize` versions won't work with our `kind` environment?

how about adding an explanation to `_run/README.md` for these?

_run/single/Makefile
```diff
@@ -45,10 +52,14 @@ kustomize-install-node:
 kustomize-install-provider:
 	kubectl kustomize kustomize/akash-provider | kubectl apply -f-
 
-.PHONY: kustmoize-install-hostname-operator
```
boz, line 48: :+1:

manifest/v2beta1/types.go
```diff
@@ -81,13 +82,15 @@ func (s Service) GetCount() uint32 {
...
-	HTTPOptions  ServiceExposeHTTPOptions
+	Port                   uint16 // Port on the container
+	ExternalPort           uint16 // Port on the service definition
+	Proto                  ServiceProtocol
+	Service                string
+	Global                 bool
+	Hosts                  []string
+	HTTPOptions            ServiceExposeHTTPOptions
+	IP                     string
+	EndpointSequenceNumber uint32
```
boz, line 93: comments explaining these would be good.

pkg/apis/akash.network/crd.yaml
```diff
@@ -183,3 +188,62 @@ spec:
...
+                service_name:
+                  type: string
+                port:
+                  type: integer
+                external_port:
+                  type: integer
+                protocol:
+                  type: string
+                sharing_key:
+                  type: string
```
boz, line 238: OT: it'd be great if we could re-use schemas - have `LeaseID` defined in one place for instance.  Not sure if it's possible.

provider/bidengine/order.go
```diff
@@ -203,6 +203,7 @@ loop:
 			err := queryBid.Error()
 			bidFound := true
 			if err != nil {
+				// Use super-advanced technique for detecting if bid is not on blockchain
```
boz, line 206: :'(

provider/cluster/inventory.go
```diff
@@ -490,15 +555,17 @@ loop:
...
 			default:
 				is.log.Debug("inventory ready")
 				close(is.readych)
 			}
 
-			inventory = res.Value().(ctypes.Inventory)
-			metrics := inventory.Metrics()
+			resultArray := res.Value().([]interface{})
+
+			state.inventory = resultArray[0].(ctypes.Inventory)
```
boz, line 567: wen generics

boz: I feel like it should check `len(resultsArray)` and also use the two-lhs form of casting.  Any reason to not?

provider/cluster/inventory.go
```diff
@@ -327,21 +343,89 @@ func updateReservationMetrics(reservations []*reservation) {
...
+	state.reservations = append(state.reservations, reservation)
+	req.ch <- inventoryResponse{value: reservation}
+	inventoryRequestsCounter.WithLabelValues("reserve", "create").Inc()
+
+}
+
+func (is *inventoryService) run(ctx context.Context, reservationsArg []*reservation) {
 	defer is.lc.ShutdownCompleted()
 	defer is.sub.Close()
-	ctx, cancel := context.WithCancel(context.Background())
```
boz, line 333: :+1:

boz (approved):
Great work!

Doc changes that we'll need to make:
* new options (pricing, ip configuration)
* ip marketplace design
* deploying provider

deitch:
> Restricting the IP operator to a specific IP pool. Right now anything in metal LB is fair game

If it helps, look at what I did in the [Packet CCM](https://github.com/equinix/cloud-provider-equinix-metal) to manage the IPs. They are allocated via the ~Packet~ Equinix Metal API, and then the CCM controller hands them off to metallb. If it is relevant, I can link to the exact locations.

deitch:
> Doc changes that we'll need to make:

Yeah, both public usage and the architecture docs. Do we have a write-up somewhere on what this does and how it works, that I could use as a jumping off point?

boz:
> > Doc changes that we'll need to make:
> 
> Yeah, both public usage and the architecture docs. Do we have a write-up somewhere on what this does and how it works, that I could use as a jumping off point?

@hydrogen18 has one I believe.

_run/kube/deployment.yaml
hydrogen18: Probably, I can check. Do we want to allow names like that and other obscure things like "-_!" ?

_run/metallb-service.yaml
hydrogen18: I put it here because it is at this point only used to support the `_run/` examples. But I don't see any reason why this wouldn't work in a real provider on mainnet. @sacreman do we have a recommended way to add an optional component to provider deployments like this?

provider/cluster/inventory.go
hydrogen18: I can double check this, I think there was some reason I wound up casting this here.

provider/cluster/operatorclients/ip_operator_client.go
```diff
@@ -0,0 +1,195 @@
...
+
+	if response.StatusCode != http.StatusOK {
+		return errNotAlive
+	}
+
+	return nil
+}
+
+func (ipoc *ipOperatorClient) newRequest(ctx context.Context, method string, path string, body io.Reader) (*http.Request, error) {
+	if ipoc.client == nil {
```
hydrogen18, line 110: I think there is a race here around initializing `ipoc.client`, but I can probably just fix it with a mutex.

_run/kube/deployment.yaml
boz: `-` and `_` would be nice.

provider/operator/ipoperator/ip_operator.go
```diff
@@ -0,0 +1,788 @@
...
+			}
+		}
+	}
+
+	op.providerSda.Stop()
+	op.mllbc.Stop()
+	return parentCtx.Err()
+}
+
+type barrier struct {
```
hydrogen18, line 746: I think this type has a race in it, but I am not sure. Need to double check.

hydrogen18: It looks like storing a value, when I meant to subtract that value. This has been corrected.

provider/cluster/inventory.go
hydrogen18: I can just add some sort of local type here that gets returned as the result I think

hydrogen18:
I've addressed the comments here and will open another PR when this is complete.

PR #1506, opened 2022-01-26
docs(storage): persistent storage deploy/teardown

troian:


dmikey (approved):

_docs/persistenstorage-provider.md
```diff
@@ -0,0 +1,103 @@
+# Lets begin
+
+At Akash we use Rook/Ceph couple to provider dynamic volumes provisioning
+Ceph - distributed filesystem
```
hydrogen18, line 4: Make this an bulleted list

_docs/persistenstorage-provider.md
```diff
@@ -0,0 +1,103 @@
...
+3. We will prevent cluster from using master nodes.
+4. We will be utilizing all available devices on each used node
+
+## Configuring
+In this example we are focusing on making slight adjustments to the `cluster.yaml` only. All other files must remain unchanged.
+
+1. get list of nodes in the cluster
+```shell
+kubectl get nodes -ojson | jq -r '.items[].metadata.labels."kubernetes.io/hostname"'
+k8s-master.edgenet-1.ewr1 # this node won't be used
```
hydrogen18, line 24: We should explain why this node is not used, not just say it isn't used. For example some people may run multi master kube clusters and would want to exclude all masters.

_docs/persistenstorage-provider.md
```diff
@@ -0,0 +1,103 @@
...
+```yaml
+nodes:
+  - name: k8s-node-0.edgenet-1.ewr1
+  - name: k8s-node-1.edgenet-1.ewr1
+```
+
+## Deploying test cluster
+```shell
+# deploy takes around 10m
+ROOK_PATH=$(AKASH_ROOT)/_docs/rook/test ./scripts/rook.sh deploy
```
hydrogen18, line 39: This tries to run a subshell running a command `AKASH_ROOT` in bash. What shell is this intended for?

_docs/persistenstorage-provider.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Teardown
+
+```shell
+ROOK_PATH=$(AKASH_ROOT)/_docs/rook/test ./scripts/rook.sh teardown
+```
+
+### Zapping Devices
+
+Disks on nodes used by Rook for osds can be reset to a usable state with the following method:
```
hydrogen18, line 62: What is an `osd`?

_docs/persistenstorage-provider.md
```diff
@@ -0,0 +1,103 @@
...
+#!/usr/bin/env bash
+
+if [ "$#" -ne 1 ]; then
+	echo "Illegal number of parameters"
+fi
+
+# Zap the disk to a fresh, usable state (zap-all is important, b/c MBR has to be clean)
+IFS=","
+for disk in ${$1}; do
+	dev=/dev/${disk}
```
hydrogen18, line 81: This seems to be overly complex? As far as I can tell it 

1. Runs `sgdisk` which removes GPT & MBR
2. Wipes the first 100 MB of disk, which definitely removes GPT & MBR
3. OR it will run blkdiscard which wipes the entire disk

Can we just run one approach here? `dd` is highly reliable for this and available on basically every linux distribution for a long time.

_docs/persistenstorage-provider.md
troian: typo

_docs/persistenstorage-provider.md
troian: The dd approach will mark all sectors on the SSD as ‘used’ as it writes zeros across the disk. This interferes with the ability of the SSD to perform sector reallocations, native trim and wear levelling, as all sectors are now deemed occupied. blkdiscard on an SSD only marks the sectors as unused so that they are treated as all zeros

hydrogen18: While correct, if overwriting the first 100MB of a disk is an actual issue then we have a huge problem in production where the disk could be completely overwritten by one lease, then overwritten again by another lease, etc. I am sure lease owners will be writing more than 100MB at a time.

Can we not configure one of the pieces of storage management middleware (rook, ceph, whatever) to tell the hardware to perform the discard operation?

boz (approved):
Looks good.

I'd like for this to be incorporated into https://github.com/ovrclk/docs-staging .

Can make a separate issue for it.

PR #1512, opened 2022-02-02
refactor: use Services api to check if lease is up

troian:

### DIFF
--- a/provider/cluster/kube/client_exec.go
+++ b/provider/cluster/kube/client_exec.go
@@ -53,52 +53,26 @@ func (c *client) Exec(ctx context.Context, leaseID mtypes.LeaseID, serviceName s
 	tsq remotecommand.TerminalSizeQueue) (ctypes.ExecResult, error) {
 	namespace := builder.LidNS(leaseID)
 
-	// Check that the deployment exists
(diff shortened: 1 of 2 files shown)


provider/cluster/kube/client_exec.go
```diff
@@ -53,52 +55,19 @@ func (c *client) Exec(ctx context.Context, leaseID mtypes.LeaseID, serviceName s
...
-		FieldSelector:        "",
-		Watch:                false,
-		AllowWatchBookmarks:  false,
-		ResourceVersion:      "",
-		ResourceVersionMatch: "",
-		TimeoutSeconds:       nil,
-		Limit:                0,
-		Continue:             "",
-	})
+	_, err := c.kc.CoreV1().Services(namespace).Get(ctx, serviceName, metav1.GetOptions{})
```
hydrogen18, line 58: We can just drop this query & check, since the service is basically always going to exist. If the pod below isn't present I think an appropriate error message comes back already.

hydrogen18 (approved):

provider/cluster/kube/client_exec.go
troian: removed

dmikey (approved):

PR #1515, opened 2022-02-07
feat: Add resource server to server loki logs

arijitAD:
## Description

This PR adds a resource server that authenticates all incoming requests using the JWT provided in `Authorization` header. This server, then routes the request to appropriate handlers.
At present, this server has one lease specific handler named `/loki-service` that forwards all incoming requests to serve lease logs to a multi-tenant loki instance running in provider's cluster. For example, if the incoming request is:
```
http://{RESOURCE_SERVER_ADDR}/lease/1/1/1/loki-service/loki/api/v1/query?query={app=".+"}
```
Then this request will be forwarded to the loki instance like this:
```
http://{LOKI_GATEWAY_ADDR}/loki/api/v1/query?query={app=".+"}
```
i.e., the part after the `/loki-service` fragment in incoming URL is what will be used for routing to loki. So, any loki API can be used in the input URL. In the example above, the loki API being used is:
```
/loki/api/v1/query?query={app=".+"}
```
The forwarded request will have the `X-Scope-OrgID` header set to the right namespace based on the tenant's address (found using the Auhorization JWT), the provider's address, and the `/{dseq}/{gseq}/{oseq}` provided in the URL (`/1/1/1` part in the example). This way a tenant can see only their logs.

If it is required that only a specific subset of loki APIs be exposed to the tenants, then a NGINX ingress can be put in-between the loki instance and the resource server with API whitelisting. `LOKI_GATEWAY_ADDR` will then point to the address of the ingress.

### References
* All the APIs provided by loki: https://grafana.com/docs/loki/latest/api/

## Motivation and Context

## Types of changes

## Checklist:


provider/gateway/rest/router.go
```diff
@@ -178,6 +183,86 @@ func newJwtServerRouter(addr sdk.Address, privateKey interface{}, jwtExpiresAfte
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		// set the X-Scope-OrgID header for fetching logs for the right tenant
+		r.Header.Set("X-Scope-OrgID", builder.LidNS(requestLeaseID(r)))
+
+		// build target url for the reverse proxy
+		scheme := "http" // for http & https
+		if strings.HasPrefix(r.URL.Scheme, "ws") {
+			scheme = "ws" // for ws & wss
+		}
+		lokiURL, err := url.Parse(fmt.Sprintf("%s://%s", scheme, lokiGwAddr))
```
hydrogen18, line 242: We should encapsulate all requests to loki into a client interface that can be defined in `provider/cluster/kube/loki`. That way it is easier to stub out in test, easier to establish a consist pattern of interaction with the 3rd party service, and easier to refactor in the future if we need to.

provider/gateway/rest/server.go
```diff
@@ -67,3 +69,21 @@ func NewJwtServer(ctx context.Context,
...
+	serverAddr string,
+	providerAddr sdk.Address,
+	cert *x509.Certificate,
+	lokiGwAddr string,
+) (*http.Server, error) {
+	srv := &http.Server{
+		Addr:    serverAddr,
+		Handler: newResourceServerRouter(log, providerAddr, cert.PublicKey.(*ecdsa.PublicKey), lokiGwAddr),
+		BaseContext: func(_ net.Listener) context.Context {
+			return ctx
```
hydrogen18, line 84: I think we can make this one line like `func(_ net.Listener) context.Context { return ctx }`

provider/gateway/rest/server.go
```diff
@@ -67,3 +69,21 @@ func NewJwtServer(ctx context.Context,
...
+func NewResourceServer(ctx context.Context,
+	log log.Logger,
+	serverAddr string,
+	providerAddr sdk.Address,
+	cert *x509.Certificate,
+	lokiGwAddr string,
+) (*http.Server, error) {
+	srv := &http.Server{
+		Addr:    serverAddr,
+		Handler: newResourceServerRouter(log, providerAddr, cert.PublicKey.(*ecdsa.PublicKey), lokiGwAddr),
```
hydrogen18, line 82: Why are we casting uncondtionally on this line to a `*ecdsa.PublicKey` ?

provider/gateway/rest/router.go
```diff
@@ -178,6 +183,86 @@ func newJwtServerRouter(addr sdk.Address, privateKey interface{}, jwtExpiresAfte
 	return router
 }
 
+func newResourceServerRouter(log log.Logger, providerAddr sdk.Address, publicKey *ecdsa.PublicKey, lokiGwAddr string) *mux.Router {
+	router := mux.NewRouter()
+
+	// add a middleware to verify the JWT provided in Authorization header
+	router.Use(func(next http.Handler) http.Handler {
```
hydrogen18, line 190: Can we make this middleware its own function in its own file somewhere? It's easier to read, test & debug that way

provider/gateway/rest/router.go
```diff
@@ -178,6 +183,86 @@ func newJwtServerRouter(addr sdk.Address, privateKey interface{}, jwtExpiresAfte
...
+			token, err := jwt.ParseWithClaims(r.Header.Get("Authorization"), &ClientCustomClaims{}, func(token *jwt.Token) (interface{}, error) {
+				// return the public key to be used for JWT verification
+				return publicKey, nil
+			})
+			if err != nil {
+				log.Error(err.Error())
+				http.Error(w, err.Error(), http.StatusUnauthorized)
+				return
+			}
+			// delete the Authorization header as it is no more needed
```
hydrogen18, line 202: Why is this needed?

provider/gateway/rest/router.go
```diff
@@ -178,6 +183,86 @@ func newJwtServerRouter(addr sdk.Address, privateKey interface{}, jwtExpiresAfte
...
+			r.Header.Del("Authorization")
+
+			// store the owner & provider address in request context to be used in later handlers
+			ownerAddress, err := sdk.AccAddressFromBech32(token.Claims.(*ClientCustomClaims).Subject)
+			if err != nil {
+				log.Error(err.Error())
```
hydrogen18, line 208: If possible just log something simple in cases like this 

```
log.Error("failed parsing owner address", "error", err, "address", address)
```

So that this is easier to debug. Usually the way we'll encounter this is someone like @sacreman reporting the error in the logs and we might have to figure out what is going on.

provider/gateway/rest/router.go
```diff
@@ -178,6 +183,86 @@ func newJwtServerRouter(addr sdk.Address, privateKey interface{}, jwtExpiresAfte
...
+				http.Error(w, err.Error(), http.StatusUnauthorized)
+				return
+			}
+			// delete the Authorization header as it is no more needed
+			r.Header.Del("Authorization")
+
+			// store the owner & provider address in request context to be used in later handlers
+			ownerAddress, err := sdk.AccAddressFromBech32(token.Claims.(*ClientCustomClaims).Subject)
```
hydrogen18, line 206: Why is this being cast here?

provider/gateway/rest/server.go
arijitAD: Done

provider/gateway/rest/server.go
arijitAD: Fixed

provider/gateway/rest/router.go
arijitAD: Done

provider/gateway/rest/router.go
arijitAD: Done

provider/gateway/rest/router.go
arijitAD: Done

provider/gateway/rest/router.go
arijitAD: If it is required that only a specific subset of loki APIs be exposed to the tenants, then a NGINX ingress can be put in-between the loki instance and the resource server with API whitelisting. LOKI_GATEWAY_ADDR will then point to the address of the ingress.

I think, this would be a better approach, than to define the loki interface in code as we can't control the loki version that provider's would use. And, each version can have different APIs exposed. So, better to let the provider's make this decision on their own.

boz: I'm ok with using nginx to make the API simpler, but we will have to document it well so that people don't run with it wide open by default.

Something we can address later.

provider/gateway/rest/router.go
```diff
@@ -178,6 +183,59 @@ func newJwtServerRouter(addr sdk.Address, privateKey interface{}, jwtExpiresAfte
...
+		}
+		reverseProxy := httputil.NewSingleHostReverseProxy(lokiURL)
+
+		// remove the "/lease/{dseq}/{gseq}/{oseq}/loki-service" path prefix from the request url
+		// before it is sent to the reverse proxy.
+		pathSplits := strings.SplitN(r.URL.Path, "/", 7)
+		if len(pathSplits) < 7 || pathSplits[6] == "" {
+			log.Error("loki api not provided in url")
+			http.Error(w, "loki api not provided in url", http.StatusBadRequest)
+			return
```
boz, line 229: small nitpick that it'd be good to do this before anything else (creating `reverseProxy`, etc...), and it'd be good to have it in a utility function that is testable.

boz (approved):
Looks good, thank you @arijitAD

boz:
@arijitAD this looks great - do you have any pending work for it?  Do you want us to merge it?

arijitAD:
> @arijitAD this looks great - do you have any pending work for it? Do you want us to merge it?

This is done. I will merge it.

PR #1518, opened 2022-02-11
fix(provider): If a pod is not ready, do not try and launch a shell

hydrogen18:
In the implementation for the provider lease shell code that talks to kubernetes, just check that the pod is actually ready before launching a shell. If not, give the user a clear error message as to what is going on here.

I also added a bunch of tests for this code since it turned out to be to relatively easy.


boz (approved):
I don't think we should have the hand-rolled log de-duplication in there.

Approving since it's in tests.

PR #1542, opened 2022-03-21
provider startup: skip active lease query

boz:
* skip problematic active lease query
* backport manifest deletion on lease close
* add docker startup helper

TODO:

[ ] guard against creating namespaces for closed leases


_build/bootstrap.sh
```diff
@@ -0,0 +1,51 @@
...
+apt-get install jq curl -y
+
+DEFAULT_META_URL=https://raw.githubusercontent.com/ovrclk/net/master/mainnet/meta.json
+META=./meta.json
+
+meta_url="${1:-$DEFAULT_META_URL}"
+
+curl -sSfl "$meta_url" > "$META"
+
+export AKASH_HOME="${AKASH_HOME:-/akash}"
```
troian, line 31: i think default dir is `.akash`

_build/bootstrap.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
```
hydrogen18, line 2: strict mode make sense here? 

http://redsymbol.net/articles/unofficial-bash-strict-mode/ 

I don't think we are expecting commands to fail in this.

troian: `set -e` definitely makes sense if `curl` fails

boz: agreed, thanks guys.

_build/bootstrap.sh
boz: but I hate that :joy:

boz: used `--home` because you never know :roll_eyes:

provider/cluster/manager.go
```diff
@@ -312,6 +314,14 @@ func (dm *deploymentManager) doDeploy() ([]string, error) {
 		case <-ctx.Done():
 		}
 	}()
+
+	if err = dm.checkLeaseActive(ctx); err != nil {
```
boz, line 318: added active lease guard here @troian @hydrogen18

boz: I took out the previous guard (`ActiveLeasesForProvider`) so I wanted to replace it with something here.  I accidentally launched 15,000 namespaces without it (didn't have manifest pruning).

Even with namespace pruning, offhand I can think of:

1. provider wins lease
2. provider receives manifest
3. provider shuts down
4. lease is closed by client
5. provider starts up

I agree that a vacuum process (handled by `orphan-ns-purge` script now) does make sense, but I figure it's not too much to do a query before creating a ns just in case as well.

hydrogen18 (approved):

provider/cluster/manager.go
boz: Now I'm second guessing this - maybe it should just be a re-implementation of `ActiveLeasesForProvider`.  Worried about a torrent of parallel queries on startup.  At least the retry has jitter? :man_shrugging:

hydrogen18: Is this due to the lack of a secondary index? Can we just query deployment first (by deploymentID) and if it is closed then we're done?

boz: Yeah, lack of secondary index made [`ActiveLeasesForProvider()`](https://github.com/ovrclk/akash/blob/13abbb6c4d88d93bf95c4b313fecf9818c0e336c/client/extra.go#L13-L30)  unuseable.  We can actually just query the lease by id, which is what we do in this PR.

hydrogen18: I don't think the place of this check really matters at this time. Also, in this change you've moved it to its own goroutine (from what I can see) so this should be fine.

hydrogen18: So at this point it seems we should really add a secondary index as part of the next upgrade? I'm not seeing an easy workaround to this presently.

boz: I think it's too late to add a secondary index, and I'm hesitant to in the first place.

* this is covering a corner-case - lease open but no manifest yet when restarting provider
* it can be solved in other ways
  * manifest server can lookup leases when a manifest for an unknown lease is received
  * active leases can be cached locally in etcd or similar

I might actually be ok with removing this second `ActiveLeasesForProvider()` invocation completely and living with the corner case not being handled for the time being.

On the other hand, having only one invocation of this seems to fix the problem in our larger clusters.

hydrogen18: I suggest we go ahead and remove it. It's a corner case at this time.

boz:
Reopening this - the final `ActiveLeasesForProvider` is preventing startup.

boz:
superseded by #1560

PR #1547, opened 2022-03-25
unset akash gas price env. var. on init targets

hydrogen18:
In the latest RC & on master I can't do `make init` on the `_run/kube` example because it tries to connect to an RPC node & fails.

This change disables the Akash gas prices config. on the `init` target, where it is no longer needed. At that point I can get `make init` to work again.


troian:
it looks like a bug in cosmos sdk. it should not go for node talk

_run/common.mk
```diff
@@ -39,6 +39,7 @@ SERVER_CERTS := provider
 
 
 .PHONY: init
+init: AKASH_GAS='' AKASH_GAS_PRICES=''
```
troian, line 42: `node-init-gentx` seems to be better place

`node-init-gentx: AKASH_GAS='' AKASH_GAS_PRICES=''`

hydrogen18: Why? There isn't any place during the `init` step where we should need to compute gas or fees.

boz: it's probably estimating gas by sending a simulation transaction

troian: fails [here](https://github.com/ovrclk/akash/blob/9671afe7f55748473cf3cb5230aea23f8727366a/_run/common.mk#L104)

troian (approved):

PR #1553, opened 2022-04-08
simpler plaintext export & import of keys

hydrogen18:
The existing export / import functionality for keys serves end-users well. For programmatic purposes (like running a provider in a docker container) it's painful. The default export format is a PEM file which contains whitespace & is encrypted.

This adds an `akash keys export-packed` and `akash keys import-packed` functionality that can import a specific set of keys or the entire wallet. The output is a single line without whitespace that is not encrypted.

For example:

```
$ akash keys export-packed bob
PT,bob,DHkoAZCbPKi1IC2fEH4FMTPl4IA8tcCqJyMBkJDQhgc
```

This `import-packed` command takes this string as either input on stdin or as a single command line argument. It won't overwrite keys by default, but has an option for that.

If you're using a ledger or OS keyring you would get prompted to unlock the device. But this isn't really a concern because this isn't meant for end users. I've hidden these commands in the help because we don't want end users accidentally winding up using them and exporting their keys to plaintext.

### DIFF
--- a/cmd/akash/cmd/export_import_packed_keys.go
+++ b/cmd/akash/cmd/export_import_packed_keys.go
@@ -1,0 +1,258 @@
+package cmd
+
+import (
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/crypto/hd"
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
+	"io"
+	"strings"
+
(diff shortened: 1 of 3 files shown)


boz (commented):
Can this not be done by `akash keys export/import` + some shell scripting?

hydrogen18:
No it can't because there is no way to export a key in a format that isn't encrypted.

boz:
can it not use a password?

hydrogen18:
Well, Steve somehow put together a solution that takes a PEM file and a password as an input to a helm chart. But it requires all sorts of odd workarounds because the CLI likes to do things like prompt the user for input on standard in.

boz:
```sh
(echo password; echo password) | akash keys import key.pem
```
should work fine.

PR #1559, opened 2022-04-20
fix: set DefaultBidMinDeposit same value as mainnet-2

troian:
allows genereting test genesis with params identical to
the mainnet


hydrogen18:
Is this necessary or can we just edit the genesis to whatever we want when we make it? Otherwise it seems like we'd keep updating this code after each proposal

boz (approved):
:rocket:

PR #1560, opened 2022-04-22
fix(provider): don't query leases on startup

boz:
`ActiveLeasesForProvider` never terminates; guard
against instantiating closed deployments instead

supersedes #1542


boz:
There is an issue here where when a manifest is submitted, it is assumed that the group data is present.

boz:
never mind, it "just works" - requests are queued until the data is fetched.

provider/cluster/service.go
```diff
@@ -360,37 +359,11 @@ func (s *service) teardownLease(lid mtypes.LeaseID) {
 	}
 }
 
-func findDeployments(ctx context.Context, log log.Logger, client Client, session session.Session) ([]ctypes.Deployment, error) {
+func findDeployments(ctx context.Context, log log.Logger, client Client, _ session.Session) ([]ctypes.Deployment, error) {
```
hydrogen18, line 362: Can't we just remove this method?

boz: yeah

boz: I kept it as a placeholder for something that looks in CRDs/etcd but I can remove it.

hydrogen18: OK, I think the same goes for the other method below

provider/cluster/manager.go
```diff
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/boz/go-lifecycle"
+	"github.com/pkg/errors"
```
hydrogen18, line 10: Just import `"errors"`.

dmikey (approved):

hydrogen18 (approved):

PR #1569, opened 2022-04-28
build(lint): install and use golangci-lint from cache

troian:
fixes #ENG-286
speeds up lint process
replace deprecated linters


hydrogen18:
This doesn't use docker anymore to run golang CI?

PR #1576, opened 2022-05-10
feat: monitor lease funds

troian:
fixes #ENG-283


client/client.go
```diff
@@ -97,6 +101,22 @@ func (c *client) Query() QueryClient {
...
+	}
+
+	status, err := node.Status(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	info := status.SyncInfo
+
+	return &info, nil
```
hydrogen18, line 117: Return this by value? I can't see why it needs to be a pointer.

provider/balance_checker.go
```diff
@@ -83,6 +102,82 @@ func (bc *balanceChecker) doCheck(ctx context.Context) (bool, error) {
...
+		lid: lid,
+	}
+
+	var syncInfo *tmrpc.SyncInfo
+	syncInfo, resp.err = bc.session.Client().NodeSyncInfo(ctx)
+	if resp.err != nil {
+		return resp
+	}
+
+	if !syncInfo.CatchingUp {
```
hydrogen18, line 125: This check seems backwards? If the node is catching up (set to true), we would want to skip everything

client/client.go
troian: main reason is mocks. need to simulate block height is going up
```
	client.On("NodeSyncInfo", mock.Anything).Run(func(args mock.Arguments) {
 		nodeSyncInfo.LatestBlockHeight = int64(time.Now().Sub(startedAt) / blockPeriod)
 	}).Return(nodeSyncInfo, nil)
```

provider/balance_checker.go
```diff
@@ -113,7 +211,58 @@ loop:
...
+				bc.leases[res.lid] = time.AfterFunc(time.Minute, func() {
+					select {
+					case <-bc.lc.Done():
+					case leaseCheckCh <- bc.doEscrowCheck(ctx, res.lid):
+					}
+				})
+				break
+			}
+
+			if res.checkAfter == 0 {
```
hydrogen18, line 250: can we just add a function receiver to the response type like "res.isOutOfFunds()" so this is easier to identify? It's checking a duration for a value at this time

provider/balance_checker.go
```diff
@@ -83,6 +102,82 @@ func (bc *balanceChecker) doCheck(ctx context.Context) (bool, error) {
...
+		if blocksRemaining <= 0 {
+			return resp
+		}
+
+		blocksPerCheckInterval := int64(bc.cfg.LeaseFundCheckInterval / blockPeriod)
+		if blocksRemaining > blocksPerCheckInterval {
+			blocksRemaining = blocksPerCheckInterval
+		}
+
+		resp.checkAfter = time.Duration(blocksRemaining) * (6500 * time.Millisecond)
```
hydrogen18, line 175: We should add a constant for estimated block time somewhere, if we don't have that already

troian: yeah, it is right above

provider/balance_checker.go
```diff
@@ -113,7 +211,58 @@ loop:
 			balanceCheckResult = runner.Do(func() runner.Result {
 				return runner.NewResult(bc.doCheck(ctx))
 			})
+		case evt := <-subscriber.Events():
+			switch ev := evt.(type) {
+			case event.LeaseAddFundsMonitor:
+				bc.leases[ev.LeaseID] = nil
+				if bc.cfg.LeaseFundCheckInterval > 0 {
+					bc.runEscrowCheck(ctx, ev.LeaseID, leaseCheckCh)
```
hydrogen18, line 219: This seems to start a goroutine and then throw away the result. What is the utility in doing this? What side effect happens that is useful?

provider/balance_checker.go
```diff
@@ -113,7 +211,58 @@ loop:
...
+					}
+				})
+				break
+			}
+
+			if res.checkAfter == 0 {
+				bc.log.Debug("lease is out of funds. closing", "leaseId", res.lid)
+
+				_ = runner.Do(func() runner.Result {
+					return runner.NewResult(nil, bc.session.Client().Tx().Broadcast(ctx, mtypes.NewMsgCloseBid(res.lid.BidID())))
```
hydrogen18, line 254: This is a bad idea because if we get our logic wrong we'll start closing bids too early. Just run the withdrawal transaction at this time. The network closes things if it is actually overdrawn.

provider/cluster/monitor.go
```diff
@@ -36,6 +36,9 @@ var (
 	}, []string{"state"})
 )
 
+type EscrowQuerier interface {
```
hydrogen18, line 39: Do we need this?

provider/cluster/service.go
```diff
@@ -259,6 +259,7 @@ func (s *service) run(deployments []ctypes.Deployment) {
 		key := deployment.LeaseID()
 		mgroup := deployment.ManifestGroup()
 		s.managers[key] = newDeploymentManager(s, deployment.LeaseID(), &mgroup)
+		_ = s.bus.Publish(event.LeaseAddFundsMonitor{LeaseID: deployment.LeaseID()})
```
hydrogen18, line 262: Ignoring this error seems like a bad idea? Just log it if something happens, so we can track down what is failing.

provider/cmd/run.go
```diff
@@ -104,6 +105,13 @@ func RunCmd() *cobra.Command {
 		Use:          "run",
 		Short:        "run akash provider",
 		SilenceUsage: true,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			leaseFundsMonInterval := viper.GetDuration(FlagLeaseFundsMonitorInterval)
+			if leaseFundsMonInterval > 0 && leaseFundsMonInterval < time.Minute {
+				return fmt.Errorf("\"lease-funds-monitor-interval\" contains invalid value. expected 0 or >=1m")
```
hydrogen18, line 111: Triple quotes here should remove the need for escapes, and you can use `errors.New` since it doesn't look like you are formatting anything.

provider/balance_checker.go
hydrogen18: Isn't `6500 * time.Millisecond` in this line some sort of rough estimate of block time?

hydrogen18:
Since we're monitoring the escrow account now, we should remove the balance checking logic. It's been problematic (burning up AKT in the form of fees) in any case. I don't really see the utility in it any longer at this time. If someone needs it we can build up an external project to monitor provider balance for being too low.

provider/balance_checker.go
```diff
@@ -113,7 +211,58 @@ loop:
...
+				})
+
+				break
+			}
 
+			bc.leases[res.lid] = time.AfterFunc(res.checkAfter, func() {
```
hydrogen18, line 260: I think it is going to be simpler to just start up a goroutine per lease that is long-lived (until the lease closes) and then have it check periodically. We can tie those into a lifecycle object like we do with other things.

provider/balance_checker_test.go
```diff
@@ -146,5 +232,43 @@ func TestBalanceCheckerStartsWithdrawal(t *testing.T) {
...
+	defer testScaffold.testBus.Close()
+
+	subscriber, err := testScaffold.testBus.Subscribe()
+	require.NoError(t, err)
+	defer subscriber.Close()
+
+	lid := mtypes.LeaseID{
+		Owner:    bc.ownAddr.String(),
+		DSeq:     1,
+		GSeq:     0,
```
hydrogen18, line 249: I think these are normally >1 in the real network? 

I usually use a `testutil` function from our package, then overwrite the address fields.

provider/balance_checker_test.go
```diff
@@ -146,5 +232,43 @@ func TestBalanceCheckerStartsWithdrawal(t *testing.T) {
...
+
+	err = testScaffold.testBus.Publish(event.LeaseAddFundsMonitor{LeaseID: lid})
+	require.NoError(t, err)
+
+	select {
+	case msg := <-testScaffold.broadcast.txCh:
+		switch msg.(type) {
+		case *mtypes.MsgCloseBid:
+		default:
+			t.Errorf("received unexpected message")
```
hydrogen18, line 264: I think this is meant to say "did not receive expected message" since it is in a default case which gets hit if the type does not match the other cases.

provider/cluster/monitor.go
troian: ✂️

provider/balance_checker_test.go
troian: so message has been received but not of the type we expected.
`did not receive expected message` fits more outside of the switch.

`require.IsType` is more suitable anyways

provider/balance_checker_test.go
troian: gseq/oseq are 0 by default. removing them

provider/balance_checker.go
troian: indeed, ideally, we want to have real block time

boz: there's a `blockPeriod` constant defined above.

I agree @hydrogen18 but I'm also ok with leaving it as a constant right now.

provider/balance_checker.go
```diff
@@ -83,6 +102,82 @@ func (bc *balanceChecker) doCheck(ctx context.Context) (bool, error) {
...
+	if resp.err != nil {
+		return resp
+	}
+
+	totalLeaseAmount := sdk.NewDec(0)
+	for _, lease := range lResp.Leases {
+		totalLeaseAmount = totalLeaseAmount.Add(lease.Lease.Price.Amount)
+	}
+
+	if resp.err == nil {
```
boz, line 159: exit early

provider/balance_checker_test.go
```diff
@@ -2,44 +2,59 @@ package provider
...
+	aqc       *akashmock.QueryClient
+	bc        *balanceChecker
+	broadcast *broadcaster
 }
 
-func (s *scaffold) start() {
-	go s.bc.lc.WatchContext(s.ctx)
-	go s.bc.run()
+func (bc *broadcaster) Broadcast(_ context.Context, msg ...sdk.Msg) error {
+	bc.txCh <- msg[0]
```
boz, line 43: why is it variadic?

provider/balance_checker_test.go
```diff
@@ -146,5 +232,43 @@ func TestBalanceCheckerStartsWithdrawal(t *testing.T) {
...
+
+	lid := mtypes.LeaseID{
+		Owner:    bc.ownAddr.String(),
+		DSeq:     1,
+		GSeq:     0,
+		OSeq:     0,
+		Provider: bc.ownAddr.String(),
+	}
+
+	time.Sleep(time.Second)
```
boz, line 254: :hankey:

boz (commented):
Can we use common calculations here and in `akash query escrow blocks-remaining`?

provider/balance_checker_test.go
troian: https://github.com/ovrclk/akash/blob/e6e207f79ea812407f99ab89f3b75c299ee1f7b5/client/broadcaster/client.go#L22

provider/balance_checker_test.go
troian: advanced delay programming

provider/balance_checker.go
hydrogen18: Really, why don't we just use 5 seconds for the math here? It would just result in the check running consistently early in almost all cases. Which is fine.

provider/balance_checker.go
```diff
@@ -130,7 +279,6 @@ loop:
 				withdrawAllNow = true
 			}
 		case withdrawAll := <-withdrawAllResult:
-
 			withdrawAllResult = nil
 			withdrawalTicker.Reset(bc.cfg.PollingPeriod) // Re-enable the timer
```
hydrogen18, line 283: This line is a bug & we should remove it while we are in here.

provider/cluster/manager.go
```diff
@@ -155,10 +154,11 @@ func (dm *deploymentManager) run() {
 		dm.log.Debug("hostnames released")
 	}()
 
+	dm.startWithdrawal()
```
hydrogen18, line 157: It's hard to see why this moved. It looks like this starts immediately now. Can we put a comment explaining why this has to be started here?

provider/cmd/run.go
```diff
@@ -104,6 +105,20 @@ func RunCmd() *cobra.Command {
...
 		SilenceUsage: true,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			leaseFundsMonInterval := viper.GetDuration(FlagLeaseFundsMonitorInterval)
+			withdrawPeriod := viper.GetDuration(FlagWithdrawalPeriod)
+
+			if leaseFundsMonInterval < time.Minute || leaseFundsMonInterval > 24*time.Hour {
+				return errors.Errorf(`flag "%s" contains invalid value. expected >=1m<=24h`, FlagLeaseFundsMonitorInterval) // nolint: goerr113
+			}
+
+			if withdrawPeriod > 0 && withdrawPeriod < leaseFundsMonInterval {
```
hydrogen18, line 116: Why does this period need to be greater than the lease funds interval?

provider/cmd/run.go
```diff
@@ -579,9 +599,10 @@ func doRunCmd(ctx context.Context, cmd *cobra.Command, _ []string) error {
 	}
 
 	config.BalanceCheckerCfg = provider.BalanceCheckerConfig{
-		PollingPeriod:           viper.GetDuration(FlagBalanceCheckPeriod),
-		MinimumBalanceThreshold: viper.GetUint64(FlagMinimumBalance),
+		// PollingPeriod:           viper.GetDuration(FlagBalanceCheckPeriod),
```
hydrogen18, line 602: Remove these lines

provider/config.go
```diff
@@ -41,9 +40,9 @@ func NewDefaultConfig() Config {
 		ClusterWaitReadyDuration: time.Second * 10,
 		BidDeposit:               mtypes.DefaultBidMinDeposit,
 		BalanceCheckerCfg: BalanceCheckerConfig{
-			PollingPeriod:           5 * time.Minute,
-			MinimumBalanceThreshold: mparams.DefaultBidMinDeposit.Amount.Mul(sdk.NewIntFromUint64(2)).Uint64(),
-			WithdrawalPeriod:        24 * time.Hour,
+			// PollingPeriod:           5 * time.Minute,
```
hydrogen18, line 43: Remove these lines

provider/service.go
```diff
@@ -119,9 +121,16 @@ func NewService(ctx context.Context, cctx client.Context, accAddr sdk.AccAddress
 		return nil, err
 	}
 
-	bankQueryClient := bankTypes.NewQueryClient(cctx)
+	bc, err := newBalanceChecker(ctx, bankTypes.NewQueryClient(cctx), aclient.NewQueryClientFromCtx(cctx), accAddr, session, bus, cfg.BalanceCheckerCfg)
+	if err != nil {
+		session.Log().Error("starting balance checker", "err", err)
```
hydrogen18, line 126: should this log line say "creating balance checker failed" ?

x/escrow/client/cli/query.go
```diff
@@ -76,45 +73,45 @@ func cmdBlocksRemaining() *cobra.Command {
...
 			}
 
-			leases := make([]marketTypes.Lease, 0)
-			for _, lease := range leasesResponse.Leases {
-				leases = append(leases, lease.Lease)
+			if len(leasesResponse.Leases) == 0 {
+				return errNoLeaseMatches
 			}
 
+			// leases := make([]marketTypes.Lease, 0)
```
hydrogen18, line 80: remove these lines?

provider/event/events.go
```diff
@@ -53,5 +54,16 @@ type ClusterDeployment struct {
 	Status  ClusterDeploymentStatus
 }
 
-// Empty type used as a marker to indicate _any_ lease should be withdrawn now
-type LeaseWithdrawNow struct{}
+// LeaseWithdraw Empty type used as a marker to indicate specified lease should be withdrawn now
+type LeaseWithdraw struct {
```
hydrogen18, line 58: Might be a Github display thing, but where is this type created at? I see some code checking if it receives it, but not creating it.

hydrogen18: Nevermind, github decided not to show diffs by default now :(

provider/balance_checker.go
```diff
@@ -2,153 +2,316 @@ package provider
...
 
+	startCh := make(chan error, 1)
 	go bc.lc.WatchContext(ctx)
-	go bc.run()
+	go bc.run(startCh)
+
+	select {
+	case err := <-startCh:
+		if err != nil {
+			return nil, err
```
hydrogen18, line 93: Can be `return bc, err` without the `if`. But I'm not sure you need this.

provider/balance_checker.go
```diff
@@ -2,153 +2,316 @@ package provider
...
+			}
+		}
+	}()
 
-	var balanceCheckResult <-chan runner.Result
-	var withdrawAllResult <-chan runner.Result
+	leaseCheckCh := make(chan leaseCheckResponse, 1)
+	resultCh := make(chan runner.Result, 1)
+
+	subscriber, err := bc.bus.Subscribe()
```
hydrogen18, line 205: So it looks like `startCh` is used to send back a possible error from this? The way I got around this constraint was just by doing the call to `bus.Subscribe()` in `newBalanceChecker` and passing the subscriber as an argument to this function.

provider/balance_checker.go
```diff
@@ -2,153 +2,316 @@ package provider
...
 	}
-	cancel()
 
 	bc.log.Debug("shutdown complete")
 }
+
+func (bc *balanceChecker) timerFunc(ctx context.Context, d time.Duration, lid mtypes.LeaseID, scheduledWithdraw bool, ch chan<- leaseCheckResponse) *time.Timer {
+	return time.AfterFunc(d, func() {
+		select {
+		case <-bc.ctx.Done():
```
hydrogen18, line 313: This seems odd to me. The `*balanceChecker` has a context & the function has a context argument?

provider/balance_checker_test.go
```diff
@@ -2,149 +2,144 @@ package provider
...
+
+	startedAt := time.Now()
+
+	nodeSyncInfo := &tmrpc.SyncInfo{
+		LatestBlockHeight: 1,
+		CatchingUp:        false,
+	}
+
+	client := akashmock.NewClient(t)
+	// client.On("Tx").Return(s.broadcast)
```
hydrogen18, line 55: :scissors:

provider/event/events.go
troian: `startWithdraw` function in `balance_checker.go`

provider/balance_checker.go
```diff
@@ -2,153 +2,316 @@ package provider
...
+	var lResp *mtypes.QueryLeasesResponse
+
+	// Fetch the balance of the escrow account
+	dResp, resp.err = bc.aqc.Deployment(ctx, &dtypes.QueryDeploymentRequest{
+		ID: lid.DeploymentID(),
+	})
+
+	if resp.err != nil {
+		return resp
+	}
```
hydrogen18, line 142: After this check if the deployment is closed, and if it is just return? We're done at that point.  It could happen if the deployment is closed as this code goroutine is running.

provider/balance_checker.go
```diff
@@ -2,153 +2,316 @@ package provider
...
-			if err := withdrawAll.Error(); err != nil {
-				bc.log.Error("failed to started withdrawals", "err", err)
+			switch res.state {
+			case respStateScheduledWithdraw:
+				// reschedule periodic withdraw if configured
+				if bc.cfg.WithdrawalPeriod > 0 {
+					lState.scheduledWithdrawAt = time.Now().Add(bc.cfg.WithdrawalPeriod)
+				}
+				fallthrough
+			case respStateOutOfFunds:
```
hydrogen18, line 272: My understanding is this should only happen once per lease, so we should probably log this at an info level? It's significant enough I would think.

provider/balance_checker.go
```diff
@@ -2,153 +2,316 @@ package provider
...
 		}
 	}
-	cancel()
 
 	bc.log.Debug("shutdown complete")
 }
+
+func (bc *balanceChecker) timerFunc(ctx context.Context, d time.Duration, lid mtypes.LeaseID, scheduledWithdraw bool, ch chan<- leaseCheckResponse) *time.Timer {
+	return time.AfterFunc(d, func() {
+		select {
```
hydrogen18, line 312: This is very similar to `bc.runEscrowCheck`. The primary difference is `case <- bc.lc.Done()` which is probably insignificant? Can this function be updated into something that just calls `bc.runEscrowCheck` ?

provider/balance_checker.go
troian: to me it looks more convenient as `return bc, nil` clearly states at this point all is good.

provider/balance_checker.go
troian: That's actually a good call. If the lease is running for a year provider won't be able to access funds until withdrawn.

tidrolpolelsef (approved):

PR #1618, opened 2022-06-16
fix(provider): remove waitgroup for lease withdraw

troian:
fixes #eng-282

### DIFF
--- a/provider/cluster/manager.go
+++ b/provider/cluster/manager.go
@@ -239,13 +239,14 @@ loop:
 		dm.log.Debug("read from runch during shutdown")
 	}
 
+	if nil != dm.withdrawal {
+		dm.log.Debug("waiting on withdrawal")
+		<-dm.withdrawal.lc.Done()
+	}
+
 	dm.log.Debug("waiting on dm.wg")
 	dm.wg.Wait()
 
-	if nil != dm.withdrawal {
-		dm.log.Debug("waiting on withdrawal")
-		dm.withdrawal.lc.Shutdown(nil)
-	}
 	dm.log.Info("shutdown complete")
 }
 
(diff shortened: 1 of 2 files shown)


provider/cluster/manager.go
```diff
@@ -239,13 +239,14 @@ loop:
 		dm.log.Debug("read from runch during shutdown")
 	}
 
-	dm.log.Debug("waiting on dm.wg")
-	dm.wg.Wait()
-
 	if nil != dm.withdrawal {
 		dm.log.Debug("waiting on withdrawal")
-		dm.withdrawal.lc.Shutdown(nil)
+		dm.withdrawal.lc.Done()
```
tidrolpolelsef, line 244: Doesn't this return a channel? What 's the point of calling it here if we don't use the channel for something?

troian: good catch!

boz: you can use `ShutdownAsync()` if you don't want to wait.  `Done()` does just return a channel that is readable when the object is done.

boz: ideally the `withdrawal` object would have `Shutdown()`, `Done()` methods & you can wait on `<-dm.withdrawal.Done()`.

troian: it supposed to be `<-dm.withdrawal.lc.Done()`. i just missed read part

.github/workflows/k8s-integration.yaml
```diff
@@ -7,7 +7,7 @@ on:
   pull_request:
     paths:
       - pkg/**
-      - provider/cluster/kube/*
+      - provider/*
```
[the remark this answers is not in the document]
tidrolpolelsef: Yeah make that a separate PR please.

tidrolpolelsef (approved):

PR #1623, opened 2022-06-17
WIP feat: pubsub

troian:
- [x] redis (needs manually started Redis instance on local machine)

refs ovrclk/engineering#235

### DIFF
--- a/pubsub/redis.go
+++ b/pubsub/redis.go
@@ -1,0 +1,224 @@
+//go:build events_redis
+
+package pubsub
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"reflect"
+	"strings"
+
+	"github.com/boz/go-lifecycle"
+	"github.com/go-redis/redis/v8"
+)
+
+type bus struct {
+	ctx    context.Context
+	client *redis.Client
+}
+
+type subscriber struct {
+	ctx           context.Context
+	cancel        func()
+	client        *redis.Client
+	lc            lifecycle.Lifecycle
+	eventch       chan Event
+	subscriptions map[string]bool
+}
+
+var _ Bus = (*bus)(nil)
+
+var _ Subscriber = (*subscriber)(nil)
+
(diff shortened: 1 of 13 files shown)


tidrolpolelsef:
@troian is this meant for review at this time?

troian:
it just poc for now. not for review

pubsub/redis.go
```diff
@@ -0,0 +1,224 @@
...
+		b.cancel()
+		_ = b.client.Close()
+
+		b.lc.ShutdownCompleted()
+	}()
+
+	var subs *redis.PubSub
+
+	for topic := range b.subscriptions {
+		if subs == nil {
```
tidrolpolelsef, line 89: The control flow inside this loop is weird & their is an empty `if err != nil` down below. What is going on in this?

pubsub/redis.go
```diff
@@ -0,0 +1,224 @@
...
+	subch := subs.Channel()
+
+	var evbuf []Event
+	var outch chan<- Event
+	var curev Event
+
+loop:
+	for {
+		if b.eventch != nil && len(evbuf) > 0 {
+			// If we're emitting events (Subscriber mode) and there
```
tidrolpolelsef, line 122: The type of this is `*subscriber`. What other mode is there?

pubsub/types.go
```diff
@@ -0,0 +1,71 @@
...
+
+func EventBusRegisterTypes(msgType ...interface{}) error {
+	defer registeredTypesLock.Unlock()
+	registeredTypesLock.Lock()
+
+	for _, mtype := range msgType {
+		msg := reflect.ValueOf(mtype)
+		mType := msg.Type().String()
+		if _, exists := registeredTypes[mType]; exists {
+			return ErrTypeAlreadyRegistered
```
tidrolpolelsef, line 55: This isn't actually very helpful of a message, since a single call registers a whole bunch of messages potentially. We'd need to make an error with the problematic type in it

PR #1630, opened 2022-06-23
feat(provider): add IP address marketplace based off metal LB

tidrolpolelsef:
This is my final PR for the ip address marketplace.


provider/cluster/kube/cluster_ip_connections.go
```diff
@@ -0,0 +1,323 @@
...
+			providerAddr: providerAddr,
+			sharingKey:   v.Spec.SharingKey,
+			protocol:     proto,
+		}
+		evData[i] = ev
+	}
+
+	data = nil
+
+	output := make(chan v1beta2.IPResourceEvent)
```
boz, line 209: any reason to not initialize it with a buffer size of `len(evData)` so that it doesn't initially block below?

boz (approved):
Looks great!

Does script-based pricing work?

PR #1638, opened 2022-06-28
fix(docs): move the metal LB service network policy to the _docs directory

tidrolpolelsef:
This needs to be in `_docs` since it is something end users needs to create in Kubernetes when setting up a provider, if they want to lease IP addresses.


boz (approved):
:rocket:

Issue #286, opened 2018-07-16
account query scoped to tennant


aastein:
- Implement `query account <acct address>` 

- a query with no address `query account` should return the tenants account.

- query will then require a key

PR #1641, opened 2022-07-01
fix(provider): apply missing volume templates on deployment update

troian:
if deployment with persistent storage is being updated it should
reuse existing persistent volume templates

ref #286


tidrolpolelsef:
Is it worth adding a test to make sure that value isn't empty when the deployment specifies an attached volume?

There is probably no need to actually run this against kubernetes in a unit test.

troian:
indeed

tidrolpolelsef:
OK, do you want to add that now or just make another issue?

tidrolpolelsef (approved):

PR #1678, opened 2022-07-28
fix(provider): add catch all CIDR for incoming request to services wi…

tidrolpolelsef:


provider/cluster/kube/builder/netpol.go
```diff
@@ -215,6 +215,15 @@ func (b *netPol) Create() ([]*netv1.NetworkPolicy, error) { // nolint:golint,unp
 				Spec: netv1.NetworkPolicySpec{
 					Ingress: []netv1.NetworkPolicyIngressRule{
 						{
+							From: []netv1.NetworkPolicyPeer{
+								{
+									PodSelector:       nil,
```
troian, line 220: shouldn't we have a namespace for the lease set?

tidrolpolelsef: I'm checking now, maybe that is what is messed up here

tidrolpolelsef: Actually thinking about it no, the traffic is external. This ingress policy applies to the pods selected by the `PodSelector` in the code below. It isn't coming from a namespace, but outside the cluster.

I will remove the `nil` members since it's not any useful info here.

troian (approved):

PR #1715, opened 2022-11-15
fix(provider): keep the command in manifest (#ENG-647)

andy108369:
PR for `0.16.8`

have tested this on my provider using image: `docker.io/andrey01/akash:0.16.8-eng-647` (digest: `fab698b418849`).

- SDL with `command` and `args`;
- `command` is now present in the manifest file;
- `akash-provider` restart did not cause the deployment change (as previously it was removing the `command`);
- hence, no additional replicaset with `CrashLoopBackOff` pod;

The `lease-shell` issue (#ENG-538) is still present though and it is easy to trigger by restarting the `akash-provider`.

### DIFF
--- a/pkg/apis/akash.network/crd.yaml
+++ b/pkg/apis/akash.network/crd.yaml
@@ -50,6 +50,10 @@ spec:
                             type: string
                           image:
                             type: string
+                          command:
+                            type: array
+                            items:
+                              type: string
                           args:
                             type: array
                             items:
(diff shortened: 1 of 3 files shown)


andy108369:
@troian  this should go to provider-services, let me know if that PR looks okay so I can open it there

troian:
superseded by https://github.com/ovrclk/provider-services/pull/50

PR #1962, opened 2024-12-22
fix: Fix Incorrect sed Regular Expression Handling Update tools.sh

0xbryer:
## Description

I noticed an issue with how a regular expression was used in a `sed` command within the bash script. Specifically, the expression:  

```bash  
sed 's/go*//'  
```  

This approach has unintended behavior because the `*` in `sed` matches zero or more occurrences of the preceding character. As a result, it removes any sequence of `g` and `o`, which might not align with the intended functionality.  

To address this, I updated the code to ensure it only removes the exact prefix `go` at the beginning of the string:  

```bash  
sed 's/^go//'  
```  

Here’s the breakdown of the fix:  
- Added `^` to anchor the match to the start of the string.  
- Adjusted the expression to match `go` strictly at the beginning, avoiding accidental removal of unintended sequences.  

---

### Author Checklist

This correction ensures that the script behaves as expected and only removes the `go` prefix where necessary.

- [x] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [x] added `!` to the type prefix if API or client breaking change
- [x] targeted the correct branch (see [PR Targeting](https://github.com/akash-network/node/blob/main/CONTRIBUTING.md#paperwork-for-pull-requests))
- [x] provided a link to the relevant issue or specification
- [x] included the necessary unit and integration [tests](https://github.com/akash-network/blob/main/CONTRIBUTING.md#testing)
- [x] added a changelog entry to `CHANGELOG.md`
- [x] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [x] updated the relevant documentation or specification
- [x] reviewed "Files changed" and left comments if necessary
- [x] confirmed all CI checks have passed

### DIFF
--- a/script/tools.sh
+++ b/script/tools.sh
@@ -16,7 +16,7 @@ function get_gotoolchain() {
     if [[ ${gotoolchain} == "" ]]; then
         # determine go toolchain from go version in go.mod
         if which go > /dev/null 2>&1 ; then
-            local_goversion=$(GOTOOLCHAIN=local go version | cut -d ' ' -f 3 | sed 's/go*//' | tr -d '\n')
+            local_goversion=$(GOTOOLCHAIN=local go version | cut -d ' ' -f 3 | sed 's/^go//' | tr -d '\n')
             if [[ $($SEMVER compare "v$local_goversion" v"$goversion") -ge 0 ]]; then
                 goversion=$local_goversion
             else


script/tools.sh
```diff
@@ -16,7 +16,7 @@ function get_gotoolchain() {
     if [[ ${gotoolchain} == "" ]]; then
         # determine go toolchain from go version in go.mod
         if which go > /dev/null 2>&1 ; then
-            local_goversion=$(GOTOOLCHAIN=local go version | cut -d ' ' -f 3 | sed 's/go*//' | tr -d '\n')
+            local_goversion=$(GOTOOLCHAIN=local go version | cut -d ' ' -f 3 | sed 's/^go//' | tr -d '\n')
```
troian, line 19: this will fail i believe. here is output of `go version` command
```
go version go1.23.4 darwin/arm64
```

PR #1963, opened 2024-12-24
chore: wrong link in adr-001-network-upgrades.md

SihanoukSolver29:
## Description

Path to `../upgrades/upgrades.go` should be `../../upgrades/upgrades.go` in [adr-001-network-upgrades.md](https://github.com/akash-network/node/blob/main/_docs/adr/adr-001-network-upgrades.md)


troian:
hi @SihanoukSolver29 thank you.
can you fix commit headline according to conventional commits

PR #1980, opened 2025-03-31
refactor(events): make events handler as service

troian:
## Description

Closes: #XXXX

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow-up issues.*

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/akash-network/node/blob/main/CONTRIBUTING.md#paperwork-for-pull-requests))
- [ ] provided a link to the relevant issue or specification
- [ ] included the necessary unit and integration [tests](https://github.com/akash-network/blob/main/CONTRIBUTING.md#testing)
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed


events/publish.go
```diff
@@ -20,57 +20,124 @@ import (
...
 		}
 	}
 
-	return err
+	return e.ctx.Err()
 }
 
-func processBlock(ctx context.Context, bus pubsub.Bus, client tmclient.Client, height int64) {
-	blkResults, err := client.BlockResults(ctx, &height)
+func (e *events) processBlock(height int64) {
```
cloud-j-luna, line 139: Could events be processed and published concurrently? Perhaps we could leverage a sized pool of goroutines to process and publish these events. I'm unsure if `Publish` relies on some sort of shared state and events should be published in the specific order. What do you think?

troian: don't think performance will greater improve with parallelizing here, neither I think it actually needed. tho order of events does not matter as well

events/publish.go
```diff
@@ -20,57 +20,124 @@ import (
...
+		ctx:    ctx,
+		group:  group,
+		client: client,
+		lc:     lifecycle.New(),
+		bus:    bus,
+	}
+
 	const (
 		queuesz = 1000
 	)
```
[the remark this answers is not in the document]
troian: does not seem queue size of receiving size matters at all, we'll remove it in sdk-47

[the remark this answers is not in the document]
cloud-j-luna: > does not seem queue size of receiving size matters at all, we'll remove it in sdk-47

Got it

events/publish.go
cloud-j-luna: > don't think performance will greater improve with parallelizing here, neither I think it actually needed. tho order of events does not matter as well

being the "relevant" work done by the consumers then, I assume. Just got a look at the event processing function and indeed it's not worth parallelizing those calls I believe

PR #2068, opened 2026-06-03
fix(market): enforce OrderMaxBids with >= instead of >

renezander030:
Hi, and thanks for all the work on the node.

This fixes the OrderMaxBids off-by-one reported in akash-network/support#413.

`CreateBid` in `x/market/handler/server.go` compared the existing bid count against `OrderMaxBids` with `>`:

```go
if ms.keepers.Market.BidCountForOrder(ctx, msg.ID.OrderID()) > params.OrderMaxBids {
```

`BidCountForOrder` counts Open, Active and Closed bids, so the count reaches `OrderMaxBids` exactly when the cap is hit. With `>`, the check only fired at `OrderMaxBids + 1`, so one extra bid was always allowed. This switches the comparison to `>=`.

### Test
Added `TestCreateBidExceedsOrderMaxBids`: it sets the cap to 1, seeds one bid via the keeper, then asserts a second bid through the handler is rejected with "too many existing bids". It fails on the old `>` (the second bid falls through to an "unknown provider" error) and passes with `>=`.

`go test ./x/market/handler/` and `go vet ./x/market/handler/` pass, and gofmt is clean.

### Note
This changes consensus behaviour (a bid that previously succeeded now fails), so it belongs in a coordinated upgrade. I saw the code freeze for the upcoming network upgrade, so feel free to queue it for whenever that lands. Happy to adjust if you would rather have `BidCountForOrder` exclude Closed bids, which the issue flags as a possible follow-up.

### DIFF
--- a/x/market/handler/handler_test.go
+++ b/x/market/handler/handler_test.go
@@ -915,6 +915,45 @@ func TestCreateBidValid(t *testing.T) {
 	require.True(t, found)
 }
 
+// TestCreateBidExceedsOrderMaxBids is a regression test for the OrderMaxBids
+// off-by-one (akash-network/support#413). The original check used `>`, which
+// allowed OrderMaxBids+1 bids per order. With the cap set to 1 and one bid
+// already present, a further bid must be rejected.
+func TestCreateBidExceedsOrderMaxBids(t *testing.T) {
+	suite := setupTestSuite(t)
+
+	order, gspec := suite.createOrder(testutil.Resources(t, testutil.WithDenom("uact")))
+
+	params, err := suite.MarketKeeper().GetParams(suite.Context())
+	require.NoError(t, err)
+	params.OrderMaxBids = 1
+	require.NoError(t, suite.MarketKeeper().SetParams(suite.Context(), params))
+
+	// Seed the single permitted bid directly via the keeper so the order is at
+	// its cap. The OrderMaxBids check runs before any provider/escrow lookup, so
+	// no provider registration or bank mocks are required here.
+	roffer := mvbeta.ResourceOfferFromRU(gspec.Resources)
+	_, err = suite.MarketKeeper().CreateBid(suite.Context(), mv1.MakeBidID(order.ID, testutil.AccAddress(t)), order.Price(), roffer, nil)
+	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


chalabi2 (dismissed):
LGTM

Interested to hear your take on excluding closed bids from the bid count without causing DOS issues or allowing close records to inflate state.

renezander030:
Thanks! I left that out of scope here since this PR is just the off-by-one, but it's worth digging into.

Closed bids count today because `BidCountForOrder` is the only backpressure on an order, so that count doubles as a lifetime cap. If you drop `BidClosed` from it, the DOS and state inflation turn into the same bug: a provider loops `CreateBid` > `CloseBid` > `CreateBid` forever, never hits the cap, and every cycle leaves a `BidClosed` record until the order settles.

So closed bids can't just be excluded, you'd need to replace the cap that count was providing. One clean option is a monotonic per-order counter (bids-ever-created), bumped on `CreateBid` and never decremented on close. `CreateBid` caps on that, so total bids stay bounded regardless of churn, which kills both the DOS and the record buildup, while a separate live count can still exclude closed bids for anything that cares about current competition. It's also O(1) instead of the current three-state scan.

I can spin that into a follow-up PR if the direction sounds right.

chalabi2:
> I can spin that into a follow-up PR if the direction sounds right.

Yeah that would be great, your implementation idea sounds similar to what I had in mind so go for it.

chalabi2:
@renezander030 can you sign your commits?

renezander030:
Commits are signed now. Follow-up PR with the monotonic counter is up: #2071.

renezander030:
Friendly ping — this one's approved and still relevant. Anything blocking the merge, or is it waiting on a release window? Happy to help if needed.

renezander030:
Same note as on #2071 — this was closed by the stale bot, not on review. @chalabi2 the off-by-one is still present in master (`BidCountForOrder(...) > params.OrderMaxBids` at `x/market/handler/server.go:45`), so this fix is still needed and isn't handled elsewhere. Commits are signed and CodeRabbit is clean.

I can't reopen it myself (repo blocks reopening bot-closed PRs); the branch is intact. Could you reopen from your side, or should I resubmit as a new PR? Happy to help however's least friction.

PR #2071, opened 2026-06-11
fix(market): bound bids per order with a monotonic bids-ever-created counter

renezander030:
Follow-up to #2068, implementing the design discussed in https://github.com/akash-network/node/pull/2068#issuecomment-4666732089 (ack: https://github.com/akash-network/node/pull/2068#issuecomment-4676418653).

**Note: stacked on #2068** — the first commit here is that PR's commit; the diff collapses to the second commit once #2068 merges.

## Problem

`CreateBid` enforced `OrderMaxBids` by scanning the order's open, active and closed bids on every call:

- Counting closed bids was the only backpressure on an order, so the count doubled as a lifetime cap. Excluding them (as proposed in the #2068 review thread) would have opened a churn loop: a provider cycles `CreateBid` > `CloseBid` forever, never hits the cap, and every cycle leaves a `BidClosed` record behind.
- The scan missed `BidLost` entirely, so lost bids escaped the cap.
- It cost an iteration over three state indexes per `CreateBid`.

## Change

- New per-order **monotonic counter** (`bid_counts`, prefix `0x12 0x06`): bumped on `CreateBid`, never decremented when a bid closes or loses. The cap check reads it in O(1), and total bid records per order stay bounded by `OrderMaxBids` regardless of churn — closed and lost bids included.
- **Lazy seeding** for orders that predate the counter: when no counter record exists, the count is derived from a one-time scan of all stored bids (all four states). The cap therefore carries over an upgrade or genesis import unchanged, and no store migration is needed. The counter is read (and seeded) *before* the new bid is stored so the seed never double-counts the bid being created.
- `BidCountForOrder` now returns only **live** (open or active) bids, for consumers that care about current competition on an order rather than the lifetime cap.

## Tests

- `TestCreateBidCapSurvivesBidChurn` — closing a bid drops the live count to zero but does not free cap capacity; a further bid is rejected.
- `TestBidsEverCreatedLazySeedsFromStore` — bids seeded directly in the store (genesis-style, no counter record) still count toward the cap, including closed and lost ones.
- `TestCreateBidExceedsOrderMaxBids` from #2068 passes unchanged, now exercising the counter path.

`go test ./x/market/... -count=1` passes; full repo builds.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/x/market/handler/handler_test.go
+++ b/x/market/handler/handler_test.go
@@ -954,6 +954,96 @@ func TestCreateBidExceedsOrderMaxBids(t *testing.T) {
 	require.Contains(t, err.Error(), "too many existing bids")
 }
 
+// TestCreateBidCapSurvivesBidChurn verifies that the OrderMaxBids cap is
+// enforced against the monotonic bids-ever-created counter rather than the
+// live bid count. Closing a bid must not free up cap capacity, otherwise a
+// provider could loop CreateBid/CloseBid to accumulate unbounded bid records.
+func TestCreateBidCapSurvivesBidChurn(t *testing.T) {
+	suite := setupTestSuite(t)
+
+	order, gspec := suite.createOrder(testutil.Resources(t, testutil.WithDenom("uact")))
+
+	params, err := suite.MarketKeeper().GetParams(suite.Context())
+	require.NoError(t, err)
+	params.OrderMaxBids = 1
+	require.NoError(t, suite.MarketKeeper().SetParams(suite.Context(), params))
+
+	roffer := mvbeta.ResourceOfferFromRU(gspec.Resources)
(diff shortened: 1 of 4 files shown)


renezander030:
@chalabi2 friendly ping — this is the monotonic-counter follow-up you green-lit on #2068. Ready for review whenever you have a moment, thanks!

renezander030:
This was auto-closed by the stale bot, not on review grounds. @chalabi2 you green-lit this direction back on #2068 ("go for it"), and it's clean — CodeRabbit found nothing actionable and the commits are signed. The off-by-one it addresses is still live in master (`x/market/handler/server.go` still uses `>` against `OrderMaxBids`).

I can't reopen from my side (the repo blocks reopening bot-closed PRs), but the branch and commits are all intact. Could you reopen it from your end, or would you rather I open a fresh PR from the same branch? Happy to rebase if `main` has drifted. Would like to get this landed rather than let it lapse a second time.

PR #2072, opened 2026-06-12
ci: fix homebrew tap script

chalabi2:
## Problem

The homebrew tap stopped receiving `akash` formula updates after v2.0.1 (2026-03-23). v2.1.0 was released on 2026-06-08 but never reached the tap.

## Root cause

The `notify-homebrew` job gates the tap dispatch on `script/mainnet-from-tag.sh`, which implements the old **even minor = mainnet** convention. That convention is obsolete: mainnet upgrades have shipped with odd minors since v1.1.0 (see `net/mainnet/upgrades/` — v1.1.0, v2.1.0). As a result the dispatch step was silently `skipped` for v2.1.0 ([run](https://github.com/akash-network/node/actions/runs/27172881532)) — and historically for all of v1.1.x as well (the tap has no v1.1.x entry).

The same dead convention affects two other release paths (fixed here as well):
- **docker tags**: `IS_STABLE` required `IS_MAINNET`, so v2.1.0's image was published as `:latest` instead of `:stable` (Makefile → `make/releasing.mk` → `.goreleaser.yaml`).
- **release notes**: `genchangelog.sh` restricted the changelog baseline to tags with the same minor parity, so v2.1.0's notes were generated against **v1.1.1** instead of v2.0.1.

Release policy per PM: formulas publish for **every stable tag, never for prereleases**.

## Changes

- `release.yaml`: gate the homebrew notify only on `is_prerelease.sh` (strict equality so script failures fail closed), matching the equivalent gate in the provider repo. Prereleases (rc/alpha/aN) still never publish.
- Delete `dispatch.yaml`: it fired on **every push** with an undefined `RELEASE_TAG`, flooding the tap with failing empty-tag dispatches (e.g. [run](https://github.com/akash-network/homebrew-tap/actions/runs/27354215802)). `notify-homebrew` in `release.yaml` is the real dispatch path.
- `Makefile`: derive `IS_STABLE` from the prerelease check alone, so every stable release publishes `:stable` docker images.
- `genchangelog.sh`: filter changelog tags only by stable vs prerelease, removing the even/odd parity matching.
- Remove `script/mainnet-from-tag.sh`: no consumers remain.


script/genchangelog.sh
```diff
@@ -26,15 +20,8 @@ fi
...
-# s1
-# shellcheck disable=SC1073
-if ! "${SCRIPT_DIR}"/mainnet-from-tag.sh "$to_tag" ; then
-	version_rel="^[v|V]?(0|[1-9][0-9]*)\\.(\\d*[13579])\\.(0|[1-9][0-9]*)$"
-	version_prerel="^[v|V]?(0|[1-9][0-9]*)\\.(\\d*[13579])\\.(0|[1-9][0-9]*)(\\-[0-9A-Za-z-]+(\\.[0-9A-Za-z-]+)*)?(\\+[0-9A-Za-z-]+(\\.[0-9A-Za-z-]+)*)?$"
-else
-	version_rel="^[v|V]?(0|[1-9][0-9]*)\.(\d*[02468])\.(0|[1-9][0-9]*)$"
-	version_prerel="^[v|V]?(0|[1-9][0-9]*)\.(\d*[02468])\.(0|[1-9][0-9]*)(\-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$"
-fi
+version_rel="^[v|V]?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)$"
```
troian: see if you can load regex from `script/semver_funcs.sh`

troian (changes requested):

troian (approved):

PR #2073, opened 2026-06-17
fix(market): reject bids when count reaches OrderMaxBids

neyy91:
## Summary

The bid count check in `CreateBid` used `>` instead of `>=`:

```go
if ms.keepers.Market.BidCountForOrder(ctx, msg.ID.OrderID()) > params.OrderMaxBids {
    return nil, fmt.Errorf("%w: too many existing bids (%v)", mv1.ErrInvalidBid, params.OrderMaxBids)
}
```

This allowed **one more bid than `OrderMaxBids`** to be created. When the existing bid count already equals the limit, the next bid passes the check (`count > limit` is false at `count == limit`), so `OrderMaxBids + 1` bids end up on the order — violating the parameter.

## Change

Use `>=` so a new bid is rejected once the count reaches `OrderMaxBids`.

```go
if ms.keepers.Market.BidCountForOrder(ctx, msg.ID.OrderID()) >= params.OrderMaxBids {
```

## Tests

Added `TestCreateBidExceedsMaxBids`: sets `OrderMaxBids = 1`, creates one bid (allowed), then a second bid from another provider must be rejected with `ErrInvalidBid`. Verified the test fails without the fix (a second bid is created) and passes with it. `go test ./x/market/...`, `gofmt`, `go vet` and `golangci-lint` (repo config) are clean.

## Note

The reporter also observed that `BidCountForOrder` counts bids in all states (Open/Active/Closed). Whether closed bids should count toward the limit is a separate, behaviour-affecting question and is intentionally left out of this change; happy to follow up if maintainers want it addressed.

refs: akash-network/support#413

### DIFF
--- a/x/market/handler/handler_test.go
+++ b/x/market/handler/handler_test.go
@@ -1142,6 +1142,66 @@ func TestCreateBidAlreadyExists(t *testing.T) {
 	require.Error(t, err)
 }
 
+// TestCreateBidExceedsMaxBids verifies that the bid count limit is enforced when
+// the existing count reaches OrderMaxBids, not only when it exceeds it. With
+// OrderMaxBids set to 1, a single bid is allowed and the next bid must be
+// rejected (otherwise OrderMaxBids+1 bids could be created).
+func TestCreateBidExceedsMaxBids(t *testing.T) {
+	suite := setupTestSuite(t)
+
+	suite.PrepareMocks(func(ts *state.TestSuite) {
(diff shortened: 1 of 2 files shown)


neyy91:
Closing in favor of #2068, which already fixes this off-by-one (> → >=) and predates this PR — I missed it before opening. Apologies for the duplicate. Thanks @renezander030.

PR #2079, opened 2026-07-09
feat(contracts): add pyth pro verifier

chalabi2:
## Summary
Adds a query-only CosmWasm verifier for upgraded Pyth Core payloads. The upgraded flow replaces Wormhole guardian signature verification with the Pyth router signer model, so Akash needs a verifier that checks the router set, expected emitter, and 3-of-5 quorum.

The verifier preserves the existing `VerifyVAA { vaa, block_time }` query shape used by `pyth.wasm`. It reuses the existing VAA envelope parser, but signature verification is router-quorum based, not Wormhole guardian based.

This also wires the contract side into the v2.1.0 upgrade: the upgrade embeds and instantiates the verifier, then migrates the existing Pyth contract to use the new verifier address. Local init now follows the same path and no longer deploys the old Wormhole contract for Pyth.

Hermes/API-key changes are intentionally out of scope for this PR.

## Validation
- `bash -n _run/init.sh script/wasm2go.sh`
- `cargo test -p pyth`
- `cargo test -p pyth-pro-verifier`
- `cargo clippy -p pyth-pro-verifier --all-targets -- -D warnings`
- `cargo build --release --target wasm32-unknown-unknown -p wormhole -p pyth -p pyth-pro-verifier`
- `GOWORK=off go test ./upgrades/...`
- `git diff --check`


_run/init.sh
```diff
@@ -22,8 +22,8 @@ if [[ -z "$CONTRACTS_DIR" ]]; then
 	exit 1
 fi
 
-WORMHOLE_WASM="${CONTRACTS_DIR}/artifacts/wormhole.wasm"
-PYTH_WASM="${CONTRACTS_DIR}/artifacts/pyth.wasm"
+PYTH_VAA_WASM="${CONTRACTS_DIR}/artifacts/pyth_vaa.wasm"
```
troian, line 25: `pyth-vaa.wasm`, `pyth-pro.wasm`

troian (dismissed):

troian (approved):

